PR #7, opened 2016-03-11
Edit to remove number of participants reference

ocdude:
Zoom recently changed normal meetings to be 50 participants. However, licencing can vary, and some institutions can have meetings of various sizes.


rlorenzo:
@ocdude Thanks for the patch!

PR #10, opened 2016-06-30
Fix MySQL-ism in db/upgrade.

nwp90:
Fix MySQL-ism in db/upgrade.php -- without this, upgrade dies for PostgreSQL users. Not tested on MySQL, but should be correct according to docs.

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -149,7 +149,7 @@ function xmldb_zoom_upgrade($oldversion) {
         $dbman->change_field_notnull($table, $field);
         $dbman->change_field_default($table, $field);
         // Meeting is recurring if type is 3.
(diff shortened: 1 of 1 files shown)


rlorenzo:
@nwp90 We fixed this in https://github.com/ucla/moodle-mod_zoom/commit/75dade451e6fa65df0a444e89cc94802631b11d7 and will be releasing this patch in version 1.5. We tested this in Postgres and MySQL. 

We tried your patch but it failed in MySQL.

nwp90:
Aren't standards great... :-/

Particularly amusing in this case, as I got the syntax I used from http://dev.mysql.com/doc/refman/5.7/en/cast-functions.html (!!)

Will test again when we next do a release here.


Cheers,

Nick

PR #16, opened 2017-02-02
use Moodle's curl class to automatically support proxy configuration

jonof:
By using Moodle's abstraction to cURL you get proxy configuration handled automatically, and it helps simplify the code.

This change also reports other kinds of connectivity errors, separate from the 404 test that already exists, during the connection test in the settings page. The actual error is logged to the webserver at Moodle's 'normal' debug level.


rlorenzo:
@jonof Thanks so much for the pull request. Accepting as is.

PR #28, opened 2018-03-14
Fixed three bugs in webservice.php

dasu2013:
1. Updated the http method in meeting update to 'PATCH'. It is not 'POST' in API version 2 (https://zoom.github.io/api/#update-a-meeting)
2. In Zoom API version 2 'alternative host' has moved inside settings. Therefore we need that in '_make_meeting_data_v2' method.
3. There is an error in 'user_getbyemail' method. It is used get all the users as a list using 'users/' API call. However this returns only 30 records bydefault and 300 maximum. So this is not correct way to get user by email address. We can use this 'https://api.zoom.us/v2/users/{userId}' instead of that.


rlorenzo:
@dasu2013 Thanks for the patch! We are busy with upgrading our Moodle instance to 3.4, but after that we will have time in summer to review and integrate your patch.

jrchamp:
I see `licensesnumber` configuration, but it is not used. I would expect something like `make_utmost` to use it?

PR #37, opened 2018-08-02
Improve participant report

rlorenzo:


classes/task/get_meeting_reports.php
```diff
@@ -0,0 +1,196 @@
...
+        $numrecords = 0;
+        $service = new \mod_zoom_webservice();
+        try {
+            $rawparticipants = $service->get_meeting_participants($meetingwebinarinstanceid, $webinar);
+        } catch (moodle_exception $error) {
+            $rawparticipants = array();
+        }
+        foreach($rawparticipants as $rawparticipant) {
+            $participant = $this->format_object_to_record($rawparticipant, $meetingwebinarinstanceid);
+            $DB->insert_record('zoom_meetings_participants', $participant);
```
rlorenzo: @Rohanator9000 @kubilayagi From my reading of this code.. this is always just inserting. But if we rerun the task for the same $meetingwebinarinstanceid it will crash, right? Since it will have duplicate data on unique keys? I can imagine us wanting to re-download data if the meeting was still going on and we called the API. Maybe like for reoccurring meetings?

Can't we catch if there is an insert failure and continue?

classes/task/get_meeting_reports.php
```diff
@@ -0,0 +1,196 @@
...
+        $params = [
+            'meetingwebinarinstanceid' => $meetingwebinarinstanceid
+        ];
+        $users = $DB->get_records_sql($sql, $params);
+        $userids = array();
+        foreach ($users as $user) {
+            $userids[] = $user->participant_universal_id;
+            $num_records += 1;
+        }
+        if (!$DB->delete_records_list('zoom_meetings_participants', 'participant_universal_id', $userids)){
```
rlorenzo: @Rohanator9000 Wouldn't this delete all the user's participant data? Or do participants get a unique id for every meeting they are in?

classes/task/get_meeting_reports.php
```diff
@@ -0,0 +1,196 @@
...
+        global $DB;
+        $num_records = 0;
+        $sql = 'SELECT zmp.participant_instance_id
+                    FROM {zoom_meetings_participants} zmp
+                    WHERE zmp.meeting_webinar_instance_id = :meetingwebinarinstanceid
+        ';
+        $params = [
+            'meetingwebinarinstanceid' => $meetingwebinarinstanceid
+        ];
+        $users = $DB->get_records_sql($sql, $params);
```
rlorenzo: @Rohanator9000 If you just want an array of users, why not just use https://docs.moodle.org/dev/Data_manipulation_API#moodle_database::get_records_list.28.29

$users = $DB->get_records_list('zoom_meetings_participants', 'meetingwebinarinstanceid', array($meetingwebinarinstanceid));

classes/webservice.php
```diff
@@ -408,21 +408,24 @@ public function get_user_report($userid, $from, $to) {
...
-        foreach ($webinars as $webinar) {
-            $uuids[] = $webinar->uuid;
-        }
-
-        return $uuids;
+    public function list_meetings($userid, $webinar) {
+        $url = 'users/' . $userid . ($webinar ? '/webinars' : '/meetings');
+        $instances = $this->_make_paginated_call($url, null, ($webinar ? 'webinars' : 'meetings'));
+        return $instances;
+        // $uuids = array();
```
rlorenzo: @Rohanator9000 Delete unused code.

locallib.php
```diff
@@ -317,3 +284,64 @@ function zoom_update_records(Traversable $zooms) {
...
+}
+
+/**
+ * Gets the Meeting UUIDs associated with a specific Meeting ID.
+ * @param int $meetingid The meeting ID whose UUIDs you want.
+ * @return array The list of UUIDs associated with the meeting ID.
+ */
+function zoom_get_all_instance_info($meetingid) {
+    global $DB;
+    $service = new mod_zoom_webservice();
```
rlorenzo: @Rohanator9000 This $service variable seems to never be used.

locallib.php
```diff
@@ -317,3 +284,64 @@ function zoom_update_records(Traversable $zooms) {
...
+
+    return $userreport;
+}
+
+/**
+ * Gets the Meeting UUIDs associated with a specific Meeting ID.
+ * @param int $meetingid The meeting ID whose UUIDs you want.
+ * @return array The list of UUIDs associated with the meeting ID.
+ */
+function zoom_get_all_instance_info($meetingid) {
```
rlorenzo: @Rohanator9000 Shouldn't this function be better named zoom_get_meeting_uuids()?

report.php
```diff
@@ -44,42 +44,27 @@
...
-$to = optional_param_array('to', $now, PARAM_INT);
-$ffrom = sprintf('%u-%u-%u', $from['year'], $from['month'], $from['day']);
-$fto = sprintf('%u-%u-%u', $to['year'], $to['month'], $to['day']);
-
-/* Cached structure: class->sessions[hostid][meetingid][starttime]
- *                        ->reqfrom
- *                        ->reqto
- *                        ->resfrom
+/**
+ * Cached structure: sessions[meeting_id][uuid]['participants'] -> participant list
```
rlorenzo: @Rohanator9000 I don't understand how we are using the cache anymore. We used to need to cache because we weren't storing participant information locally so we had to store it in a cache. We got the meeting ids that the host had that matched the Zoom meetingid and stored the participants that way.

In this new report we want to pull all the participant data from the mdl_zoom_meetings_participants table now.

report.php
```diff
@@ -89,30 +74,30 @@
                              get_string('participants', 'mod_zoom'));
         $table->align = array('left', 'left', 'left', 'left', 'left');
 
-        foreach ($meetsessions as $starttime => $meet) {
+        foreach ($meetsessions as $uuid => $meet) {
             $row = array();
 
-            $row[] = $meet->topic;
+            $row[] = $meet['title'];
```
rlorenzo: @Rohanator9000 Shouldn't the meeting information come from the mdl_zoom_meetings_participants table? I don't see 'title' in that table, just 'topic'.

classes/task/get_meeting_reports.php
rlorenzo: @Rohanator9000 Or on duplicate key it can perform an update instead?

classes/task/get_meeting_reports.php
```diff
@@ -0,0 +1,196 @@
...
+            if ($this->get_num_calls_left() < 1) {
+                break;
+            }
+            // Get user data and put it into the database for participants.
+            $webinar = $DB->get_record('zoom', array('meeting_id' => $meeting->meeting_webinar_universal_id))->webinar;
+            $this->add_participants_to_db($meeting->meeting_webinar_instance_id, $webinar);
+            // Update the records to marked them as retrieved.
+            $this->mark_as_retrieved($meeting);
+            // Delete the meeting from the queue so we don't get repeated/wasted API calls for future cron jobs.
+            //$this->delete_meeting_from_queue($meeting->meeting_webinar_instance_id);
```
rlorenzo: @Rohanator9000 Delete this unused line. This code is not deleting users anymore.

classes/task/get_meeting_reports.php
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    /**
+     * Update the meeting record to indicate that the meeting report has already been retrieved.
+     * @param stdClass $meetingwebinarinstance The meeting record we want to update.
+     */
+    private function mark_as_retrieved($meetingwebinarinstance) {
+        global $DB;
+        $meetingwebinarinstance->retrieved = 1;
+        $DB->update_record('zoom_meetings_queue', $meetingwebinarinstance);
```
rlorenzo: @Rohanator9000 This can be done in one line: https://docs.moodle.org/dev/Data_manipulation_API#Setting_a_particular_field_in_the_database

classes/task/get_meeting_reports.php
kubilayagi: @rlorenzo I don't think there should be a case where it is rerun on the same meeting. We added a flag to indicate whether we have retrieved the report already for that instance. Also, we wait for the end time before we get the report for a specific instance

classes/task/get_meeting_reports.php
Rohanator9000: @kubilayagi What is this function used for? I can't find any calls to it in the code.

report.php
Rohanator9000: In zoom_get_sessions_for_display() the array is populated with key 'title' (but I'm going to change that because it should match the database).

classes/task/get_meeting_reports.php
```diff
@@ -0,0 +1,226 @@
...
+                                                    'meeting_id, recurring');
+        foreach ($existingwebinars as $existingwebinar) {
+            if ($existingwebinar->recurring || !($DB->record_exists('zoom_meeting_details', array('meeting_webinar_universal_id' => $existingwebinar->meeting_id)))) {
+                $webinarinfo = $service->get_webinar_details_report($existingwebinar->meeting_id);
+            }
+        }
+
+        foreach ($allmeetings as $meeting) {
+            if (!($DB->record_exists('zoom_meeting_details', array('meeting_webinar_instance_id' => $meeting->uuid)))) {
+                $zoomrecord = $DB->get_records_select('zoom', 'meeting_id = :meetingid', array('meetingid' => $meeting->id))[0];
```
jrchamp: Another opportunity to use $DB->get_record()?

PR #40, opened 2018-09-28
performance: Simplify get_user() by not providing login_type

jrchamp:
login_type is an optional parameter to the "Retrieve a user" API call.
The purpose of get_user() was to get the user, not to limit the set of
valid authentication methods. As such, this commit removes the optional
parameter, the logintypes module setting, the login_* language strings,
the ZOOM_SNS_* defined constants and the entry in the README.


rlorenzo:
Looks great!

PR #41, opened 2018-09-28
fix paginated_call issues

jrchamp:
- When you run out of calls_left, report calls would still run, just without waiting.
- Only try to operate on callresult when it's not null.
- Use array_merge() so that we get the combination of the pages and not just the first page of results.


rlorenzo:
Thanks for catching this.

PR #45, opened 2018-12-18
 3.5 mobile support

nstefanski:
Added support for viewing and joining meetings via mobile app in 3.5+

See https://docs.moodle.org/dev/Mobile_support_for_plugins for more details on mobile support in 3.5

Does not currently support features specific to host users, ie displaying the start link instead of join and meeting password

Does not currently support the option to view all meetings in a course -- this will probably need a new webservice in /zoom/classes/external.php

I'm hoping to be able to circle back and add support for these advanced features, but I wanted to at least get the basic functionality out to you for testing before I go on vacation for the holidays!

As I discussed with Rex Lorenzo, this should not alter support for anyone using the app with version 3.4 or lower.

Finally, I should note that I was not able to test the webinar feature because our school does not actually have this on our account. However, I mirrored my app template very closely on /zoom/view.php, so I don't expect any problems there.


classes/output/mobile.php
```diff
@@ -0,0 +1,79 @@
...
+    	$args = (object) $args;
+    	$cm = get_coursemodule_from_id('zoom', $args->cmid);
+ 
+    	// Capabilities check.
+    	require_login($args->courseid , false , $cm, true, true);
+ 
+    	$context = context_module::instance($cm->id);
+ 
+    	require_capability ('mod/zoom:view', $context);
+	  	// right now we're just implementing basic viewing, otherwise we may need to check other capabilities
```
jrchamp: Indentation whitespace is not consistent throughout the file.

classes/output/mobile.php
```diff
@@ -0,0 +1,79 @@
...
+	  	$data = array(
+	  		'zoom' => $zoom,
+	  		'available' => $zoom_state['available'],
+	  		'status' => $zoom_state['status'],
+	  		'start_time' => $start_time,
+	  		'duration' => $duration,
+	  		'option_audio' => $option_audio,
+			//'userishost' => $userishost,
+	  		'cmid' => $cm->id,
+	  		'courseid' => $args->courseid
```
jrchamp: Is `courseid` used?

templates/mobile_view_page.mustache
```diff
@@ -0,0 +1,73 @@
...
+		
+		<%^zoom.webinar%>
+			<ion-item>
+				<p class="item-heading">{{ 'plugin.mod_zoom.passwordprotected' | translate }}</p>
+				<%#zoom.password%><p>{{ 'core.yes' | translate }}</p><%/zoom.password%>
+				<%^zoom.password%><p>{{ 'core.no' | translate }}</p><%/zoom.password%>
+			</ion-item>
+		<%/zoom.webinar%>
+	  
+		<%^zoom.webinar%>
```
jrchamp: There's two "is not webinar" blocks. If keeping them separate is on purpose, it would be good to add a comment explaining why.

nstefanski: In this case, I was just copying the layout of the table in `/zoom/view.php`. See https://github.com/ucla/moodle-mod_zoom/blob/master/view.php#L166 and https://github.com/ucla/moodle-mod_zoom/blob/master/view.php#L180, two separate "is not webinar" blocks in the PHP.

classes/output/mobile.php
nstefanski: Good catch, I put this in here for future development of the host button and "all meetings" links, but not currently used.

classes/output/mobile.php
nstefanski: this was edited on two different editors, I can make a pass through and fix this if it's an issue.

rlorenzo:
@nstefanski @jrchamp Thanks for the review and work. I am going to merge this as is. @nstefanski If you circle back around to improving this feature, please feel free to submit another pull request.

PR #50, opened 2019-03-08
Suggestion: Allow full zoom object in calendar description object

gagathoscai:
This allows us to include the Zoom meeting ID separate from the URL for instance, to show call-in instructions in the calendar invite.  The code in this PR is a suggested way to do it without breaking old installations (they would have to set a config checkbox to change what {$a} is populated with)

It is up to you the best way to do this ultimately of course! I also have not tested this code yet (though I have modified our copy to give the full $zoom object as in e8ebf3c and it works great.

### DIFF
--- a/exportical.php
+++ b/exportical.php
@@ -43,7 +43,7 @@ $context = context_module::instance($cm->id);
 $PAGE->set_context($context);
 
 require_capability('mod/zoom:view', $context);
-
+$config = get_config('mod_zoom');
 // Start ical file.
 $ical = new iCalendar;
 $ical->add_property('method', 'PUBLISH');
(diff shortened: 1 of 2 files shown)


jrchamp:
Would you be open to updating the existing string or adding a new string so that we can see an example? Right now, you have to customize the string for it to work.

gagathoscai:
{$a->zoom_link} would be the new string to replace the existing functionality.  But you could for instance say, 
Join link: {$a->zoom_link}
By phone:
Zoom call in #s:
#####
#####

Meeting ID: {$a->zoom_id}

If I were to put this into code, would I create an extra string template ("Advanced calendar invite") which will be used instead of the other one, when the config is enabled?

rlorenzo:
@gagathoscai No need to make this a new config. I would merge in a pull request if you would update the string to be what you want or just pass the full zoom object and update the string to use $a-> join_url.

gagathoscai:
My concern is that if someone already has the plugin, and has configured a string translation it will stop working when they update to that new version

rlorenzo:
@gagathoscai Good point. Maybe create a new string and put a deprecation notice. See: https://docs.moodle.org/dev/String_deprecation

rlorenzo:
Closing old pull request due to inactivity

PR #57, opened 2019-05-08
Fix #56, Use core JWT lib

### LINKED ISSUE #56: fails to install in Moodle 3.7
Due to changes that landed in master this week (MDL-62599) JWT is now in core so mod_zoom no longer installs due to a conflict with the core JWT lib.

Fatal error: Cannot declare class Firebase\JWT\JWT

I'll file a pull request for this but it might mean you need to branch to support Moodle 3.7 rather than supporting multiple versions in your master branch (better than keeping the old JWT lib in your codebase...)

danmarsden:
Requires Moodle 3.7 and higher.


rlorenzo:
Fixed in https://github.com/ucla/moodle-mod_zoom/commit/2b3e7e858169ec12a2022c9479613755de1d1ddb

PR #59, opened 2019-06-18
issue#58 change start time to not null

tuanngocnguyen:
issue#58


rlorenzo:
Thanks for the patch!

PR #62, opened 2019-07-18
zoom_update_instance: correctly return boolean

jrchamp:
Hi all,

We ran into an issue where `zoom_update_instance()` was calling `update_meeting()` and allowing a Zoom "meeting not found" exception bubble up to the default exception handler. This was breaking multi-update tools like the OU Dates Report. The `zoom_update_instance()` function is expected by Moodle to return a boolean anyway, so it seems like this solution is more aligned with what Moodle is expecting.

If you would like additional changes, please let me know. Ideas that I had included:
* moving the calendar and grade item updates before the webservice update (if we want to update everything locally even if the remote fails)
* moving the `update_record()` after the webservice update (so that we don't update anything locally if the remote is failing)


rlorenzo:
Thanks for the patch!

PR #63, opened 2019-07-23
adding new 'not found on this account' message to list of user errors

nstefanski:
Starting June 22, 2019, we started seeing an error when student users without Zoom accounts tried to join a meeting:
Screenshot 2019-07-22 at 3 01 23 PM

I'm not sure if Zoom recently changed the language in their errors (trying to confirm with them), but since this is only occurring when the plugin is checking whether the user is the host, this error message should be added to the list of responses that do not throw a general Moodle error, but just return no user id.

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -182,7 +182,8 @@ function zoom_is_meeting_gone_error($error) {
  * @return bool
  */
 function zoom_is_user_not_found_error($error) {
-    return strpos($error, 'not exist') !== false || strpos($error, 'not belong to this account') !== false;
+    return strpos($error, 'not exist') !== false || strpos($error, 'not belong to this account') !== false
+        || strpos($error, 'not found on this account') !== false;
 }
 
 /**


jrchamp:
I know they did a set of updates within the past couple days, but I don't see a mention of string updates: https://support.zoom.us/hc/en-us/articles/204758419

nstefanski:
> I know they did a set of updates within the past couple days, but I don't see a mention of string updates: https://support.zoom.us/hc/en-us/articles/204758419

I haven't been able to find anything official about a string update, all I know is that students started seeing this error when trying to join a Zoom meeting yesterday. If you need to test this, the error only seems to be occurring for users who never created a free account, as far as I can tell, but I could be wrong there.

jrchamp:
@nstefanski It looks like this is tracking master and there's now a patch about activity dates on restore. I'm not sure how to fix it so that the two separate issues are separate.

nstefanski:
@jrchamp sorry about that, removed that last commit from my master, so this pull request is back to the single issue

rlorenzo:
Looks good! Thanks for the patch.

PR #65, opened 2019-08-02
increment activity dates during restore

nstefanski:


rlorenzo:
@nstefanski How does this feature work? Suppose I had a meeting Jan 7 and the start date for the course was Jan 1, if I restore into a new course with start date Feb 1 does it restore the new meeting to Feb 7 rather than Jan 7th?

I thought that backup/restore should bring the values they had in the backup file and then someone would use a tool like https://moodle.org/plugins/report_editdates to bulk edit the dates. How do you know that the offset would be appropriate? Our instructors plan content on a weekly basis and if you just offset by number of days that will shift dates by 1 or so depending on the start date of the course. It does not seem dependable or expected to shift dates, right?

nstefanski:
@rlorenzo it works exactly as you describe in the first paragraph. It works with both course restore and (optionally) with the course reset feature.

I understand your concerns regarding shifting the dates, but this change is just bringing the Zoom activity plugin in line with how the core activity plugins already behave during a course restore. While it may be odd to have the Zoom meetings shift one day, for instance, in this case all the course's assignments, quizzes, etc would have similarly shifted, so the better solution would be to restore the new course to start the same weekday as the original and manually change the course start date afterwards, if necessary.

rlorenzo:
@nstefanski okay since this patch will make this the same behavior as the other plugins I'll include it into our next release

PR #66, opened 2019-12-17
TASK: Add auto recording and mute on entry feature

meyyappanv:
Needed to add provision for auto recording and mute on entry for audio


meyyappanv:
Will revisit implementation and recreate the same after a while

PR #73, opened 2020-03-18
Add global mod setting 'proxyurl'.

pefeigl:
Adds a new setting 'proxyurl' that can be used to set a proxy as hostname:port.
This will be used for communication with the Zoom API (but not anywhere else in Moodle).
This way, you can enable a proxy *only* for mod Zoom, and not for any other parts of Moodle.
proxy


pefeigl:
I've tried this on our setup, works fine (on Moodle 3.8). Please tell me if you'd prefer any changes to the PR.

classes/webservice.php
```diff
@@ -121,9 +121,35 @@ public function __construct() {
      * @throws moodle_exception Moodle exception is thrown for curl errors.
      */
     protected function _make_call($url, $data = array(), $method = 'get') {
+        global $CFG;
         $url = API_URL . $url;
         $method = strtolower($method);
-        $curl = new curl();
+        $proxyhost = get_config('mod_zoom', 'proxyhost');
+        $cfg = new stdClass();
+        if(!empty($proxyhost)) {
```
rlorenzo: @pefeigl Can you put a space after the if statement. See https://docs.moodle.org/dev/Coding_style#If_.2F_else

classes/webservice.php
```diff
@@ -121,9 +121,35 @@ public function __construct() {
...
+            $cfg->proxyhost = $CFG->proxyhost;
+            $cfg->proxyport = $CFG->proxyport;
+            $cfg->proxyuser = $CFG->proxyuser;
+            $cfg->proxypassword = $CFG->proxypassword;
+            $cfg->proxytype = $CFG->proxytype;
+            // parse host:port
```
rlorenzo: @pefeigl Can you please make sure comments start with a capital letter and end in a period? See https://docs.moodle.org/dev/Coding_style#Inline_comments

classes/webservice.php
```diff
@@ -121,9 +121,35 @@ public function __construct() {
...
+            list($host,$port) = explode(':', $proxyhost);
+            // temporarily set new values
+            $CFG->proxyhost = $host;
+            $CFG->proxyport = $port;
+            $CFG->proxytype = 'HTTP';
+            $CFG->proxyuser = '';
```
rlorenzo: @pefeigl Is it worth it to provide the full proxy settings? Like being able to specify user/password?

settings.php
```diff
@@ -106,4 +106,7 @@
             '', 0, 1, 0);
     $settings->add($defaultjoinbeforehost);
 
+    $proxyhost = new admin_setting_configtext('mod_zoom/proxyhost', get_string('option_proxyhost', 'zoom'),
+            get_string('option_proxyhost_desc', 'mod_zoom'), '', '/^[a-zA-Z0-9.-]+:[0-9]+$/');
```
rlorenzo: @pefeigl Does PARAM_URL not work instead of the regular expression?

classes/webservice.php
pefeigl: Done.

classes/webservice.php
pefeigl: Done.

classes/webservice.php
pefeigl: What would be the best way to do this? I could let the user specify a full string (schema://user:password@host:port), which I parse, or set up 5 separate input fields.

settings.php
pefeigl: That's a full URL, I *only* want host:port for now.

danmarsden: but... that regex forces a proxy to be entered into the setting and doesn't allow it to be left empty. I've created #80 to cover this.

PR #81, opened 2020-03-31
Allow proxyurl to be empty.

### LINKED ISSUE #80: proxyurl regex forces a proxy to be used
Issue #73 has a regex on the new global setting which prevents the proxyurl from being left as empty - which in turn prevents people from using this plugin if they do not have a proxy. The regex needs to be updated to allow an empty string or should be removed.

pefeigl:
Fix #80.


pefeigl:
I am really sorry about this, I was sure I tried whether empty also worked, but it must have been before adding the regex :-/

rlorenzo:
Thanks @pefeigl !

PR #104, opened 2020-05-29
conditionally mask participant data in reports

### LINKED ISSUE #103: Feature Request: Zoom Meeting plugin modification - Mask Participant Data
### Background

Zoom logs participant data for each meeting/session, such as Join Time, Duration, and Attentiveness Score (see screenshot below). But UCSF's Zoom implementation is set for HIPAA compliance, which means that name and email information are masked. The following screenshot shows how this appears in the Zoom UI:

zoom participant - masked

The Moodle Zoom Meeting plugin is designed to bring session data from Zoom into Moodle, and includes the ability to view participant data. But since participant data is not stored in UCSF's Zoom, there isn't anything to transfer into Moodle. The Zoom instance in Moodle links to the participant info. On the sessions report, clicking on the participant number (in the Participants column) serves up the participants report:

zoom sessions

But there is no detailed participant data, except for the meeting host. Session data for other users is blanked in the participants report. The number of participants is reported, but the link goes to a participants report with no useful data (just blanked user session data):

zoom moodle participant info

This is how the Moodle-Zoom participant report looks in the CLE. This isn't useful since there isn't any participant data to present. It's also confusing. 

Rex Lorenzo at UCLA, the plugin developer, confirmed this is how it works: "We download data from Zoom using their APIs for the participants data and try to match users to their Moodle user accounts via their email or their name. But if the data returned is masked or not matchable we just show what data we get back as is. So the HIPAA compliance is preventing the plugin from getting data that is matchable."

Since the participant data isn't available for HIPAA Zoom sites like UCSF's, the plugin should not present information that isn't available, and provide a note why there isn't participant data.

### Proposed Modification

Include a "Mask Participant Data" checkbox option in the Zoom Meeting plugin admin settings. If this is checked, then the session page would be modified to replace the hyperlinked "number of participants" – that currently appears under the participant column – with the number of participants (but not linked) and the text "Details not available." Next to each instance of "Details not available," an info tip (i.e., a question mark pop-up) will display that, when clicked, will provide a brief note why there is no detailed participant data. See screenshot below:

zoom moodle sessions update

The help pop-up text could say:  "Participant data is not available for this Zoom session (e.g., due to HIPAA-compliance)."

The help text could reside in the language pack (and be editable there, if desired).

### Participants page
Since the session participants page is no longer linked on the session page, it's not expected that the users will come across it (they would need to know the link). But the participants page should still indicate that there is no data here, and it should include the help tooltip text:

zoom sessions participants

### Administration page

Add a "Mask participant data" option to the plugin administration page. When checked (set to YES), the participant data will be masked, as described above:

zoom plugin admin enhancement1

stopfstedt:
fixes #103


participants.php
```diff
@@ -49,6 +49,23 @@
...
+
+$maskparticipantdata = (bool) get_config('mod_zoom', 'maskparticipantdata');
+// If participant data is masked then display a message stating as such and be done with it.
+if ($maskparticipantdata) {
+    echo $OUTPUT->header();
+    echo $OUTPUT->heading($strname);
+    echo $OUTPUT->heading($strtitle, 4);
+
+    // Stop if there is no data.
+    if (empty($participants)) {
```
rlorenzo, line 61: @stopfstedt You can just use the zoom_fatal_error to avoid duplicating code to create an error that will stop code execution:

https://github.com/ucla/moodle-mod_zoom/blob/master/locallib.php#L101

report.php
```diff
@@ -50,6 +50,7 @@
 
 $sessions = zoom_get_sessions_for_display($zoom->meeting_id, $zoom->webinar, $zoom->host_id);
 if (!empty($sessions)) {
+    $maskparticipantdata = (bool) get_config('mod_zoom', 'maskparticipantdata');
```
rlorenzo, line 53: @stopfstedt Generally in PHP you don't need to cast this as a bool. It should work fine without this.

report.php
```diff
@@ -67,12 +68,22 @@
...
+
+            } else {
+                $url = new moodle_url('/mod/zoom/participants.php', array('id' => $cm->id, 'uuid' => $uuid));
+                $row[] = html_writer::link($url, $meet['count']);
+            }
         } else {
             $row[] = 0;
         }
 
+
```
rlorenzo, line 86: @stopfstedt Extra line added

participants.php
stopfstedt: done. thanks for the tip!

report.php
stopfstedt: removed casting to boolean.

report.php
stopfstedt: removed blank line.

rlorenzo:
@stopfstedt Thanks!

PR #106, opened 2020-06-05
webservice: 429 Retry-After improvements

jrchamp:
* If it's going to exit anyway based on number of retries, do that before calculating sleep time.
* If they're telling us to wait more than one minute, just give up.

Hi @rlorenzo! I noticed that e88f877 might behave strangely when we're told to wait for more than one minute (as in, I think it will do 20 requests back-to-back, instantly hitting the retry maximum). Also, it was doing work that would then go unused.

It might be good to have a different exception message and constant like `MAX_RETRY_WAIT` so that the 60 second max wait is more visible.

I don't think this code is ready to be merged, because I want to know what your preferences are.

### DIFF
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -42,9 +42,6 @@ if (!class_exists('Firebase\JWT\JWT')) {
     }
 }
 
-define('API_URL', 'https://api.zoom.us/v2/');
-define('MAX_RETRIES', 20);
-
 /**
  * Web service class.
  *
@@ -53,6 +50,24 @@ define('MAX_RETRIES', 20);
  * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 class mod_zoom_webservice {
+
+    /**
+     * API base URL.
+     * @var string
+     */
+    const API_URL = 'https://api.zoom.us/v2/';
+
+    /**
+     * API calls: maximum number of retries.
+     * @var int
+     */
+    const MAX_RETRIES = 20;
+
+    /**
+     * API calls: maximum allowed retry wait time.
+     * @var int
+     */
+    const MAX_RETRY_WAIT = 60;
 
     /**
      * API key
@@ -129,7 +144,7 @@ class mod_zoom_webservice {
      */
     protected function _make_call($url, $data = array(), $method = 'get') {
         global $CFG;
-        $url = API_URL . $url;
+        $url = self::API_URL . $url;
         $method = strtolower($method);
         $proxyhost = get_config('mod_zoom', 'proxyhost');
         $cfg = new stdClass();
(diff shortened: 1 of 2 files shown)


classes/webservice.php
```diff
@@ -183,15 +183,16 @@ protected function _make_call($url, $data = array(), $method = 'get') {
...
                     if ($timediff <= 60) {
                         sleep($timediff);
                         debugging('Received 429 response, sleeping ' . strval($timediff) . ' seconds until next retry. Current retry: ' . $this->makecallretries);
+                        return _make_call($url, $data, $method);
                     }
-                    if ($this->makecallretries > MAX_RETRIES) {
-                        throw new zoom_api_retry_failed_exception($response->message);
-                    }
-                    return _make_call($url, $data, $method);
+                    throw new zoom_api_retry_failed_exception($response->message);
```
rlorenzo, line 195: @jrchamp Yeah, here it would be good to have another error then if the wait time is too long. The error should mention the info in the header like category, type, limit, remaining, and retry after: https://marketplace.zoom.us/docs/api-reference/rate-limits

rlorenzo:
@jrchamp Yeah, a MAX_RETRY_WAIT constant is good to have.

rlorenzo:
> might behave strangely when we're told to wait for more than one minute (as in, I think it will do 20 requests back-to-back, instantly hitting the retry maximum). Also, it was doing work that would then go unused.

How would it do 20 requests back to back? The only task that makes use of the heavy API that would have a longer retry than 1 second would be the get_meeting_reports. That is run on a scheduled task and entries are processed serially. So I don't think it will hit the 40 requests/second. But if it does the script would just wait 1 minute. It shouldn't queue up more API calls.

The issue is the 60,000 requests/day. So the task will just fail. But the task will record the date of the last meeting processed and should continue from where it left off.

jrchamp:
> How would it do 20 requests back to back?

https://github.com/ucla/moodle-mod_zoom/blob/e88f877e9716f324c2ad437a015f6f3fa7217123/classes/webservice.php#L185-L194

If the first call fails with HTTP 429, then the code recursively calls `_make_call()` to retry. This retry request uses the same parameters.

The problem occurs when the `Retry-After` time is more than 60 seconds in the future. In that situation, the code doesn't `sleep()` at all and immediately retries. Because there is no delay, it is very likely that the maximum of 20 retries will all fail very quickly, well before the 60+ seconds of wait time expires. This would also be an issue if the system clock was out of sync.

jrchamp:
Sorry, I need to head out. I got distracted and hyper-focused on the global `define()` instead of adding the custom exception message. Is that something you would consider doing? If not that's totally fine. Have a great weekend!

PR #112, opened 2020-06-30
#1 Schedule_for parameter added

mhughes2k:
This should add the schedule_for parameter and send it to the Zoom API.

### DIFF
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -417,6 +417,9 @@ class mod_zoom_webservice {
         if (isset($zoom->password)) {
             $data['password'] = $zoom->password;
         }
+        if (isset($zoom->schedule_for)) {
+            $data['schedule_for'] = $zoom->schedule_for;
+        }
         if (isset($zoom->alternative_hosts)) {
             $data['settings']['alternative_hosts'] = $zoom->alternative_hosts;
         }
--- a/mod_form.php
+++ b/mod_form.php
@@ -191,6 +191,10 @@ class mod_zoom_mod_form extends moodleform_mod {
         $mform->addElement('advcheckbox', 'option_authenticated_users', get_string('option_authenticated_users', 'mod_zoom'));
         $mform->setDefault('option_authenticated_users', $config->defaultauthusersoption);
 
+        // Add Schedule for
+        $mform->addElement('text', 'schedule_for', get_string('schedule_for', 'zoom'));
+        $mform->setType('schedule_for', PARAM_EMAIL);
+
         // Add alternative hosts.
         $mform->addElement('text', 'alternative_hosts', get_string('alternative_hosts', 'zoom'), array('size' => '64'));
         $mform->setType('alternative_hosts', PARAM_TEXT);


jrchamp:
If I'm understanding the API documentation correctly, it might be worthwhile to make this a select list from the list of valid scheduler_for values: https://marketplace.zoom.us/docs/api-reference/zoom-api/users/userschedulers

mhughes2k:
Yes that might be better but for the sake of expediency I kept it to the same UI that Zoom provides (which is just a text box).

rlorenzo:
@mhughes2k Can you please add a help icon to explain what this field is for and that it is expecting an email address?

rlorenzo:
@mhughes2k Also, how is the error handling if the user enters in an invalid email? Can they only choose an email they can schedule for? I have a similar concern as @jrchamp... Can they enter in the email of any Zoom user on the account or only a subset? If it is just a subset how do people get the list of emails they can schedule for? It would be better UX to have a dropdown list. But if it is for anyone, then we can leave it as is.

mhughes2k:
All good points.
 
Zoom docs indicate that it's a choose user from a list, so I'll see if I can modify that, using the get schedulers (the UI that seems to be present is that of a textbox, but I'm guessing it's really doing autocomplete.

I should have flagged this up as a WIP / proof of concept, and I'm working slightly blind. Unfortunately I don't have test Zoom account that I can run this against that's isolate from our real one.

I think that if there is a list of people that you can schedule for though it'll possibly need 2 steps to produce the list of users, as I don't think it should be allowing you to schedule a meeting in a Moodle course that is for a user that isn't able to create Zoom meeting activities *in that course* (even if you hold the scheduling privilege on that user). So I think I'll try to do has_capability() check on the Mod-zoom addinstance capability and only users that appear on that *and* on the current user's list of people are presented.

This would also fix the UI concerns by no longer being a text box and there needing to be explanatory text about the format.

I'll see what I can pull together over the next few days.

rlorenzo:
@mhughes2k  Yeah not having a test environment is a pain. We just test using the real Zoom.. but make sure we don’t login as our create meetings for non test users. It works out fine.

classes/webservice.php
```diff
@@ -384,6 +384,24 @@ public function get_user($identifier) {
...
+    
+    /**
+     * @param string $identifier The user's email or the user's ID per Zoom API.
+     * @return array|false If schedulers are returned array of {id,email} objects. Otherwise returns false.
+     * @link https://marketplace.zoom.us/docs/api-reference/zoom-api/users/userschedulers
+     */
+    public function get_schedule_for_users($identifier) {
+        $url = "users/{$identifier}/schedulers";
+        
+        $schedulers = false;
```
jrchamp, line 396: Not sure which style @rlorenzo prefers, but it might be nice if this always returned an array.
```suggestion
        $schedulers = [];
```

Really, since this value is just the default, you can probably move it into the `catch`

mod_form.php
```diff
@@ -56,6 +56,28 @@ public function definition() {
...
+        // Get list of schedule for users if supported.
+        // List of users who can use Zoom mod in this class.
+        $moodleusers = get_enrolled_users(context $context, 'mod/zoom:addinstance', 0, $'u.*', 'lastnane');
+        
+        /**
+         * @var $scheduleusers array Array of emails and proper names of Moodle users in this course that can add Zoom meetings, and the user can schedule.
+        */
+        $scheduleusers = [];
+        // This will either be false (they can't) or the list of users they can schedule.
+        if(($canschedule = $service->get_schedule_for_users($USER->email)) !== false) {
```
jrchamp, line 69: If it's always an array, you can skip the if check
```suggestion
        $canschedule = $service->get_schedule_for_users($USER->email);
```

mod_form.php
```diff
@@ -56,6 +56,28 @@ public function definition() {
...
+            foreach($canschedule as $zoomuser) {
```
jrchamp, line 70: Just a little breathing room for that keyword
```suggestion
            foreach ($canschedule as $zoomuser) {
```

mod_form.php
```diff
@@ -56,6 +56,28 @@ public function definition() {
...
+                $zoomid = $zoomuser->id;
+                $zoomemail = $zoomuser->email;
+                foreach($moodleusers as $muser) {
```
jrchamp, line 73: Moodle code style
```suggestion
                foreach ($moodleusers as $muser) {
```

mod_form.php
```diff
@@ -56,6 +56,28 @@ public function definition() {
...
+                    if ($muser->email == $zoomemail) {
```
jrchamp, line 74: They're both strings, so exact match is more accurate and better performance.
```suggestion
                    if ($muser->email === $zoomemail) {
```

mod_form.php
```diff
@@ -56,6 +56,28 @@ public function definition() {
             $nexturl = $PAGE->url;
             zoom_fatal_error($errstring, 'mod_zoom', $nexturl, $config->zoomurl);
         }
+        
+        // Get list of schedule for users if supported.
+        // List of users who can use Zoom mod in this class.
+        $moodleusers = get_enrolled_users(context $context, 'mod/zoom:addinstance', 0, $'u.*', 'lastnane');
```
jrchamp, line 62: I'm not sure on this one, but why are we specifying the variable type if the function definition is responsible for that? Also, maybe a stray `$` and field typo?
```suggestion
        $moodleusers = get_enrolled_users($context, 'mod/zoom:addinstance', 0, 'u.*', 'lastname');
```

jrchamp, line 62: Since `$moodleusers` is only used when $canschedule is not empty, it would be good to move this line down a block and inside of an `if` such as:
```
if (!empty($canschedule)) {
    $moodleusers = get_enrolled_users($context, 'mod/zoom:addinstance', 0, 'u.*', 'lastname');
}
```

mhughes2k: Sorry rogue, since I copied the function call's signature from somewhere else.

mhughes2k: Yep good idea

classes/webservice.php
```diff
@@ -384,6 +384,24 @@ public function get_user($identifier) {
...
+        $url = "users/{$identifier}/schedulers";
+        
+        $schedulers = [];
+        try {
+            $response = $this->_make_call($url);
+            $schedulers = $response->schedulers;
+        } catch (moodle_exception $error) {
+            return false;
+        }
+        return $schedulers;
```
jrchamp, line 403: The code works, but line 396 doesn't do anything right now. It's up to you if you want to leave line 401 as false or use the default empty array there. You could also just delete line 401 and add a comment that we don't really care if it threw an exception. I think I would feel more at ease if there was a check to make sure the `$response->schedulers` value either isn't empty or at least is an array.

Again, all of this is optional finishing touches, because the code should work.

mhughes2k:
```
    public function get_schedule_for_users($identifier) {
        $url = "users/{$identifier}/schedulers";
        
        $schedulers = [];
        try {
            $response = $this->_make_call($url);
            if (is_array($response->schedulers)) {
                $schedulers = $response->schedulers;
            }
            // We don't care if this throws an exception.
        }
        return $schedulers;
    }
```

Would seem better

mhughes2k:
I've made changes to the web service, so I guess it's on to testing. Can I throw that over to you guys to look at? (I'm off work now until Monday, and they're not gonna let me run this until we get a test environment...which is apparently feasible)

rlorenzo:
@mhughes2k Thanks, I put this on our next Sprint to try out and test.

kubilayagi:
@mhughes2k This feature works well, but the drop down for choosing a person to scheduler for did not include the current user (since zoom's API response for schedulers excludes the current user). I added a line which inserts the current user as the first one in the list of users that they can schedule meetings for:
https://github.com/ucla/moodle-mod_zoom/blob/b666751d59d91f7be24a245da84706cf8c242b3b/mod_form.php#L71

mhughes2k:
I don't think we should specific the current user in the schedule for parameter if its for themselves. As if you were scheduling your own meeting you wouldn't send a value for this parameter.

I think this probably should have an "empty" self option or a tickbox to enable the schedule for for *other* users.

rlorenzo:
@kubilayagi Add an empty option that wouldn't set schedule option. It doesn't make sense to schedule a meeting for yourself. That is what the normal create meeting does.

mhughes2k:
I've added a "" empty option but with a language string that indicates it's scheduling for "yourself" (as oppposed to someone else).

kubilayagi:
@mhughes2k couple of things:
In the scenario when we want to edit the settings of the meeting, I think we would want to be able to edit the person who the meeting is scheduled for as well (waiting for confirmation from @rlorenzo about this though). In that scenario, if I schedule a meeting for someone else, and then edit settings to assign it back to myself, since the default "schedule for yourself" has an empty string as its value, I can't assign it back to myself. We should replace the empty string key in your change with `$USER->email`

Also, I think the default value of the dropdown should change when we go to edit settings to reflect who it is currently assigned to. As of now, whenever someone opens the edit settings form, the default is always "schedule for yourself", so the user would have to update that manually back to the correct person every time they wanted to make any change to the zoom meeting settings.

However, if we don't want the user to be able to edit who the meeting is scheduled for, then we would still need to do something to prevent people from changing the assigned user

rlorenzo:
@mhughes2k @kubilayagi If you edit the meeting from schedule for to blank, what happens to the meeting? Is it now owned by the user making the change? Or is there an error from the Zoom API?

kubilayagi:
@rlorenzo if the meeting is assigned to someone else, and you use the "schedule for yourself" blank option, the meeting remains assigned to the original person. Other settings that you change will be reflected, so it doesn't seem like assigning to blank string affects the update request at all. No error from zoom

rlorenzo:
@kubilayagi So if you switch the schedule for to another email then does that other meeting get deleted and then assigned to another person?

If not, then maybe we need to lock this field after it is set on initial meeting creation since Zoom is not handling the case when the schedule for field is changed.

kubilayagi:
@rlorenzo I don't think the meeting gets deleted, the host is just changed to the other email. In the case where I assign it first to myself, and then switch it to another person, this feature works well. It's just the other way around that doesn't work. This is easily solved by putting `$USER->email` in the default option. 

The change that would take more time is correctly showing who the meeting is assigned to on zoom when we open the edit settings page.

Also, it might be a good idea to show who the meeting is scheduled for on the view page of the zoom meeting as well. Just a thought

mhughes2k:
So the Zoom UI for schedule for is to have schedule for display Myself / list of people I can schedule for.
If I go back into to the meeting then I can change this to some one else on the list.

Presumably when it's set to myself it's actually "" for the schedule_for parameter (as it were).

The meeting appears on *my* list of meetings and I can go back in and change it back to either myself again or to another host that I can schedule for. So I don't think we can disable it once it's set.

It would seem that, when editing, there's some sort of value that needs to be explicitly set in schedule for to make it change otherwise it seems to just stay the same. 

So a tweak to the form so that if it is a "new" module instance the "myself" option is empty, but when editing an existing one the "myself" option is actually the current user's email address.

I think there's also the fact that "myself" might not be the person who created the meeting, say I created it the coursemodule but @rlorenzo subsequently changes it...the "self" value should still be Michael Hughes, unless Rex explicitly sets it to another person.

rlorenzo:
@mhughes2k Yeah, but that's the problem. Moodle doesn't track who creates a module. So if we don't set the schedule for to be the person creating the meeting then if it is later edited and schedule for is set to "self" but it would be the person now editing right?

So if we set the schedule for to be the user creating, then later editing will retain that value.

However, what happens if the person editing it does not have the user who created the meeting as one of the options for people they can schedule for? Does Zoom still let the user update the meeting?

mhughes2k:
Yes "self" would be the person editing the course module.
Doesn't the Zoom data that comes back from the creation of the Zoom meeting have a value that indicates who it is schedule for / who the host is.

So the schedule for should display the current "zoom" host as the selected value. If that value doesn't appear in the current user's list of people they can schedule for then presumably we should lock the option so it can't be changed as the current user doesn't hold the right to change the scheduled for option.

The Zoom UI would seem to prevent you scheduling for someone you don't have on your list so sidesteps the issue :-)

mhughes2k:
So:

1.  if schedule_for is "empty" the meeting host doesn't change
2. the host details in the meeting is host id, not email address
3. the mod can be edited by someone who isn't the current host
4. if the schedule_for is not empty the meeting host get changes to that user
5. we can't let the editing user change the meeting host to someone they can't schedule for

How about: current implementation for creation is fine, schedule_for is either empty or the email of a designated host the creating/editing user can schedule for.

On editing an existing one, we can't show the current host details (or can we?), so we could do a select box of the users that the editing user can schedule for, but make it an "enabled" one so that they have to turn on the option to "change the host", but they still can only change the host to people they can schedule for...

mhughes2k:
I'm also thinking that this has precedent in existing Moodle behaviour, see changing completion conditions or the max-grade in gradebook. Both require the user to unlock the ability to make a change to particular settings.

kubilayagi:
@mhughes2k With the new implementation, I created a meeting and scheduled it for @rlorenzo. That behavior still works fine, but when I go to edit settings, the dropdown only shows Rex and no "schedule for self" option. I don't think that that is the expected behavior for this. The self option is only added to the dropdown if it is a new zoom meeting being created. (line 65 in mod_form.php)

kubilayagi:
@mhughes2k @rlorenzo check out this new commit. I fixed the issue where the host id was being stored as the person who was creating the meeting, not who it was scheduled for. I also added the default to point to whoever was assigned to it originally, keeping the checkbox to enable or disable the dropdown when the meeting is being edited. Let me know what you think

mhughes2k:
The only query in my head is one that really lies in a problem in the Zoom API docs.

It doesn't seem to make it clear if it expects that a value for {{schedule_for}} parameter to be sent each time or only sent when you are changing the host.

But from a UI perspective this definitely is better in that it displays the assigned host, and it'll be harder to inadvertently change this.

jrchamp:
It looks like cunning users might be able to arbitrarily schedule meetings for other users not in the schedulers list.

Does Moodle's web form automatically verify that the value provided was in the select list or do you need to add a specific check for that in the form submit processing workflow?

mhughes2k:
I've added a mod-form validation() clause to check that the provided data value is in the list of valid zoom users that the user can schedule.

jrchamp:
When the meeting is being updated, it uses [PATCH /meetings/{meetingId}](https://marketplace.zoom.us/docs/api-reference/zoom-api/meetings/meetingupdate)

The only body value here that allows for changing the host is `schedule_for`, not `host_id`. So with the change by @kubilayagi I don't think changing the host is possible. Unless the documentation is wrong.

mhughes2k:
Isn't that to track on the Moodle side who the notional "host" is? I hadn't digged into the API side of the request.

The $zoom->host_id value doesn't appear to be sent to the Zoom API.

mhughes2k:
Also just noted that the schedule_for details doesn't get persisted to the database.

jrchamp:
You don't need to persist schedule_for, because the schedule_for value sets the host_id. To make it work for update, you would need to pass schedule_for, but update host_id (on success).

mhughes2k:
So on this basis @kubilayagi's patch to update the host_id to schedule_for isn't appropriate either.

1. For creates the request URL is made to the calling user's userid
2. For updates it doesn't use this a url with the a user's id in it, but should update the DB's host_id to be the schedule_for user's id. Host_id isn't returned in the `update_meeting` web service data so we have to this post API call on the Moodle side

lib.php
```diff
@@ -124,13 +136,20 @@ function zoom_update_instance(stdClass $zoom, mod_zoom_mod_form $mform = null) {
...
 
     // Update meeting on Zoom.
-    $service = new mod_zoom_webservice();
     try {
         $service->update_meeting($zoom);
+        if (!empty($zoom->schedule_for)) {
+            // Only update this if we actually get a valid user.
+            if ($correcthostzoomuser = $service->get_user($zoom->schedule_for)) {
+                $zoom->host_id = $correcthostzoomuser->id;
+                $DB->update_record('zoom', $zoom);
```
jrchamp, line 145: It would be nice if we could move this line (the one for `update_record()`) and its duplicate (the original `update_record()` line a few lines above) below the try/catch block. That way we only update the record if it also correctly updated Zoom.

@rlorenzo I'm confused by the `update_record()`, `get_record()` sequence that happens right above here in `zoom_update_instance()`. I'm guessing the purpose might be to merge some data before we send the final results to Zoom, but it looks like it's only fetching meeting_id and webinar fields?

mod_form.php
```diff
@@ -56,9 +56,38 @@ public function definition() {
...
+                $zoomid = $zoomuser->id;
+                $zoomemail = $zoomuser->email;
+                foreach ($moodleusers as $muser) {
+                    if ($muser->email === $USER->email) {
+                        // They're already in the $scheduleusers array in 1st position with "Schedule for self"
+                        continue;
+                    }
+                    if ($muser->email === $zoomemail) {
+                        $scheduleusers[$muser->email] = fullname($muser);
+                    }
```
jrchamp, line 85: We don't need to loop through the rest of the users if we found the one that we want.
```suggestion
                        break;
                    }
```

mod_form.php
```diff
@@ -56,9 +56,38 @@ public function definition() {
             $nexturl = $PAGE->url;
             zoom_fatal_error($errstring, 'mod_zoom', $nexturl, $config->zoomurl);
         }
+        $isnew = empty($this->_cm);
```
jrchamp, line 59: This can be moved back to where it was originally now that it's not used for `$scheduleusers`

mod_form.php
```diff
@@ -244,13 +284,27 @@ public function validation($data, $files) {
...
+        require_once($CFG->dirroot.'/mod/zoom/classes/webservice.php');
+        $service = new mod_zoom_webservice();
         if (!empty($data['requirepassword']) && empty($data['meetingcode'])) {
             $errors['meetingcode'] = get_string('err_password_required', 'mod_zoom');
         }
-
+        if ($data['schedule_for'] !== $USER->email) {
+            $scheduleusers = $service->get_schedule_for_users($USER->email);
+            $scheduleok = false;
+            foreach($scheduleusers as $zuser) {
```
jrchamp, line 296:
```suggestion
            foreach ($scheduleusers as $zuser) {
```

version.php
```diff
@@ -25,7 +25,7 @@
 defined('MOODLE_INTERNAL') || die();
 
 $plugin->component = 'mod_zoom';
-$plugin->version = 2020061200;
+$plugin->version = 2020061201;
```
jrchamp, line 28: I don't remember, but is this needed when you add new language strings?

mhughes2k: No, it was because I'd added some code to install.xml and upgrade.php to add a schedule_for field to the database table. 

There's no version bump required now by this code

lib.php
mhughes2k: With all of this being in the try...catch it would only run if the `update_meeting()` against Zoom is successful. 

The original bit ($updatedzoomrecord = $DB->get_records[ ....]) before the try is in the original code, so I'm not sure. Like you say @rlorenzo it looks like it is updating the $zoom object (which is originally from the moodleform, with values that are guaranteed to have come from the database, so is this to try to ensure that no one has tried to inject false values in to the moodle form?

So the reason I put a 2nd update here is only to mutate the host_id if the `update_meeting()` (with a schedule_for parameter) succeeds, so that we know that Zoom _should_ have updated the host_id, so our data is then changed to match (as the API call doesn't return the changed host_id).

classes/webservice.php
```diff
@@ -384,6 +384,46 @@ public function get_user($identifier) {
...
+    }
+
+    /**
+     * @param string $identifier The user's email or the user's ID per Zoom API.
+     * @return array|false If schedulers are returned array of {id,email} objects. Otherwise returns false.
+     * @link https://marketplace.zoom.us/docs/api-reference/zoom-api/users/userschedulers
+     */
+    public function get_schedule_for_users($identifier) {
```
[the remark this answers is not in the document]
mhughes2k: Fixed

version.php
mhughes2k: I've un-bumped version.php, so I'll leave it for you guys to change as part of integrating this with any other changes

mhughes2k:
OK that should only leave the issue around update_record in lib.php open

lib.php
mhughes2k: It'll also fail the whole section too if for any reason the `get_user()` call to Zoom throws an exception (which it *might* do).

jrchamp: If an exception is thrown, it'll correctly return false in the `catch`, so the "success" code flow is roughly the same both at the end of the `try`'s body and below the try/catch. For now, it's safe to leave this here. With Rex's help, we can remove some of the confusion around the `$updatedzoomrecord` piece either as part of this PR review or later.

version.php
```diff
@@ -25,7 +25,7 @@
 defined('MOODLE_INTERNAL') || die();
 
 $plugin->component = 'mod_zoom';
-$plugin->version = 2020061200;
+$plugin->version = 0;
```
jrchamp, line 28: Here's the original value:
```suggestion
$plugin->version = 2020061200;
```

mhughes2k: Damnit thought I’d done exactly that!

mod_form.php
```diff
@@ -191,6 +222,17 @@ public function definition() {
...
         $mform->setDefault('option_authenticated_users', $config->defaultauthusersoption);
 
+        // Add Schedule for if current user is able to.
+        if (!empty($scheduleusers)) {
+            $mform->addElement('select', 'schedule_for', get_string('schedulefor', 'zoom'), $scheduleusers);
+            $mform->setType('schedule_for', PARAM_EMAIL);
+            if (!$isnew) {
+                $mform->disabledIf('schedule_for', 'change_schedule_for');
+                $mform->addElement('checkbox', 'change_schedule_for', get_string('changehost', 'zoom'));
+                $mform->setDefault('schedule_for', $service->get_user($this->current->host_id)->email);
```
jrchamp, line 232: If you set the default here, does the value need to already exist in the $scheduleusers array of options?

mod_form.php
```diff
@@ -57,6 +57,37 @@ public function definition() {
...
 
+
+        /**
+         * @var $scheduleusers array Array of emails and proper names of Moodle users in this course that can add Zoom meetings, and the user can schedule.
+        */
+        $scheduleusers = [];
+        $scheduleusers[$USER->email] = get_string('scheduleforself', 'zoom');
+
+        // This will either be false (they can't) or the list of users they can schedule.
+        $canschedule = $service->get_schedule_for_users($USER->email);
```
jrchamp, line 68: Here's something that we should test.

Getting the list of users that `$USER->email` can schedule for makes sense when creating a new Meeting, but when we go to update the Meeting, is Zoom going to be validating the `scheduler_for` value against the current host's list of possible `schedule_for` values? If so, may need to decide if we want to limit the list to the intersection of the two groups of users (people I can schedule_for and people that the current host can schedule_for) 😢. If instead Zoom lets us change the host to whomever we want because we're using global admin access, then I think this will still work.

lib.php
```diff
@@ -104,6 +109,13 @@ function zoom_add_instance(stdClass $zoom, mod_zoom_mod_form $mform = null) {
 function zoom_update_instance(stdClass $zoom, mod_zoom_mod_form $mform = null) {
     global $CFG, $DB;
     require_once($CFG->dirroot.'/mod/zoom/classes/webservice.php');
+    $service = new mod_zoom_webservice();
+
+    // Removing this as per above for the moment.
+/*    if (!empty($zoom->schedule_for)) {
```
jrchamp, line 115: I wonder if we should unset `$zoom->schedule_for` if the `host_id` is already the `schedule_for` user.

mod_form.php
kubilayagi: Good point. If the person editing the meeting can't schedule it for the person it's already assigned to, maybe we should prevent them from editing `schedule_for` at all

kubilayagi:
@mhughes2k Do you have any updates on this? If you can't work on it, then we can try taking over to make the changes as well

mhughes2k:
I've had a look at the conflicts that were raise following changes to master, and I think I've accomodated them so it mereges cleanly again.

mhughes2k:
@kubilayagi I was actually just about to say with the conflicts fixed, I think it's in a state to pass over to you guys, as I'm not going to be able to look at it again for a while, but we're keen on getting it in ASAP.

kubilayagi:
@mhughes2k Sounds good, I'll take a look! I'm planning to get it in this week

rlorenzo:
@mhughes2k @jrchamp Our developer @kubilayagi cleaned up the branch and made some fixes. It has been tested on our end. But I wanted to let you guys review it if there's anything else that needs addressing: https://github.com/ucla/moodle-mod_zoom/commit/419164edc18053c762d8e75e0ea6585344bad533

mhughes2k:
I've had a quick look at the commit, and nothing to add.

kubilayagi:
@jrchamp if you have time, I'd appreciate another look. I added in the functionality to only show the intersection between the current `$USER` list of schedulers and the previously saved host's schedulers. It's worth noting that there is a loophole to get around that functionality which is to first change the meeting to be scheduled for yourself, and then schedule it for whoever you want afterwards. But at least this way, zoom won't respond with an error

kubilayagi:
@jrchamp i made some final updates to handle the errors and the logic changes you suggested. I'd appreciate one last look before we get this merged onto our test server. Thanks again for all your help on this!

jrchamp:
Looks pretty good. I had a couple comments, but with those small tweaks, I think it's ready for test.

rlorenzo:
Merged feature in https://github.com/ucla/moodle-mod_zoom/commit/220db34b7002910daee57f356e5358b34427eecb

mhughes2k:
Is there a anticipated time scale for a version with this functionality to be formally released?

rlorenzo:
@mhughes2k Released new version at https://moodle.org/plugins/pluginversion.php?id=22417

Issue #110, opened 2020-06-23
Support Breakout Rooms


mhughes2k:
(not checked the Zoom API if this is supported but...)
It would be *really* nice if you could specify a Moodle grouping that would be used to (pre-)configure Zoom breakout rooms, with a breakout room per group in the grouping.

jrchamp:
That would definitely be helpful, particularly if it synchronized with Zoom when there were changes to group membership.

Unfortunately, I don't see it anywhere in the Create Meeting or Update Meeting APIs at this time:
https://marketplace.zoom.us/docs/api-reference/zoom-api/meetings/meetingcreate
https://marketplace.zoom.us/docs/api-reference/zoom-api/meetings/meetingupdate

mhughes2k:
I've started an issue over on our fork about this to try and work out a spec for this feature.

I have a gut feeling that that at least being able to generate the CSV from within the mod-zoom activity would be good...

https://github.com/University-of-Strathclyde-LTE-Team/moodle-mod_zoom/issues/2

jrchamp:
I like that approach - it helps bridge the gap between the current "you're on your own" and the currently impossible "the API lets us do that". Good idea!

aspark21:
Has anyone asked Zoom if they would provide an API for it?

Seems like staff our having trouble with the csv approach mostly because students don't yet have zoom accounts when they upload the csv so it can't match them.

API would ideally allow for the just in time provisioning when the student joined the session.

jrchamp:
It was difficult to find, but it seems that Zoom added support to manage pre-defined Breakout Rooms via the API in January 2021: https://noticeable.news/8xmdmkir8ctlkfj8dttx/publications/api-updates-1-31-21

PR #115, opened 2020-07-06
Support Breakout Rooms (in progress)

mhughes2k:
This is an in-progress branch to implement #110 so that you guys can see what I'm doing.

### DIFF
--- a/breakout.php
+++ b/breakout.php
@@ -22,7 +22,7 @@
  * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 
-require(__DIR__.'/../../config.php');
+require_once(__DIR__.'/../../config.php');
 require_once(__DIR__.'/locallib.php');
 
 $config = get_config('mod_zoom');
--- a/breakout.php
+++ b/breakout.php
@@ -15,7 +15,7 @@
 // along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
 
 /**
- * EXport breakout room CSV for Zoom.
+ * Export breakout room CSV for Zoom.
  *
  * @package    mod_zoom
  * @copyright  2020 Michael Hughes, University of Strathclyde, <michaelhughes@strath.ac.uk>
--- a/breakout.php
+++ b/breakout.php
@@ -49,4 +49,4 @@ $filesafename = preg_replace('/[^a-z0-9-]/', '_',core_text::strtolower(strip_tag
 $export->set_filename('mod-zoom-break-'. $filesafename . '.csv');
 
 $grouping = groups_get_grouping($groupingid);
-groups_get_grouping_members()
+$groups = groups_get_all_groups($course->id, 0, $grouping->id);
--- a/breakout.php
+++ b/breakout.php
@@ -1,0 +1,52 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * EXport breakout room CSV for Zoom.
+ *
+ * @package    mod_zoom
+ * @copyright  2020 Michael Hughes, University of Strathclyde, <michaelhughes@strath.ac.uk>
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+require(__DIR__.'/../../config.php');
+require_once(__DIR__.'/locallib.php');
+
+$config = get_config('mod_zoom');
+
+list($course, $cm, $zoom) = zoom_get_instance_setup();
+
+$context = context_course::instance($course->id);
+
+$iszoommanager = has_capability('mod/zoom:addinstance', $context);
+
+$PAGE->set_url('/mod/zoom/breakout.php', ['id' => $cm->id]);
+
+// Check group mode is actually on.
+
+// Check if there is a grouping set on the activity.
+// Generally feel that it is bad if there is no grouping set, but not everyone.
+$groupingid = 3; // TODO Fake data.
+// We're probably going to need a UI form here to allow user to set some configuration(?)
+
+// Do the export.
+$export = new csv_export_writer();
+$filesafename = preg_replace('/[^a-z0-9-]/', '_',core_text::strtolower(strip_tags($zoom->name)));;
+// ^^ this is what moodle uses in completion/index.php
+$export->set_filename('mod-zoom-break-'. $filesafename . '.csv');
+
+$grouping = groups_get_grouping($groupingid);
+groups_get_grouping_members()
--- a/classes/local/breakoutexporter.php
+++ b/classes/local/breakoutexporter.php
@@ -102,13 +102,13 @@ class breakoutexporter
         if (!$preview) {
             $exp = new csv_export_writer();
         }
-        foreach($this->groups as $group) {
+        foreach ($this->groups as $group) {
             $groupname = $group->name;
-            foreach($this->groupmembers[$group->id] as $members) {
+            foreach ($this->groupmembers[$group->id] as $groupmember) {
                 if ($preview) {
-                    $previewdata[] = [$groupname, $members->email];
+                    $previewdata[] = [$groupname, $groupmember->email];
                 } else {
-                    $exp->add_data([$groupname, $members->email]);
+                    $exp->add_data([$groupname, $groupmember->email]);
                 }
             }
         }
--- a/classes/local/breakoutexporter.php
+++ b/classes/local/breakoutexporter.php
@@ -22,7 +22,7 @@ class breakoutexporter
      */
     private $dataloaded = false;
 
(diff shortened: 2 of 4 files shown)


classes/local/breakoutexporter.php
```diff
@@ -0,0 +1,124 @@
...
+    private $zoom;
+    private $course;
+    private $cm;
+    private $grouping;
+    private $groups;
+    private $groupmembers;
+    /**
+     * @var list of users that we've already seen, and the groups they're in.
+     */
+    private $seenbefore;
```
jrchamp, line 18: This private variable is never reused and if it were, it would result in empty group memberships because the value is never reset.
```suggestion
```

breakout.php
```diff
@@ -0,0 +1,52 @@
...
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+/**
+ * EXport breakout room CSV for Zoom.
```
jrchamp, line 18:
```suggestion
 * Export breakout room CSV for Zoom.
```

breakout.php
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * @package    mod_zoom
+ * @copyright  2020 Michael Hughes, University of Strathclyde, <michaelhughes@strath.ac.uk>
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+require(__DIR__.'/../../config.php');
```
jrchamp, line 25:
```suggestion
require_once(__DIR__.'/../../config.php');
```

breakout.php
```diff
@@ -0,0 +1,52 @@
...
+require_once(__DIR__.'/locallib.php');
+
+$config = get_config('mod_zoom');
+
+list($course, $cm, $zoom) = zoom_get_instance_setup();
+
+$context = context_course::instance($course->id);
+
+$iszoommanager = has_capability('mod/zoom:addinstance', $context);
```
jrchamp, line 34: Should this be `require_capability`?

breakout.php
```diff
@@ -0,0 +1,52 @@
...
+// Check group mode is actually on.
+
+// Check if there is a grouping set on the activity.
+// Generally feel that it is bad if there is no grouping set, but not everyone.
+$groupingid = 3; // TODO Fake data.
+// We're probably going to need a UI form here to allow user to set some configuration(?)
+
+// Do the export.
+$export = new csv_export_writer();
+$filesafename = preg_replace('/[^a-z0-9-]/', '_',core_text::strtolower(strip_tags($zoom->name)));;
```
jrchamp, line 47: This generates a duplicate when the same name is used. This might be okay, but adding the $cm->id could guarantee uniqueness per-activity.

classes/local/breakoutexporter.php
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * @var bool|int False if the data has never been loaded or a timestamp for when it was updated last.
+     */
+    private $dataloaded = false;
+
+    public function __construct(StdClass $zoom)
```
jrchamp, line 25: I believe the `s` is lowercase.
```suggestion
    public function __construct(stdClass $zoom)
```

classes/local/breakoutexporter.php
```diff
@@ -0,0 +1,124 @@
...
+                foreach($gms as $user) {
+                    if (!isset($this->seenbefore[$user->id])) {
+                        $this->seenbefore[$user->id] = [$group->id];
+                        $members[$group->id][] = $user;
+                    } else {
+                        $this->seenbefore[$user->id][] = $group->id;
+                    }
+                }
+            }
+            $this->groupmembers = $members;
```
jrchamp, line 57: Here's my thought on this section, keep the parts you like.
```suggestion
        $this->groupmembers = [];
        if (!empty($this->groups)) {
            $seenbefore = [];
            foreach ($this->groups as $group) {
                $gms = groups_get_members($group->id);
                foreach ($gms as $user) {
                    // A users can only be in one breakout group, so use the first one.
                    if (!isset($seenbefore[$user->id])) {
                        $this->groupmembers[$group->id][$user->id] = $user;
                        $seenbefore[$user->id] = true;
                    }
                }
            }
```

classes/local/breakoutexporter.php
```diff
@@ -0,0 +1,124 @@
...
+     * Preview the data.
+     * @param int $limit
+     * @return array
+     */
+    public function preview() {
+        $data = [];
+        foreach($this->groups as $group) {
+            $groupname = $group->name;
+            foreach($this->groupmembers[$group->id] as $members) {
+                $data[] = [$groupname, $members->email];
```
jrchamp, line 72: Spacing and singular variable name.
```suggestion
        foreach ($this->groups as $group) {
            $groupname = $group->name;
            foreach ($this->groupmembers[$group->id] as $groupmember) {
                $data[] = [$groupname, $groupmember->email];
```

classes/local/breakoutexporter.php
```diff
@@ -0,0 +1,124 @@
...
+            }
+        }
+        return $data;
+    }
+
+    /**
+     * Allow the data to get this old before we refresh it.
+     */
+    const MAX_LIFESPAN = 1800;
```
jrchamp, line 81: Is this being preserved across requests? I have a feeling that it's reloading the data as needed, so this lifespan code might not be doing anything.

classes/local/breakoutexporter.php
```diff
@@ -0,0 +1,124 @@
...
+        if (!$preview) {
+            $exp = new csv_export_writer();
+        }
+        foreach($this->groups as $group) {
+            $groupname = $group->name;
+            foreach($this->groupmembers[$group->id] as $members) {
+                if ($preview) {
+                    $previewdata[] = [$groupname, $members->email];
+                } else {
+                    $exp->add_data([$groupname, $members->email]);
```
jrchamp, line 111:
```suggestion
        foreach ($this->groups as $group) {
            $groupname = $group->name;
            foreach ($this->groupmembers[$group->id] as $groupmember) {
                if ($preview) {
                    $previewdata[] = [$groupname, $groupmember->email];
                } else {
                    $exp->add_data([$groupname, $groupmember->email]);
```

jrchamp (commented):
Did a pass through the code and I see where you're headed. If they haven't selected a grouping, you'll probably want to given them an error message and refer them back to the edit activity page.

breakout.php
mhughes2k: Yes I agree however I was in two minds about it, as we sometimes have people getting distracted by all the "unique" bits when just trying to find the file that was just downloaded.

classes/local/breakoutexporter.php
mhughes2k: I did this this way round so that the $seenbefore array would have enough data to be able to explain to the end user why a user may be appearing in an unexpected group (e.g. if they expect them in a later group).

This would allow the generation of something like;
Michael Hughes appeared in multiple groups (Group A*, Group B, Group Z). They will be in breakout room "Group A").

classes/local/breakoutexporter.php
mhughes2k: My intention was that once the export was done, to cache the results for a "short" period so this was a const to define what the longest time permitted would be. I was anticipating it being substantially lower, and ideally the cached data would be invalidated by events (i.e. if group membership in a class changed).

classes/local/breakoutexporter.php
jrchamp: I understand, but there isn't a function that within the class that uses it, so since it is private, no one can access the value. If you want to keep it, please initialize it to empty when you call `filldata()` so that multiple calls to `filldata()` return the same value.

mhughes2k: You're right, not yet there isn't! But I'll bear this in mind when I'm looking at this again (hopefully this week).

breakout.php
mhughes2k: I think I was intending to do more than bail out immediately if they didn't have the ability to add it...

classes/local/breakoutexporter.php
jrchamp: It's possible that the Moodle core functions that you are using are already cached appropriately. If you find the performance to be suffering, then it could make sense to use the MUC caching and some of the existing event invalidation options (such as listening for group membership change events).

breakout.php
jrchamp: My point here is that because there is no restrictions right now, anyone can get the list of all users and the breakout group they are assigned to. Either requiring the capability or wrapping the export code in an `if ($iszoommanager) {` block is needed.

classes/local/breakoutexporter.php
mhughes2k: Yep, that's where I was intending to go with it ultimately. 

I think the main goal was more around the avoid unecessary writing to disk or processing things in the case that a user is just re-downloading the same data. 

Again based on past experience we've seen people re-download the same, unchanged file to use again (even though it's in their downloads).

breakout.php
mhughes2k: Yes of course. I'm not saying this is production-ready code ( nor anywhere close to it yet!), just opened this so you can have sight of the feature and make sure I'm not going off down a route that would be in conflict with things you guys may be doing.

breakout.php
mhughes2k: Something like "zoom_breakout_rooms_111831_My_great_meeting.csv"

abias:
Hey @mhughes2k , 

I skimmed over the code in this PR and would like to say that I am interested to see this feature finally realized. I particularly like your approach to use Moodle course groups for the provisioning of the breakout rooms.

May I ask you if you had the time to finish this locally since mid 2020 and just didn't update the PR or did this got stuck...?

Thanks in advance,
Alex

Issue #43, opened 2018-11-07
Request for enhancement: another option for 'Activity completion'


bparry13:
When you create a new ZOOM activity, the only option you have for Activity completion is 'Student must receive a grade to complete this activity'.

Currently we host Moodle for multiple clients, and we are getting requests to add an option for 'Activity completion'. The option requested is 'Student must view this activity to complete it'.

The teachers say they are not using the Zoom session for grading, but more as evidence of participation in the course.

jrchamp:
+1 This seems like a valuable change. If I read the documentation correctly, this may be easy to implement: https://docs.moodle.org/dev/Activity_completion_API#Completion_on_view

gsyoung:
+1 From me too. Zoom/ webinars within are mostly about activities not Grading. The 'Viewed activity' is a little more about confirming who did/didn't participate in the webinar. Having both options is the best and most functional outcome.

nstefanski:
I think one potential pitfall of completion on view is that a student can just view the page with the meeting information and never join the meeting. A better analogy might be the Moodle assignment: students can view the assignment, make a submission, and receive a grade for the assignment, and each of these three is a possible completion setting. Moodle also has information on how to add a custom completion setting similar to the assignment submission: https://docs.moodle.org/dev/Activity_completion_API#Custom_completion_rules

rlorenzo:
Fixed in https://github.com/ucla/moodle-mod_zoom/pull/118

PR #117, opened 2020-07-17
Custom completion

nstefanski:
This addresses issue #43 by adding a new completion condition when a student joins the meeting.

I only realized as I was testing but this currently relies on logging being enabled to check completion, because Moodle's {module}_get_completion_state needs to be able to run its own check outside of loadmeeting.php. The alternative would be to add a new db table that records whenever a student clicks the join link in Moodle, similar to the {assign_submission} table. Let me know if you think this way would be preferred and I can update my branch.

Just fyi, the {zoom_meeting_participants} table is not useful for this because it only gets the participant date from Zoom after the fact, long after completion checks have run.

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -440,7 +440,7 @@ function xmldb_zoom_upgrade($oldversion) {
             $dbman->add_field($table, $field);
         }
         
-        upgrade_mod_savepoint(true, 2020052100, 'zoom');
+        upgrade_mod_savepoint(true, 2020071500, 'zoom');
     }
 
     return true;
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -431,5 +431,17 @@ function xmldb_zoom_upgrade($oldversion) {
         upgrade_mod_savepoint(true, 2020052100, 'zoom');
     }
 
+    if ($oldversion < 2020052100) {
+        // Add completionjoin.
+        
+        $table = new xmldb_table('zoom');
+        $field = new xmldb_field('completionjoin', XMLDB_TYPE_INTEGER, '2', null, XMLDB_NOTNULL, null, '0');
+        if (!$dbman->field_exists($table, $field)) {
+            $dbman->add_field($table, $field);
+        }
+        
+        upgrade_mod_savepoint(true, 2020052100, 'zoom');
+    }
+
     return true;
 }
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -431,7 +431,7 @@ function xmldb_zoom_upgrade($oldversion) {
         upgrade_mod_savepoint(true, 2020052100, 'zoom');
     }
 
-    if ($oldversion < 2020052100) {
+    if ($oldversion < 2020071500) {
         // Add completionjoin.
         
         $table = new xmldb_table('zoom');
--- a/lib.php
+++ b/lib.php
@@ -590,7 +590,7 @@ function zoom_get_completion_state($course, $cm, $userid, $type) {
                     WHERE component = 'mod_zoom' AND action = 'clicked'
                       AND target = 'join_meeting_button' 
                       AND objectid = :zoomid AND userid = :userid";
-        return $DB->get_field_sql($logsql, array('userid'=>$userid,'zoomid'=>$zoom->id));
+        return $DB->get_field_sql($logsql, array('userid' => $userid, 'zoomid' => $zoom->id));
     } else {
         // Completion option is not enabled so just return $type
         return $type;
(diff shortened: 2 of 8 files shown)


db/upgrade.php
```diff
@@ -431,5 +431,17 @@ function xmldb_zoom_upgrade($oldversion) {
         upgrade_mod_savepoint(true, 2020052100, 'zoom');
     }
 
+    if ($oldversion < 2020052100) {
```
jrchamp, line 434:
```suggestion
    if ($oldversion < 2020071500) {
```

db/upgrade.php
```diff
@@ -431,5 +431,17 @@ function xmldb_zoom_upgrade($oldversion) {
...
+        // Add completionjoin.
+        
+        $table = new xmldb_table('zoom');
+        $field = new xmldb_field('completionjoin', XMLDB_TYPE_INTEGER, '2', null, XMLDB_NOTNULL, null, '0');
+        if (!$dbman->field_exists($table, $field)) {
+            $dbman->add_field($table, $field);
+        }
+        
+        upgrade_mod_savepoint(true, 2020052100, 'zoom');
```
jrchamp, line 443:
```suggestion
        upgrade_mod_savepoint(true, 2020071500, 'zoom');
```

lib.php
```diff
@@ -565,3 +566,33 @@ function mod_zoom_get_fontawesome_icon_map() {
...
+ * @return bool True if completed, false if not, $type if conditions not set.
+ */
+function zoom_get_completion_state($course, $cm, $userid, $type) {
+    global $CFG, $DB;
+
+    // Get zoom details
+    $zoom = $DB->get_record('zoom', array('id' => $cm->instance), '*', MUST_EXIST);
+
+    // If completion option is enabled, evaluate it and return true/false 
+    if($zoom->completionjoin) {
```
jrchamp, line 587:
```suggestion
    if ($zoom->completionjoin) {
```

lib.php
```diff
@@ -565,3 +566,33 @@ function mod_zoom_get_fontawesome_icon_map() {
...
+        $logsql = "SELECT COUNT(1) 
+                     FROM {logstore_standard_log} l 
+                    WHERE component = 'mod_zoom' AND action = 'clicked'
+                      AND target = 'join_meeting_button' 
+                      AND objectid = :zoomid AND userid = :userid";
+        return $DB->get_field_sql($logsql, array('userid'=>$userid,'zoomid'=>$zoom->id));
```
jrchamp, line 593:
```suggestion
        return $DB->get_field_sql($logsql, array('userid' => $userid, 'zoomid' => $zoom->id));
```

loadmeeting.php
```diff
@@ -72,4 +72,10 @@
 // Record user's clicking join.
 \mod_zoom\event\join_meeting_button_clicked::create(array('context' => $context, 'objectid' => $zoom->id, 'other' =>
         array('cmid' => $id, 'meetingid' => (int) $zoom->meeting_id, 'userishost' => $userishost)))->trigger();
+// Update completion state
+$completion = new completion_info($course);
+if($completion->is_enabled($cm) && $zoom->completionjoin) {
```
jrchamp, line 77:
```suggestion
if ($completion->is_enabled($cm) && $zoom->completionjoin) {
```

loadmeeting.php
```diff
@@ -72,4 +72,10 @@
...
+    $completion->update_state($cm,COMPLETION_COMPLETE);
```
jrchamp, line 78:
```suggestion
    $completion->update_state($cm, COMPLETION_COMPLETE);
```

db/upgrade.php
nstefanski: @jrchamp good catch, you are correct, this should be the new plugin version

nstefanski:
@rlorenzo do you think this can be integrated into the Zoom activity plugin as is, or would you like to see any changes, like a new DB table?

rlorenzo:
@nstefanski Since this queries the log table, how often is the custom completion called? It can be a performance concern especially since our log tables are in the millions.

But also, we purge our logs every 90 days or so.. Would losing log records make an activity now not complete? Or does it query once and then it caches it?

I also took a look at the docs: https://docs.moodle.org/dev/Activity_completion_API#Completion_on_grade

Why wouldn't "Completion on grade" or "Completion on view" work? We set a grade here if a user made this meeting a graded activity: https://github.com/ucla/moodle-mod_zoom/blob/master/loadmeeting.php#L53

In the same loadmeeting.php file can't you set this for "Completion on view"?:

$completion = new completion_info($course);
$completion->set_module_viewed($cm);

For "Completion on grade", is should work with the grade feature already.

nstefanski:
@rlorenzo you're right, I was too quick to dismiss Completion on View as an option. I'm closing this request, I opened a new request #118

PR #120, opened 2020-08-04
Create composer.json

michaelmeneses:


rlorenzo:
@michaelmeneses Thanks! We don't use composer for our project, but if this works for you I am sure it is good to go.

PR #123, opened 2020-08-22
webinar: Allow passcodes

### LINKED ISSUE #107: Feature Request: Allow to password protect webinars
Currently the plugin allows to use a password when creating a meeting. Also if the zoom account allow to embed the password in the link, the password is included in the plugin link in Moodle.

But the plugin currently not support to put a password to a webinar which could be a security problem.

The Zoom API currently support it: https://marketplace.zoom.us/docs/api-reference/zoom-api/webinars/webinarcreate

Thanks!

jrchamp:
I believe these are the only barriers, but it would be good to have someone with a webinar confirm.

Side note: With the rebranding of "passwords" to "passcodes", should we update the two affected language strings?

Fixes #107


rlorenzo:
@jrchamp I created a separate ticket for us to rename Passwords to Passcodes, but for now we will test your patch as it is

Issue #126, opened 2020-09-08
Zoom received a bad request: {$a}


JF10R:
The error happens on Meeting creation when using the Zoom Meeting activity in a Moodle course.

Moodle 3.8.4+

## Plugin configuration

Plugin version: v3.2 (2020090400)

 * Added JWT/API Key and Secret in default config.
 * Using `https://zoom.us` as the home page URL.

## Error details

    Error: Zoom received a bad request: {$a}

![image](https://user-images.githubusercontent.com/2191497/92478684-8409e300-f1b0-11ea-92b3-70d7504151ac.png)

rlorenzo:
Fixed in https://github.com/ucla/moodle-mod_zoom/pull/127

JF10R:
I'm not sure if I'm missing anything here, but I installed the 3.3 version of the plugin and still cannot have a detailed error message:
![image](https://user-images.githubusercontent.com/2191497/93255628-2d7c4600-f768-11ea-881a-7c48609733fd.png)

![image](https://user-images.githubusercontent.com/2191497/93255734-500e5f00-f768-11ea-9ca2-9f04a644ade2.png)

Happens on Zoom Meeting activity creation on Moodle 3.8.4+

rlorenzo:
@jrchamp I merged in your changes for #127 (https://github.com/ucla/moodle-mod_zoom/commit/e5d513cd3f4cde6384a3dbe05d4a41c9995670d3), so using the v3.3 of the plugin should expand that error message so that you can see what $a says for better debugging. Don't know why it isn't working...

@JF10R This plugin is designed to work for Educational or Business Zoom accounts. So the homepage URL shouldn't be https://zoom.us. It should be https://<CUSTOM DOMAIN>.zoom.us

JF10R:
@rlorenzo  Thank you for the heads up. I added the correct subdomain (https://uqat.zoom.us) and still see the $a variable instead of a full error message. I'll try debugging this step by step when I get the chance...

jrchamp:
I don't know why it didn't work. 🤷 I tried adding a unit test, but the unit test succeeded. 😞

PR #127, opened 2020-09-08
bad_request_exception: Include response message

jrchamp:
#126 highlighted that the response message in the text string wasn't being passed by the exception constructor.

This doesn't solve that issue, but it should add a bit more context when the request fails.


jrchamp:
To clarify, this fixes the part where the error message shows "Error: Zoom received a bad request: **{$a}**" instead of included the error message.

PR #129, opened 2020-09-09
Fix for issue with rate limiting. The URLs from subsequent requests w…

kbowlerarden:
…ere being concatenated, making the requests return no data. This fix simply preserves the $url parameter by creating a new local variable ($zoom_url) then appending the $url variable to it, and then uses that in the request.

### DIFF
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -167,7 +167,7 @@ class mod_zoom_webservice {
      */
     protected function _make_call($url, $data = array(), $method = 'get') {
         global $CFG;
-        $zoom_url = self::API_URL . $url;
+        $zoomurl = self::API_URL . $url;
         $method = strtolower($method);
         $proxyhost = get_config('mod_zoom', 'proxyhost');
         $cfg = new stdClass();
@@ -206,7 +206,7 @@ class mod_zoom_webservice {
             $curl->setHeader('Content-Type: application/json');
             $data = is_array($data) ? json_encode($data) : $data;
         }
-        $response = $this->_make_curl_call($curl, $method, $zoom_url, $data);
+        $response = $this->_make_curl_call($curl, $method, $zoomurl, $data);
 
         if ($curl->get_errno()) {
             throw new moodle_exception('errorwebservice', 'mod_zoom', '', $curl->error);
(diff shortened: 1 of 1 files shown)


classes/webservice.php
```diff
@@ -167,7 +167,7 @@ protected function _get_curl_object() {
      */
     protected function _make_call($url, $data = array(), $method = 'get') {
```
jrchamp, line 168:
```suggestion
    protected function _make_call($path, $data = array(), $method = 'get') {
```

classes/webservice.php
```diff
@@ -167,7 +167,7 @@ protected function _get_curl_object() {
...
         global $CFG;
-        $url = self::API_URL . $url;
+        $zoomurl = self::API_URL . $url;
```
jrchamp, line 170:
```suggestion
        $url = self::API_URL . $path;
```

classes/webservice.php
```diff
@@ -206,7 +206,7 @@ protected function _make_call($url, $data = array(), $method = 'get') {
             $curl->setHeader('Content-Type: application/json');
             $data = is_array($data) ? json_encode($data) : $data;
         }
-        $response = $this->_make_curl_call($curl, $method, $url, $data);
+        $response = $this->_make_curl_call($curl, $method, $zoomurl, $data);
```
jrchamp, line 209:
```suggestion
        $response = $this->_make_curl_call($curl, $method, $url, $data);
```

jrchamp (commented):
First, I like the idea of not overwriting the variable that has been passed in.
Second, because strings are passed by value and not pass-by-reference, I don't see how this has any net effect unless the calling method is incorrectly using an ampersand.
Last, it's clear now that the first parameter is not a URL, but is instead part of the URL path.

I've made some suggestions to address the first and last issues, but due to restrictions in the GitHub review process, the function definition will also need to be updated to reflect the "$url" to "$path" correction.

kubilayagi:
@kbowlerarden Thanks for finding this! I created a branch with your fix and @jrchamp 's naming suggestions. I am going to make the unit tests more robust to double check that we are using the correct url

kubilayagi:
added in with https://github.com/ucla/moodle-mod_zoom/tree/feature/CCLE-9507-zoom-fix-url-append

rlorenzo:
Merged in https://github.com/ucla/moodle-mod_zoom/commit/8d9f8729be64cd015ef028bb734b39819f05a81a

PR #135, opened 2020-09-16
tests: Bad request message

jrchamp:
Initially, I'm just trying to use GitHub to run the test for me.


jrchamp:
The test worked, so I'm not sure why they were still seeing {$a}...

Issue #121, opened 2020-08-19
Password


mhughes2k:
Am i right in thinking that this plugin doesn't actually show the meeting password to participants?

![image](https://user-images.githubusercontent.com/694841/90642799-c0d66000-e22a-11ea-9cb0-785eb33b663c.png)

jrchamp:
I think you are correct. It's also worth noting that they rebranded it to "Passcode" because it's not a username/password.

mhughes2k:
Also, with the embed pass(word|code) in link option being a settable at various levels (so for instance in our case it's turned off by default), shouldn't the plugin:
* display it in the UI, rather than having to get the staff to copy & paste it in to the description
* allow the host to set whether or not it is embeded (based on if its allowed) so that if it is not locked they can select it themselves.

jrchamp:
Yes, I do think that it would be good to display the passcode in the UI.

I don't think the Moodle plugin has the ability to embed if it's not already embedded by Zoom because the format is different. Maybe it's something simple like base64, but I haven't looked into the format.

Soap box: Embedding the password is what I recommend, because there's not security benefit to not embedding and it just makes the user experience worse.

mhughes2k:
> Soap box: Embedding the passcode is what I recommend, because there's no security benefit to not embedding and it just makes the user experience worse.

I do agree with this too, but you know...decisions. 

Hopefully with people being more aware of the other settings then it's less of any avenue for all the bad things that were happening because the password was the *only* measure.

mhughes2k:
@rlorenzo i don't see a problem with obscuring the direct link, the problem is not showing the passcode and having to manually communicate it. 

For instance if the feature is disabled, you still get the link and passcode on the same email just as separate pieces of text. 

With a large number of meetings being set up and staff needing to also remember to copy the passcode into the description for the students, it would just be a step that could be eliminated.

ndunand:
I second @jrchamp and @mhughes2k on not displaying the password to students, which really is a problem for us and forces us to manually communicate passwords for each Zoom activity.

@rlorenzo could there be a site-wide option that would allow a display logic like this:

```
if ($haspassword && ($userishost || $showpasswordtostudents)) { 
     $table->data[] = array($strpassword, $zoom->password); 
 }
```

mhughes2k:
Pull request for this #139

rlorenzo:
Merged https://github.com/ucla/moodle-mod_zoom/pull/139

PR #139, opened 2020-09-23
Setting to display passwords (#121)

mhughes2k:
I've used @ndunand's snippet from #121 , and added the settings page setting for this.


rlorenzo:
@mhughes2k Rather than have a site level setting, do you think it would be better to have the setting be per meeting? But have the default value be assigned at the site level?

settings.php
```diff
@@ -143,4 +143,11 @@
             get_string('option_mute_upon_entry', 'mod_zoom'),
             '', 1, 1, 0);
     $settings->add($defaultmuteuponentryoption);
+
+    // TODO Need a heading
+    $settings->add(new admin_setting_heading('othersettings', 'Other Settings', ''));
```
rlorenzo, line 148: @mhughes2k No need for the separate heading... especially if this will be a meeting setting. It will be like the other settings.

mhughes2k: > @mhughes2k Rather than have a site level setting, do you think it would be better to have the setting be per meeting? But have the default value be assigned at the site level?

I thought about this, from our perspective we're sort of looking at things being consistent across the service. Given that the necessity of this is driven by the systems level configuration of Zoom I didn't feel that it needed be per-meeting.

settings.php
```diff
@@ -143,4 +143,10 @@
             get_string('option_mute_upon_entry', 'mod_zoom'),
             '', 1, 1, 0);
     $settings->add($defaultmuteuponentryoption);
+
+    $settings->add(new admin_setting_heading('othersettings', 'Other Settings', ''));
```
rlorenzo, line 147: @mhughes2k If you want to keep this heading, can you move this to the string file? Also, rename it to fit Moodle's coding standard: Other settings (Sentence casing)

mhughes2k: Removed, but tweaked language pack strings for the setting

mhughes2k:
How do you guys deal with the translations? I can add empty strings for the other non-english language packs...

rlorenzo:
@mhughes2k Don't touch the other lang files. The proper way to do translations is to actually use https://lang.moodle.org

mhughes2k:
Cool, wasn't sure if you were doing them in-house/by hand.

lang/en/zoom.php
```diff
@@ -51,6 +51,8 @@
 $string['connectionstatus'] = 'Connection status';
 $string['defaultsettings'] = 'Default Zoom settings';
 $string['defaultsettings_help'] = 'These settings define the defaults for all new Zoom meetings and webinars.';
+$string['displaypassword'] = 'Display Passcode';
```
rlorenzo, line 54: @mhughes2k Can you change this to "Display passcode"? This is to fit coding standards: https://docs.moodle.org/dev/Coding_style#Capitals

rlorenzo:
@mhughes2k Thanks for the patch!

PR #141, opened 2020-09-28
Fix to upgrade.

dvdcastro:
Hello,

This plugin is failing to pass DB schema checks.

```
php admin/cli/check_database_schema.php
```

It's throwing this error:
```
zoom
 * column 'option_host_video' has default '0', expected '1' (I)
 * column 'option_participants_video' has default '0', expected '1' (I)
 * column 'option_mute_upon_entry' has default '1', expected '0' (I)
```

To fix it, I've matched `install.xml` with the values set in `upgrade.php`


jrchamp:
It seems like your database schema is different that the default. Any chance your team changed the defaults at the database level?

jrchamp:
Side note: It would be nice if these settings were fetched from Zoom because individual users and organizations will have their own default settings.

rlorenzo:
Thanks @dvdcastro!

PR #143, opened 2020-09-30
schedule for: only show if can schedule for others

### LINKED ISSUE #137: Flaw in logic testing
https://github.com/ucla/moodle-mod_zoom/blob/35b51ee2f2928181f335f739b0ca9104df5f3377/mod_form.php#L71

$canschedule is never going to be empty! 

1. get_schedule_for_users() will *always* return an array.
2. if the array returned is empty, the first thing that gets done is to add an element to it, so it' can *never* be empty.

jrchamp:
I moved the "self" inside the empty check, which should address the issue. I also adjusted the documentation and code order to move a variable initialization closer to where it is used.

@mhughes2k can you confirm this works for you?

Fixes #137

### DIFF
--- a/mod_form.php
+++ b/mod_form.php
@@ -64,17 +64,14 @@ class mod_zoom_mod_form extends moodleform_mod {
         // can add Zoom meetings, and the user can schedule.
         $scheduleusers = [];
 
-        // This will either be false (they can't) or the list of users they can schedule.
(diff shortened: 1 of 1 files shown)


rlorenzo:
@mhughes2k Does this work for you? Let me know if you can test so I can merge this in

Issue #79, opened 2020-03-31
Links being changed 


mhughes2k:
We seem to have an odd behaviour with respect to the description / intro field.

We have put an HTML link into the field (as a URL and then the text linked as an <a> tag) and 24 hrs later the only thing that remains is the URL text (the <a> has been removed).

Is it possible that the update_meetings task is stripping out HTML tags?

M

rlorenzo:
@mhughes2k I verified this. There is a call to strip tags: https://github.com/ucla/moodle-mod_zoom/blob/master/classes/webservice.php#L367

So when we sent data to the web service the tags are stripped out of the description. Then when update_meetings is called, the new description is pulled from Zoom with the stripped tags and overwrites local description.

We will try removing strip_tags and see if that breaks anything.

mhughes2k:
We've had a supplementary case come in and I was wondering if something else might now be happening with the Zoom description, as I thought from the details here the description was pulled from Zoom and had all of the call details including the password. 

This report however would seem to indicate that it doesn't...

> When I create a Zoom activity for my class and choose the option to set a password, it appears that it does not display the password to the students via the activity link in MyPlace. This is despite the "Student View" showing that the password is displayed. Therefore, the "Student View" is not accurate. This has caused confusion and problems with students joining a session. If it had been clear that the password was NOT available in this way, then it could have been mailed out in advance. 

Could there be some other change to the details returned by Zoom for the description?

mhughes2k:
The line seems to now be https://github.com/ucla/moodle-mod_zoom/blob/eaa1bb36e625a1f4e5f06acdf0c4cd71f775bf3d/classes/webservice.php#L471

PR #144, opened 2020-10-03
Update webservice.php

mhughes2k:
This implements the discussed removal of Strip_tags() from the web service call so that a plain text description isn't sent to Zoom and then subsequently pulled back in plain text.


mhughes2k:
See #79 for issue details

PR #157, opened 2020-11-03
Testing for Moodle 3.10 and master

rlorenzo:


jrchamp:
It looks like the `- env` lines that do not have a php version line immediately above it are using php 7.2

jrchamp:
That looks right now. 😄

PR #163, opened 2020-11-12
Fixes #161

### LINKED ISSUE #161: Bug: Calendar icon in non-Boost theme not getting loaded
The calendar icon on view.php isn't getting loaded in non-Boost themes, it seems. Maybe it's just non-Font-Awesome-using themes, I don't know.

Anyway, the reason seems to be that you explicitly want the i/calendar icon from the plugin mod_zoom, but the calendar icon is Moodle core.

https://github.com/ucla/moodle-mod_zoom/blob/366bffafb3b8faf41bf25ac6951972e0d0655b53/view.php#L157

Fix:

`$calendaricon = $OUTPUT->pix_icon('i/calendar', get_string('calendariconalt', 'mod_zoom'));`

Without the fix, the generated image url is something like:

> https://moodle.instance/theme/image.php/theme_name/mod_zoom/1605014640/i/calendar

Which of course leads nowhere.

danowar2k:
i/calendar is a Moodle core icon. mod_zoom doesn't have one, so the component being 'mod_zoom' is wrong

### DIFF
--- a/view.php
+++ b/view.php
@@ -154,7 +154,7 @@ if ($iszoommanager) {
 // Generate add-to-calendar button if meeting was found and isn't recurring.
 if (!($showrecreate || $zoom->recurring)) {
     $icallink = new moodle_url('/mod/zoom/exportical.php', array('id' => $cm->id));
-    $calendaricon = $OUTPUT->pix_icon('i/calendar', get_string('calendariconalt', 'mod_zoom'), 'mod_zoom');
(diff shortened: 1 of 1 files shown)


danowar2k:
Did this using the github UI, therefore the branch name might be not that good.

danowar2k:
So, how about the patch? Is it good? Do I have to do something else to get it approved?

PR #166, opened 2020-11-17
New Community of enquiry analytics classes.

### LINKED ISSUE #165: Add analytics classes
It would be nice if we could get information back from Zoom's API on how much time participants spend speaking so we could adjust the indicator based on their participation, but it doesn't look like that's possible:
https://devforum.zoom.us/t/summary-of-each-participants-total-time-speaking/34063

In the meantime a simple indicator like the one BBB have written would be a nice "first version".

I'll probably submit a PR for this at some point in the next few weeks.

danmarsden:
Fixes #165

really basic "first version" of the community of enquiry model, in future it might be nice to write some new indicators based on duration of participation, but those would be new indicators rather than extensions to the community of enquiry ones.

to extend the community of enquiry one we'd need to be able to see the difference between viewing the stream and "participation" - eg - a user that speaks would have a higher social breadth.


rlorenzo:
@danmarsden  I am not sure how these new analytics get triggered as being satisfied, but I will merge since you submitted. I will accept any more pull requests regarding this since you have a lot more experience on this topic.

We don't even have analytics enabled on our Moodle.

danmarsden:
haha - thanks @rlorenzo - I'd reccomend you take a look at some point - it's been massively valuable during covid to help identify students that may need some extra support - particularly when the reporting is managed centrally by someone specific within the organisation who is able to reach out and make direct contact with identified students. Quite a different approach from general reports where you don't really know what is important for what group of students.

Issue #48, opened 2019-02-26
Feature request: Include upcoming meetings in users' Timeline


matturn:
It would be good if upcoming Zoom meetings were included in enrolled users' Timelines on their Dashboard.

The ability for Moodle to send a reminder message to enrolled users would be good too.

Both might be options with a default set to "on".

RadBoris:
Hi @rlorenzo, 

I can submit a PR by early next week that solves the timeline display issue for upcoming meetings but wanted to confirm that sending reminders is also something that you would like to be included. If yes, I will need a bit more time.

rlorenzo:
@RadBoris We don’t use timeline in our Moodle so whatever you feel most users would want or fine useful is fine.

rlorenzo:
This has now been implemented in https://github.com/ucla/moodle-mod_zoom/pull/169

PR #169, opened 2020-11-24
Timeline events

nstefanski:
This is a simple fix for issue #48, adding all Zoom meetings to the timeline block by modifying the event type (https://docs.moodle.org/dev/Calendar_API#Action_events). I did not, however, add a reminder feature.


rlorenzo:
@nstefanski We don't use the timeline block, but the code looks solid. Merging in.

PR #170, opened 2020-12-02
Add Zoom meeting invitation to the Meeting Page

jackvanb:
The meeting invitation is shown when the user clicks the 'Show meeting invitation' link, and is hidden when the user clicks the 'Hide meeting invitation' link.

show-invite-text-revised


amd/src/toggle_text.js
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Toggles text to be shown when a user hits 'Show More' and
+ * hides text when user hits 'Show Less'
+ *
+ * @package    mod_zoom
+ * @copyright  2020 UC Regents
+ * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
+ */
+
+export const init = (showMoreText, showLessText) => {
```
rlorenzo, line 25: @jackvanb Rather than passing in the strings as parameters, try using the core/str library. See the example at https://docs.moodle.org/dev/Javascript_Modules#.22Hello_World.22_I_am_a_Javascript_Module

import Str from 'core/str';

Then in your code you can get the string for the buttons like: 

Str.get_string('meeting_invite_show', 'mod_zoom');

view.php
```diff
@@ -70,6 +70,9 @@
     $showrecreate = zoom_is_meeting_gone_error($error);
 }
 
+$response = $service->get_meeting_invitation($zoom->meeting_id);
+$meetinginvite = $response->invitation;
```
rlorenzo, line 74: @jackvanb Seems we will always be calling $response->invitation everytime we call get_meeting_invitation().

I see in the API documentation that it only returns one item in the object: https://marketplace.zoom.us/docs/api-reference/zoom-api/meetings/meetinginvitation

So to simplify things, can you make get_meeting_invitation() return $response->invitation so subsequent calls don't need to do the same thing? The get_meeting_invitation() method should just return the invitation text and not an object with the invitation as a member variable.

rlorenzo (changes requested):

amd/src/toggle_text.js
jackvanb: Done

view.php
jackvanb: Done

amd/src/toggle_text.js
```diff
@@ -22,16 +22,18 @@
...
+
+export const init = () => {
   const button = document.querySelector("#show-more-button");
   const body = document.querySelector("#show-more-body");
-  button.addEventListener("click", () => {
+  button.addEventListener("click", async () => {
     if (body.style.display == "") {
       body.style.display = "none";
-      button.innerHTML = showMoreText;
+      button.innerHTML = await getString("meeting_invite_show", "mod_zoom");
```
rlorenzo, line 33: @jackvanb Wow.. each call to getString will be its own web service call? That looks to be inefficient. Can you change it back to what you had before by passing in the string as a parameter?

jackvanb: @rlorenzo Looks like there's a cache, so the promise will return immediately after the first time. So this should be equivalent to doing it the other way.

https://github.com/ucla/moodle/blob/master/lib/amd/src/str.js#L33

rlorenzo: @jackvanb Let's try it. Put it on TEST, but if there's a delay when clicking on the button it is better if we just pass it the strings right away

styles.css
```diff
@@ -4,6 +4,11 @@
     border-radius: 0.25em;
 }
 
+#show-more-button {
```
jrchamp, line 7: This doesn't say zoom in it and might apply to all "show more button"s anywhere this is installed.
```
#show-more-button {
```

rlorenzo: @jackvanb Use the existing Bootstrap spacer utility classes. See https://getbootstrap.com/docs/4.0/utilities/spacing/

I found that adding the class "pl-0" for padding-left zero removed the extra space when added to the button.

jackvanb: Switched to adding 'pl-0' to avoid using css rules.

rlorenzo (approved):
@jackvanb Looks good. Please squash and merge to development so we can review on TEST

PR #177, opened 2021-01-08
Support more advanced passcode requirements

jackvanb:
Adds the ability to support advanced passcode requirements defined by the [Zoom API](https://marketplace.zoom.us/docs/api-reference/zoom-api/users/usersettings). 

When the advanced passcode requirements are present, a default password is created that satisfies the requirements (having a minimum length, letters, numbers, special characters, etc.). Additionally, a description for the password requirements is created and put underneath the passcode form entry.


locallib.php
```diff
@@ -400,3 +400,73 @@ function zoom_get_participants_report($detailsid) {
...
+    // Fill in the rest of the passcode with numbers.
+    if ($meetingpasswordrequirement->length) {
+        $remlength = $meetingpasswordrequirement->length - strlen($passcode);
+        if ($remlength > 0) {
+            for ($i = 0; $i < $remlength; $i++) {
+                $passcode .= strval(rand(0, 9));
+            }
+        }
+    } else {
+        $passcode .= strval(rand(100000, 999999));
```
jrchamp, line 431: 11% more random:
```suggestion
        $passcode .= str_pad(strval(rand(0, 999999)), 6, '0', STR_PAD_LEFT);
```

mod_form.php
```diff
@@ -198,15 +198,31 @@ public function definition() {
...
-        $mform->addElement('text', 'meetingcode', get_string('password', 'zoom'), array('maxlength' => '10'));
-        $mform->setType('meetingcode', PARAM_TEXT);
-        // Check password uses valid characters.
-        $regex = '/^[a-zA-Z0-9@_*-]{1,10}$/';
-        $mform->addRule('meetingcode', get_string('err_invalid_password', 'mod_zoom'), 'regex', $regex, 'client');
-        $mform->setDefault('meetingcode', strval(rand(100000, 999999)));
+
+        // Set default passcode and desciption from Zoom security settings.
+        $securitysettings = $service->get_user_security_settings($zoomuser->id);
+        if (isset($securitysettings->meeting_password_requirement)) {
```
rlorenzo, line 204: @jackvanb Why would meeting_password_requirement not be set? In any case, it would be better to fill in a default set of meeting_password_requirement if it doesn't exist.

Have the default set of values be our current default, I believe that is the Zoom default. 

Having get_user_security_settings() return something for meeting_password_requirement would remove the extra if/else case here, simplifying the code and its use.

locallib.php
```diff
@@ -400,3 +400,73 @@ function zoom_get_participants_report($detailsid) {
...
+
+        if ($meetingpasswordrequirement->have_number) {
+            $description .= get_string('password_number', 'mod_zoom') . ' ';
+        }
+        if ($meetingpasswordrequirement->have_special_character) {
+            $description .= get_string('password_special', 'mod_zoom') . ' ';
+        }
+    }
+
+    if ($meetingpasswordrequirement->length) {
```
rlorenzo, line 461: @jackvanb What is the max length a passcode can be for Zoom? In our database we have a limit of 10 characters for the column. Do we need to increase that?

tests/advanced_passcode_tests.php
```diff
@@ -0,0 +1,203 @@
...
+        $passcode = zoom_create_default_passcode($this->zoomdata);
+        $this->assertEquals(strlen($passcode), 6);
+        $this->assertTrue(ctype_digit($passcode));
+    }
+
+    /**
+     * Tests that a password has the given minimum length.
+     */
+    public function test_settings_length() {
+        $data = array('length' => 10,
```
rlorenzo, line 65: @jackvanb All the other tests already test that length is 10. Just for variance, this should test another length value.

rlorenzo (changes requested):
@jackvanb Great unit tests and the implementation is almost perfect. Just some comments from @jrchamp and myself to consider.

locallib.php
```diff
@@ -400,3 +400,73 @@ function zoom_get_participants_report($detailsid) {
...
+        $remlength = $meetingpasswordrequirement->length - strlen($passcode);
+        if ($remlength > 0) {
+            for ($i = 0; $i < $remlength; $i++) {
+                $passcode .= strval(rand(0, 9));
+            }
+        }
+    } else {
+        $passcode .= strval(rand(100000, 999999));
+    }
+    return $passcode;
```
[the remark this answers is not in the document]
jrchamp: Here's the alternative method that I described which still guarantees the presence of non-numeric values but allows for random placement within the passcode.

```php
/**
 * Creates a default passcode from the user's Zoom meeting security settings.
 *
 * @param stdClass $meetingpasswordrequirement
 * @return string passcode
 */
function zoom_create_default_passcode($meetingpasswordrequirement) {
    $length = max($meetingpasswordrequirement->length, 6);
    $random = rand(0, pow(10, $length) - 1);
    $passcode = str_pad(strval($random), $length, '0', STR_PAD_LEFT);

    // Get a random set of indexes to replaces with non-numberic values.
    $indexes = range(0, $length - 1);
    shuffle($indexes);

    if ($meetingpasswordrequirement->have_letter || $meetingpasswordrequirement->have_upper_and_lower_characters) {
        // Random letter from A-Z.
        $passcode[$indexes[0]] = chr(rand(65, 90));
        // Random letter from a-z.
        $passcode[$indexes[1]] = chr(rand(97, 122));
    }

    if ($meetingpasswordrequirement->have_special_character) {
        $specialchar = '@_*-';
        $passcode[$indexes[2]] = substr(str_shuffle($specialchar), 0, 1);
    }

    return $passcode;
}

locallib.php
jackvanb: It has to be less than 32 characters (https://zoom.github.io/api/#update-a-users-password).

If we want to remove the if/else in mod_form mentioned in the above comment, then we will probably need to update the db to hold up to 31 characters. This way we can set the max length of the password input to be 31.

mod_form.php
jackvanb: For UCLA's settings, the meeting_password_requirement object is not returned when calling this endpoint. I figure that is it not included if none of the advanced security options have been selected.

What should be the default value for length? It seems that currently we can set a 1 character password.

locallib.php
rlorenzo: That is for the user's account password, but we are just setting the meeting password. I see in the docs it says this: https://zoom.github.io/api/#create-a-meeting

Password to join the meeting. Password may only contain the following characters: [a-z A-Z 0-9 @ - _ *]. Max of 10 characters.

mod_form.php
rlorenzo: Yes, we have none of the settings enabled, so it should default them to all false if no records are returned. I guess the default length will be 1.

rlorenzo: Since this was requested in https://github.com/ucla/moodle-mod_zoom/issues/160 and https://github.com/ucla/moodle-mod_zoom/issues/88, I was wondering if @haietza or @danowar2k can help us test this feature. We cannot make these changes on our Zoom instance.

locallib.php
jackvanb: Ah, thanks for pointing that out!

tests/advanced_passcode_tests.php
jackvanb: Updated.

mod_form.php
jackvanb: Added the default object if it is not present.

locallib.php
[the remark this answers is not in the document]
jackvanb: This looks great! Really helps to simplify the logic and add more randomness. I updated it to include this function.

rlorenzo (approved):
@jackvanb Looks good. Please merge to development and put this onto our TEST server so we can verify that the default settings work.

rlorenzo:
@jackvanb Re-reading the passcode requirements and having an admin look at the Zoom admin settings, there is no user-level ability to specify passcode requirements. 

It seems that https://marketplace.zoom.us/docs/api-reference/zoom-api/users/usersettings is just passing along the passcode requirements from the account.

So we should be using this setting instead: https://marketplace.zoom.us/docs/api-reference/zoom-api/accounts/accountsettings

Also, since this is for the account and wouldn't change often, it should be cached to prevent too many web service calls from being made.

Here is the caching API for Moodle: https://docs.moodle.org/dev/Cache_API

We have an existing session cache to store the current user's Zoom ID, you can do something similar to store the passcode requirements. See existing cache definition here: https://github.com/ucla/moodle-mod_zoom/blob/master/db/caches.php#L28

We will want it to have a mode of cache_store::MODE_APPLICATION, since the setting will not change between different users.

jackvanb:
> @jackvanb Re-reading the passcode requirements and having an admin look at the Zoom admin settings, there is no user-level ability to specify passcode requirements.
> 
> It seems that https://marketplace.zoom.us/docs/api-reference/zoom-api/users/usersettings is just passing along the passcode requirements from the account.
> 
> So we should be using this setting instead: https://marketplace.zoom.us/docs/api-reference/zoom-api/accounts/accountsettings
> 
> Also, since this is for the account and wouldn't change often, it should be cached to prevent too many web service calls from being made.
> 
> Here is the caching API for Moodle: https://docs.moodle.org/dev/Cache_API
> 
> We have an existing session cache to store the current user's Zoom ID, you can do something similar to store the passcode requirements. See existing cache definition here: https://github.com/ucla/moodle-mod_zoom/blob/master/db/caches.php#L28
> 
> We will want it to have a mode of cache_store::MODE_APPLICATION, since the setting will not change between different users.

Sounds good, will implement.

rlorenzo:
@jackvanb The change to using the system passcode requirements looks good to me. Please merge to development and put this onto TEST.

Issue #193, opened 2021-02-01
Improvement: Better control who is allowed to see join links and dial-in informations within the "Phone/Dial-In info" row  and the iCal download


abias:
Hi Rex,

this has an overlap with #173, however I think it's still separate.

On the Zoom activity instance overview page (/mod/zoom/view.php?id=...), mod_zoom currently offers teachers and students a row labeled "Phone/Dial-In info" which contains a full invitation text including everything to join the meeting  (including meeting link, dial-in information, SIP information and H.323 information). The same invitation text is contained in the iCal file which can be downloaded on the same page.

Both places get this text from the get_meeting_invitation() function (https://github.com/ucla/moodle-mod_zoom/blob/master/classes/webservice.php#L631) which fetches it from the https://marketplace.zoom.us/docs/api-reference/zoom-api/meetings/meetinginvitation Zoom API Endpoint.

The invitation text looked similar to this:

```
Eva Example is inviting you to a scheduled Zoom meeting.

Topic: Zoom Meeting
Time: Mar 15, 2021 06:08 AM London

Join Zoom Meeting https://us02web.zoom.us/j/123456789?pwd=ABCABCACBABCABCABC

Meeting ID: 123 456 789
Passcode: 064981

One tap mobile
+61861193900,,123456789#,,,,*064981# Australia
+61871501149,,123456789#,,,,*064981# Australia

Dial by your location
        +61 8 6119 3900 Australia
        +61 8 7150 1149 Australia
        +61 2 8015 6011 Australia
        +61 3 7018 2005 Australia
        +61 7 3185 3730 Australia
Meeting ID: 123 456 789
Passcode: 064981
Find your local number: https://us02web.zoom.us/u/kcOV0px514 
```

At our university, we are aware that the Moodle-Zoom integration with mod_zoom just works with redirecting the user to the Join URL and that this integration isn't bulletproof against leaking Join URLs to non-course participants. However, we still would like to have more control who is directly seeing the Join URLs and Dial-In information and who is not. Instead, we would like to see users primarily join into the meetings through the Moodle course or the Moodle App.

Against this background, I would like to propose this improvement:

[...]

PR #198, opened 2021-02-10
Add setting to hide "Download iCal" button - Fixes #192

### LINKED ISSUE #192: Improvement: Suppress "Download iCal" button
Hi Rex,

this has an overlap with #173, however I think it's still separate.

On the  Zoom activity instance overview page (/mod/zoom/view.php?id=...), mod_zoom currently offers teachers and students to download an iCal file with an eye-cathing button. By clicking this button, the user is able to download an iCal file for the meeting (including meeting link and dial-in information) and add it to his local calendar application.

At our university, we
a) encourage our users to use the Moodle calendar and subscribe to this calendar in their local calendar application.
b) would want our users to access the meetings through the Moodle course and not with deep links from their calendar.

Against this background, I would like to propose this improvement:
* Add a capability mod/zoom:downloadical to better control who is allowed to download the iCal file
* Grant this capability to the authorized user role (to keep the Status Quo of the plugin)
* Show the button only to users who have this capability
* Hide the button from users who don't have this capability and prevent the download even if such a users directly accesses the URL to download the file

Cheers,
Alex

abias:
This patch adds a capability mod/zoom:downloadical to better control who is allowed to download the iCal file.
The capability is granted to the authorized user role by default to keep the Status Quo of the plugin.

Afterwards, the "Download ical" button is only shown to users who have this capability and hidden from users who don't have this capability.
Of course, the download of the iCal file is prevented even if such a user directly accesses the URL to download the file.

-----

Please note that this patch is based on #206 (as both modify the same files) and should be merged after this PR.


abias:
I missed the lang string for the new capability in the initial implementation. This is now added and the branch is force-pushed.

abias:
I rebased this PR now on top of #206 so that we have just one stream of PR to be merged one after the other. The branch is now force-pushed.

jackvanb:
I don't think this makes the most sense to put in as a capability. Access wise, I can't think of a scenario where you wouldn't want to allow a user to download the Ical but still allow them to view the rest of the data (i.e. the join link and phone-in info).

I see #192, and if you want the option to remove the iCal button it makes more sense to put it as an admin setting. The same with #193.

rlorenzo (changes requested):
This pull request has code from https://github.com/ucla/moodle-mod_zoom/pull/202 co-mingled with it. Please remove the extra code

abias:
> I don't think this makes the most sense to put in as a capability. Access wise, I can't think of a scenario where you wouldn't want to allow a user to download the Ical but still allow them to view the rest of the data (i.e. the join link and phone-in info).

Well, I really thought about the question if I should implement this as admin setting kill-switch similar to my other recent PRs.
But I decided for the capability approach.

Please think about the need when an admin wants to provide the iCal button just for teachers and not for students as teachers should be able to join the meeting directly but students should always come through the Moodle course. This is not possible with an admin setting kill switch.

@rlorenzo , what do you think, should this be changed to an admin setting killl switch or is the capability approach ok?

abias:
> This pull request has code from #202 co-mingled with it. Please remove the extra code

I have rebased and force-pushed this branch after updates in #206, but didn't make any changes yet regarding the scope of this PR.

This PR is just about the commit f0a955d0701cb7bc8efdccfb32a8d7e3115924eb. As soon as the underlying PRs up to #206 are merged into master, this should become clearer.

rlorenzo:
@abias @jackvanb The iCal download has the same info that is in "Phone/Dial-In info". So if there is something preventing one, it should also prevent the other from displaying.

I think we should still keep the iCal feature though. The only reason to disable it would be to suppress the direct meeting link and dial-in info. See https://github.com/ucla/moodle-mod_zoom/blob/ee35c20e72da5a55b67d879401876d02553ca287/exportical.php#L74

We can just add a condition here to prevent the meeting invite text from being shown and still keep the iCal download.

Having an admin-only setting seems very restrictive. Ideally, there would be a meeting setting that says:

"Show Phone/Dial-In info"
Help text: "Disabling Phone/Dial-In info will prevent the direct meeting link and phone/dial-in info from appearing in the iCal download and meeting details page."

That would default to what the admin sets at the site level. 

The new setting would show/hide the meeting invite text from showing in the iCal and Phone/Dial-In info locations.

abias:
@rlorenzo thanks for your replies in https://github.com/ucla/moodle-mod_zoom/pull/198#issuecomment-783683949.

I would like to stress that this PR is only about getting rid of the iCal download button based on the reasoning which I have described in #192. And the question to you was mainly if the approach to use a capability is fine or if you would want a kill switch instead.

This PR is _not_ about restricting the access to the join URLs and dial-in information through the "Phone/Dial-In info" row and the iCal download. This is what #193 is about and where I have already added an implementation approach. We have contracted Moodle partner Catalyst IT to have a look at #193 and there will be another PR for #193, but this will take some more time.

Against I would like to keep things separate and hope that you support this as well.

Looking back at this PR here, I would be interested to hear from you if this could be merged as it is implemented now or if you would like to see some changed.
Additionally, you can surely already comment on #193 if you see things differently.

Cheers,
Alex

rlorenzo:
@abias Why would someone want to remove the iCal download? The only reason would be to suppress the meeting invite info, right?

The meeting iCal file includes the Moodle meeting link and is useful for people to be able to have the meeting on their calendar. This should be changed to have the capability to show/hide the meeting invite text.

abias:
I just rebased this branch onto the latest master and force-pushed this branch . I will comment onto the review feedback later.

rlorenzo:
@abias Don't really like having a capability to control this.. It should be a setting. Like in https://github.com/ucla/moodle-mod_zoom/issues/193

Disallowing iCal download shouldn't be removed. It helps in people keeping a schedule. The problem would be having the call-in info.

abias:
@rlorenzo , so I just changed this PR to a admin setting kill switch instead of a capability. And I have changed to PR to be a discrete PR.

At the point, I am somehow unsure if you support our goals to just get rid of the iCal button with this PR and to streamline to way how teachers and students access meetings. But I would be grateful if you could merge this now as is.

We wil definitely work on #193 after this PR to get more control over the call-in information as well.

PR #197, opened 2021-02-10
Add more headings to admin settings page - Fixes #181, fixes #182

abias:
This patch adds some headings to the admin settings page to cluster the various settings for better understandability.
Additionally, it fixes a small flaw with the identifier of the default settings heading.

PR #200, opened 2021-02-10
Add admin setting to disable webinars - Fixes #188

### LINKED ISSUE #188: Improvement: Don't offer webinar checkbox in Zoom activity instances unless it is really allowed to host webinars
Hi Rex,

Zoom licenses don't come with the possibility to host webinars by default, at least not at our university.

mod_zoom is already checking if the teacher is allowed to host webinars and shows the webinar checkbox in Zoom activity instances as disable if the teacher is not allowed. This is happening on https://github.com/ucla/moodle-mod_zoom/blob/master/mod_form.php#L182-L189.

Unfortunately, as the checkbox is there and disabled, teachers start asking "Hey support, what's a webinar? How can I use it?". This produces some overhead if the organisazation really does not want to provide webinars over Zoom.

To avoid this, I would like to propose this improvement:

Add an admin setting to better control the behaviour of the webinar checkbox in Zoom activity instances. This admin setting will be a dropdown with three values:
1. *Disable webinars* - This will act as a kill switch for webinars so that the webinar checkbox will never be shown, regardless if the teacher has a webinar license or not.
2. *Offer webinars if license is available* - This will show the webinar checkbox if the teacher is allowed to host webinars. If he is not, don't show the checkbox at all.
3. *Always offer webinars regardless of license* - This will be the behaviour as the plugin does it now. 

Cheers,
Alex

abias:
This patch adds an admin setting which allows the admin to control if the option to create a webinar is offered to users during the creation of a meeting or not.
This setting only affects the Moodle activity instance settings. Even if the admin decides to always show the option (which is the default and the Status Quo before this patch), the user will still need a valid license for webinars to host a webinar.

This setting is especially helpful to organizations who don't have webinar licenses or don't want to use them from within Moodle.


abias:
PR branch updated and force pushed due to changes and feedback in #197

abias:
After the feedback in https://github.com/ucla/moodle-mod_zoom/pull/202#issuecomment-780855738, I changed to string constants to int constants on https://github.com/ucla/moodle-mod_zoom/pull/200/commits/10a64a757465b553476ec5011db9911c028a8106#diff-3494b45596f0f648d894a5f22070eaae01f3e0ea664be0430b8d23eb7d93e35aR55-R58.

However, doing an exact comparison on https://github.com/ucla/moodle-mod_zoom/pull/200/commits/10a64a757465b553476ec5011db9911c028a8106#diff-aa9511a4531a05e51ca336401c5f892ca096725e2f94b6fce8dd8f0c9fe3b181R183-R192 is not possible as the plugin settings in $config are of type string and not int.

The branch is now force-pushed.

rlorenzo:
@abias Why have the "Always show webinars"? If the user does have the license to create webinars it shouldn't now. Also, if the case of "organizations who don't have webinar licenses" should already be satisfied by the current implementation, because we query if the user has the license to use webinars before allowing them to create one.

However, the condition of "don't want to use them from within Moodle" is valid.

Can you make this just to enable or disable webinar support? Don't need to have 3 options. Just need:

* Disable webinars
* Show webinar option only if the user has a license to host webinars

rlorenzo (changes requested):
Remove option to "Always show webinar option regardless if the user has a license to host webinars"

abias:
Hey @rlorenzo ,

the difference between the ZOOM_WEBINAR_SHOWONLYIFLICENSE and the ZOOM_WEBINAR_ALWAYSSHOW option is that with ZOOM_WEBINAR_SHOWONLYIFLICENSE, the webinar checkbox is not shown at all if the user does not have the right to host webinars, and with the ZOOM_WEBINAR_ALWAYSSHOW option, the checkbox is shown but is disabled if the user does not have the right to host webinars.

I included ZOOM_WEBINAR_ALWAYSSHOW because a) it's the way like the plugin works _currently_ and b) admins _might_ be interested to show the webinar checkbox as disabled as "advertisement" for their teachers.

But I am happy to remove this option from the PR now as requested.

rlorenzo:
@abias Okay, that's a good point. Let's keep the 3 options.

abias:
> @abias Okay, that's a good point. Let's keep the 3 options.

Oh my, now I have force-pushed the branch with 2 options before I read your message ;)
Then I will revert the changes now.

abias:
Allright, @rlorenzo . The branch is now rebased on the latest master branch and, from my point of view, ready to be merged

rlorenzo (approved):

PR #201, opened 2021-02-12
Improve the UX of the activity instance settings page

abias:
This page improves the UX of the activity instance settings page in several aspects, see the commit message for details. After these changes, it aligns the admin settings page and the activity overview page based on these changes


mod_form.php
```diff
@@ -173,98 +176,138 @@ public function definition() {
...
+        } else {
+            $mform->setDefault('requirepasscode', 1);
+        }
+        $mform->addHelpButton('requirepasscode', 'requirepasscode', 'zoom');
+
+        // Set default passcode and description from Zoom security settings.
         $securitysettings = zoom_get_meeting_security_settings();
         // Add password.
-        $mform->addElement('text', 'meetingcode', get_string('password', 'zoom'), array('maxlength' => '10'));
+        $mform->addElement('text', 'meetingcode', '', array('maxlength' => '10'));
```
jrchamp, line 238: @abias I'm looking at the code and not the webpage, but the meetingcode field should still have the text label for accessibility.

abias: @jrchamp I was mainly looking at the webpage and had the goal to make the passcode widgets more clean and thus more understandable. But you are right, I forgot accessibility.

Due to the limitations of forms lib, I have now introduced another string label for the passcode field. I hope that this matches your expectations.

abias:
I have updated my PR and have force-pushed the branch.

abias:
I have fixed a small flaw that, after the redesign in this PR, the "Host video" and "Participants video" widgets were not disabled when webinar was enabled. This is now working again and the branch is force-pushed.

abias:
I have fixed a small flaw in the language pack which was modified by this patch. The branch is force-pushed.

abias:
I have rebased and force-pushed this branch after updates in #200

lang/en/zoom.php
```diff
@@ -84,6 +86,7 @@
 $string['invalid_status'] = 'Status invalid, check the database.';
 $string['join'] = 'Join';
 $string['joinbeforehost'] = 'Join meeting before host';
```
rlorenzo, line 88: @abias So this 'joinbeforehost' string is no longer used right? Why add 'joinbeforehostenable' when you can change the string for 'joinbeforehostenable'?

lang/en/zoom.php
```diff
@@ -36,9 +36,11 @@
 $string['apisecret_desc'] = '';
 $string['apiurl'] = 'Zoom API url';
 $string['apiurl_desc'] = '';
-$string['audio_both'] = 'VoIP and Telephony';
-$string['audio_telephony'] = 'Telephony only';
-$string['audio_voip'] = 'VoIP only';
+$string['audio_both'] = 'Computer audio and Telephone';
```
rlorenzo, line 39: @abias If you are trying to match the Zoom UI, then the options should say "Telephone", "Computer Audio", "Both"

Screen Shot 2021-02-19 at 11 24 12 AM

rlorenzo (changes requested):
Reviewing the current Zoom "Schedule a Meeting" UI did they remove the "Join before host" option? Should that be removed here as well?

lang/en/zoom.php
abias: @rlorenzo , my intention was mainly to get rid of the term "VoIP" which might not be understood by teachers.

I am aware that the Zoom interface has the terminology from your screenshot, but I am unsure if the Moodle plugin should adapt this wholeheartedly. My concern is that, Zoom on the activity overview page, where these strings are used as well, this would then say:

> Audio options: Both

and the teacher might wonder "Both _what_?"

So, what do you think, can we leave the change as it is contributed now or do you want to match the strings with Zoom completely?

lang/en/zoom.php
abias: No, the string 'joinbeforehostenable' is still used on the Zoom activity overview page and the mobile view. But I wanted to have another string on the activity settings to align the terminology again with the Zoom original.

Would you support this idea?

abias:
I have rebased and force-pushed this branch after updates in #200, but didn't make any changes yet regarding the scope of this PR

lang/en/zoom.php
rlorenzo: @abias I can see just the text both being confusing. Let's just keep what you have now

abias:
> Reviewing the current Zoom "Schedule a Meeting" UI did they remove the "Join before host" option? Should that be removed here as well?

No, that option is still there in my Zoom instance:

grafik

lang/en/zoom.php
rlorenzo: @abias  Okay

lang/en/zoom.php
abias: All right. Thank you.

lang/en/zoom.php
abias: Thank you.

abias:
I have rebased and force-pushed this branch after #200 has been merged without doing any further changes regarding the scope of this PR.

@rlorenzo , if I didn't miss anything, this PR could then be merged now.

rlorenzo (approved):

rlorenzo:
@abias Looks great! Information is better organized. Thanks so much!

PR #202, opened 2021-02-16
Add support for E2E encryption - Fixes #194

### LINKED ISSUE #194: Improvement: Allow teachers to enable E2E encryption within Moodle
Hi Rex,

in the Zoom activity settings, a teacher can currently control many aspects of the meeting to be created, but he can't control E2E encryption yet like he is able when he scheduled a meeting in Zoom directly:

grafik

At our university, we will use encourage Zoom users to use E2E encryption optionally as soon as it is clear that they won't need one of the meeting features which is not possible yet with E2E encryption. And we would be interested to offer this possibility from within Moodle as well.

Against this background, I would like to propose this improvement:

* Add a new meeting instance setting called "Encryption" which has two values: 1. Enhanced encryption and 2. End-to-end encryption.
* Add a useful help icon to this meeting instance setting which especially links to the Zoom docs on https://support.zoom.us/hc/en-us/articles/360048660871-End-to-end-E2EE-encryption-for-meetings.
* Add a new admin setting which lets the admin define the default for this meeting instance setting
* Make this admin setting lockable (similar to the existing "zoom | requirepasscode" setting)

Cheers,
Alex

abias:
This patch adds these improvements:
* An admin setting which acts as a kill switch for this supplementary feature, similar to the kill switch for the webinar feature.
* An admin setting to set the default encryption type for meetings which are initiated from within the Moodle activity.
* A activity instance setting which lets the teacher set the encryption type. This includes a help icon which describes both available types with the strings which are also used in the Zoom webfrontend.
* An additional row on view.php which shows the selected encryption type to the teachers and students before joining the meeting.

Please note:
This patch does _not_ address the mobile view of the plugin as I didn't have the time and experience to apply the same additions to the mobile view like I have done to view.php. 
This should be done by a more knowledgable person in a subsequent patch.


abias:
This PR was updated and force-pushed after changes in #201

abias:
I fixed a flaw which prevented the Zoom meeting from being created when the encryption widget was disabled and force-pushed the branch.

jrchamp:
The defined values for ZOOM_ENCRYPTION_TYPE_* are strings and should probably use exact comparisons `===` or `!==`: https://github.com/ucla/moodle-mod_zoom/pull/202/commits/33dec8d00b237409658098e49ce1156ce809c765

There are 6 new define() calls that are using integers as strings such as `'0'` instead of `0` for Webinar and Encryption "off" values. This is different than the existing code and exact comparisons would not work if the types do not match. I don't know what the best option is here.

abias:
@jrchamp , I have followed your advice and have changed the defined values of `ZOOM_ENCRYPTION_DISABLE`, `ZOOM_ENCRYPTION_SHOWONLYIFPOSSIBLE` and `ZOOM_ENCRYPTION_ALWAYSSHOW` from strings to integers.

However, I do not see where exact comparisons are possible and benefical in this patch. I will be happy to follow your advice if you can tell me where you want to see exact comparisons.

classes/webservice.php
```diff
@@ -534,6 +534,9 @@ protected function _database_to_api($zoom) {
             $data['type'] = $zoom->recurring ? ZOOM_RECURRING_MEETING : ZOOM_SCHEDULED_MEETING;
             $data['settings']['participant_video'] = (bool) ($zoom->option_participants_video);
             $data['settings']['join_before_host'] = (bool) ($zoom->option_jbh);
+            $data['settings']['encryption_type'] = (isset($zoom->option_encryption_type) &&
+                    $zoom->option_encryption_type == ZOOM_ENCRYPTION_TYPE_E2EE) ?
```
jrchamp, line 538:
```suggestion
                    $zoom->option_encryption_type === ZOOM_ENCRYPTION_TYPE_E2EE) ?
```

mod_form.php
```diff
@@ -363,6 +437,18 @@ public function validation($data, $files) {
             }
         }
 
+        // Supplementary feature: Encryption type.
+        // Only validate if the admin did not disable this feature completely.
+        if ($config->showencryptiontype != ZOOM_ENCRYPTION_DISABLE) {
+            // Check if given encryption type is valid.
+            if ($data['option_encryption_type'] != ZOOM_ENCRYPTION_TYPE_ENHANCED &&
+                    $data['option_encryption_type'] != ZOOM_ENCRYPTION_TYPE_E2EE) {
```
jrchamp, line 445:
```suggestion
            if ($data['option_encryption_type'] !== ZOOM_ENCRYPTION_TYPE_ENHANCED &&
                    $data['option_encryption_type'] !== ZOOM_ENCRYPTION_TYPE_E2EE) {
```

view.php
```diff
@@ -197,12 +200,21 @@
 }
 
 if (!$zoom->webinar) {
+    $strwr = ($zoom->option_waiting_room) ? $stryes : $strno;
+    $table->data[] = array($strwwaitingroom, $strwr);
+
+    if ($config->showencryptiontype != ZOOM_ENCRYPTION_DISABLE) {
+        $strenc = ($zoom->option_encryption_type == ZOOM_ENCRYPTION_TYPE_E2EE) ? $strencryptionendtoend : $strencryptionenhanced;
```
jrchamp, line 207:
```suggestion
        $strenc = ($zoom->option_encryption_type === ZOOM_ENCRYPTION_TYPE_E2EE) ? $strencryptionendtoend : $strencryptionenhanced;
```

jrchamp (commented):
@abias These are the exact comparisons that I think are safe.

abias:
All right, thanks @jrchamp . I have applied these refinements, have tested that the patch still does what it is expected to do and have force-pushed the branch.

lang/en/zoom.php
```diff
@@ -112,15 +125,26 @@
 $string['oldmeetings'] = 'Concluded Meetings';
 $string['on'] = 'On';
 $string['option_audio'] = 'Audio options';
-$string['option_authenticated_users'] = 'Only authenticated users';
+$string['option_audio_help'] = 'With this option, you can allow users to call in using Telephone only, Computer audio only or both';
+$string['option_authenticated_users'] = 'Require authentication to join';
+$string['option_authenticated_users_help'] = "Enabling this option requires all attendees to sign in with their authorized zoom account to be able to join the meeting. It does <em>not</em> relate to logging into Moodle in any way.";
+$string['option_authenticated_users_help'] = "Enabling this option requires all attendees to sign in with their authorized zoom account to be able to join the meeting. It does <em>not</em> to logging into Moodle in any way.";
```
jrchamp, line 131: It looks like this line is sort-of duplicate, but this line is missing the word "relate".
```suggestion
```

jrchamp:
Sorry - that comment is probably on the wrong pull request. It's a bit hard to review these with the different sets of changes intermingled.

lang/en/zoom.php
```diff
@@ -112,15 +125,26 @@
...
+$string['option_encryption_type'] = 'Encryption';
+$string['option_encryption_type_help'] = "With this option, you control the level of encryption and privacy of this meeting.\n\n*Enhanced encryption* means that the encryption key is stored in the Zoom cloud.\n\n*End-to-end encryption* means that the encryption key is stored on your local device and no one else can obtain your encryption key, not even Zoom.";
```
rlorenzo, line 133: @abias There should be clear guidance that enabling E2EE disables a bunch of important features, see https://support.zoom.us/hc/en-us/articles/360048660871-End-to-end-E2EE-encryption-for-meetings

rlorenzo (changes requested):
Some suggestions. However, I am unable to independently test this since E2EE is not enabled for our account. @jrchamp Do you have E2EE enabled?

rlorenzo:
@abias Do we need the site admin setting to enable/disable E2EE? Isn't that already done on the Zoom end? It would be confusing to enable E2EE in Moodle and then it is off on Zoom's end. Better to keep settings all in Zoom and query Zoom if the setting is enabled when we can.

lang/en/zoom.php
jrchamp: It may not be future-proof to list all of the "missing features" as some will become available in E2EE over time. Examples (as listed on the linked documentation) from 5.5 include 1:1 chat and reactions.

The key ones that are unlikely to change are phone audio, cloud recording, streaming, live transcription (unless there is a creative workaround, these situations require that Zoom is able to decrypt the traffic on the Zoom servers).

jrchamp:
> Some suggestions. However, I am unable to independently test this since E2EE is not enabled for our account. @jrchamp Do you have E2EE enabled?

E2EE starts as locked off at the top level. We unlocked it, but left it off at the top level. This allowed individual users to turn it "on" which makes the functionality available for their meetings. It does not require that all of their meetings use E2EE.

lang/en/zoom.php
abias: Oh, my bad. That duplicate must have sneaked in during a rebase when I added the missing "relate". I will remove the duplicate line.

lang/en/zoom.php
abias: I think you are both right.
Hosts should be informed that some meeting features are missing, but we also can't hardcode the missing features in the language pack and this is won't be stable forever and might require future changes.

I will modify this help string now in a way that I highlight the fact that things will be missing and point to the Zoom docs for further information:
grafik

Is this ok for you?

In the end, every instutution which is using end-to-end encryption should inform their users about the implications anyway.

lang/en/zoom.php
abias: > Sorry - that comment is probably on the wrong pull request. It's a bit hard to review these with the different sets of changes intermingled.

The comment was on the right PR ;)
My mistake was that I introduced this duplicate line within the E2EE patch

abias:
> @abias Do we need the site admin setting to enable/disable E2EE? Isn't that already done on the Zoom end? It would be confusing to enable E2EE in Moodle and then it is off on Zoom's end. Better to keep settings all in Zoom and query Zoom if the setting is enabled when we can.

@rlorenzo , I would like to give a similar feedback like I have done in #200 : 
I have provided three admin options and I see usage scenarios for each of them:
1. ZOOM_ENCRYPTION_DISABLE = Moodle admin wants to disallow E2EE even if it enabled by the Zoom admin and if the user has enabled it in his Zoom profile as well. This is to avoid any hickups with large teaching meetings where E2EE would be counterproductive.
2. ZOOM_ENCRYPTION_SHOWONLYIFPOSSIBLE = Moodle admin wants to allow the teacher to use E2EE if he wants to use it. This checks if E2EE is already enabled in the user's Zoom profile and only shows the widget if this is true. This is to support teachers who have "discovered" this feature in Zoom already but to avoid to confuse everyone else.
3. ZOOM_ENCRYPTION_ALWAYSSHOW = Moodle admin wants to show the widget to all Moodle users and to promote this feature to all teachers. However, if the teacher did not enable E2EE in his Zoom profile yet, he may have to go there first and enable it before he can use it from within Moodle.

Does this sound reasonable to you?

abias:
I have rebased and force-pushed this branch after updates in #201, removed the duplicate help string which sneaked into this PR and improved the help string for the E2EE setting

rlorenzo:
@abias We don't have E2E enabled on our Zoom, but on the Zoom end can a user enable E2E for the meetings they created in Moodle?

If so, the setting to control access to E2E setting will be ineffective. Just adds confusion in setting up and supporting the tool.

abias:
> @abias We don't have E2E enabled on our Zoom, but on the Zoom end can a user enable E2E for the meetings they created in Moodle?

I am not an expert for the Zoom end, but this is how far I have understood it:
If the Zoom administrator did not enable E2EE globally, a user can't enable it in his Zoom account and can't use it in his meetings.
If the Zoom administrator did not enable E2EE globally, the Moodle option from within this patch won't have any effect as well. The Moodle admin can enable it, but the teacher will never see the E2EE option in Moodle.
And if the Zoom administrator did enable E2EE globalls, the user can always go to the Zoom web frontend and enable E2EE for a meeting which was created from within Moodle. This patch just makes this additional step obsolete and lets the user control E2EE from within Moodle.

> If so, the setting to control access to E2E setting will be ineffective. Just adds confusion in setting up and supporting the tool.

Well, in an ideal world the Zoom admin and the Moodle admin should talk before the E2EE option is enabled and offered in Moodle.

Currently, in my implementation, the default for the new setting is ZOOM_ENCRYPTION_ALWAYSHOW. Would you prefer if I changed the default to ZOOM_ENCRYPTION_DISABLE so that the Moodle admin has to enable E2EE actively and think about the consequences?

abias:
I have rebased this branch after #201 has been merged and have force-pushed the branch. 

@rlorenzo , from my point of view this would be ready to be merged unless you have further change requests for me (especially the admin setting default).

rlorenzo:
@abias Found a minor bug. If the meeting default is End-to-end encryption but my account does not allow E2EE, I still see the encryption options and it is set to E2EE. However, when I save and create the meeting I see the encryption is set to Enhanced Encryption.

The bug is that the encryption option should show that my meeting will be using Enhanced Encryption if my account doesn't allow E2EE regardless of the meeting defaults.

Also, I think the better default for site-wide "Encryption type" would be "Show encryption type chooser only if the user can use end-to-end encryption" to avoid confusion of another setting showing up and people wondering why they cannot enable it.

rlorenzo:
Fixed those problems in https://github.com/ucla/moodle-mod_zoom/commit/33196d5cd80e020f617cbed33e94982c8425c1b4

abias:
Thank you, @rlorenzo , for tackling these remaining problems. I was busy with other things for some days and am happy to see the PR merged.

PR #206, opened 2021-02-18
Make handling hosts more usable

abias:
*Remove maximumchars rule for 'alternative_hosts' activity instance setting*
* This patch removes a maximumchars rule for the 'alternative_hosts' field within the activity instance settings due to complete uselessness.
* This rule was introduced with commit 58e27d8f8b0be0273aefb923662a63add1c1d930, but it was broken from the very beginning as it added a rule to the 'name' field instead of the 'alternative_hosts' field.
Additionally, when the length restriction of the 'alternative_hosts' field was removed in commit 350d0d0ccb0b2338fa15a93ade19a68cdd30966b, this rule was not removed as well but remained in the activity instance settings.

*Make alternative hosts more usable*
* Add the capability mod/zoom:alternativehost to the plugin which defines who is selectable as alternative host when creating a new meeting. This capability is assigned to the editing teacher and non-editing teacher role by default.
* Add a admin setting to let the admin control if he a) wants to completely disable alternative hosts, b) provide a plain text input field for alternative hosts (as it was before and as it remains to be the default), c) provide a user picker for alternative hosts which offers all users who are enrolled into the course, have a Zoom account and have the capability mod/zoom:alternativehost.
* If enabled by the admin, add the user picker with autocompletion to the activity instance settings page
* Improve the wording of the help icon for the alternative hosts activity instance settings.
* If enabled by the admin, improve the presentation of the alternative hosts list on the activity overview page.
* Move the error message which is shown if an alternative host user is not found on Zoom into the gradebook

*Make scheduling privilege more usable*
* Add a admin setting to let the admin control if he wants to show the scheduling privilege option on the activity instance settings page or not
* Add a help icon to the scheduling privilege activity instance setting.
* Change the "Schedule for yourself" entry in the dropdown simply to "Yourself"


abias:
I fixed a flaw with the default admin setting which threw fatal errors in GHA and have force-pushed the branch.

rlorenzo:
@abias Generally I disable adding more and more admin settings. Why would an admin want to disable an alternative host? I don't think that should be an option.

Can't we combine the current implementation of a textbox with adding autocomplete for additional emails? Similar to the way Tags are implemented?

Screen Shot 2021-02-19 at 11 40 08 AM

So someone can type in any email address, but also the dropdown lets them select someone that has the alternative host capability

rlorenzo (changes requested):
Try to keep the commits clean. If this is dependent on another commit, then we should focus on fixing the dependent commit first.

abias:
> Try to keep the commits clean. If this is dependent on another commit, then we should focus on fixing the dependent commit first.

Sure. As my contributions often touch the same files, I can't provide independent PRs.
That's why I maintain a clear stream of PRs and rebase each PR as preceding PRs get merged / changed.

abias:
> @abias Generally I disable adding more and more admin settings. 

Well, there are two types of admins out there: One group which dislikes too many admin settings and wants to keep things simple. And the other group which dislikes hardcoded features and wants to keep things as configurable as possible. I'm in the second group as you can imagine ;)

> Why would an admin want to disable an alternative host? I don't think that should be an option.

Well, it was a no-brainer to implement this option and I think that there might be admins out there who don't want to support this feature if they always have exactly one teacher in the course who always is the host and who just need a simple UI without confusing additional options.

> Can't we combine the current implementation of a textbox with adding autocomplete for additional emails? Similar to the way Tags are implemented?
> 
> Screen Shot 2021-02-19 at 11 40 08 AM
> 
> So someone can type in any email address, but also the dropdown lets them select someone that has the alternative host capability

The reasons why I implemened this autocomplete chooser are these:
1. The teacher should not need to handle email addresses. In some institutions, email addresses are even not self-explaining / guessable.
2. The teacher should not need to produce comma-separated lists and have to make sure that the syntax is right
3. The teacher should not be able to add users who don't have a Zoom account (yet) and run into errors when creating the meeting because of that.
4. The teacher should not be able to add users who are not a member of the course or even users who don't have a Moodle account at all.

The only reason to add additional emails would be that you want to drop the constraint number 4. I am aware that I could tell the autocomplete widget to accept additional values which are not within the suggestion list. However, the teacher would still have to know and input the email addresses and the plugin would have to better deal with the cases that the email address is not valid or the Zoom account does not exist.

I solved this catch-22 by letting the admin choose between the existing widget and the more controlled and comfortable autocomplete widget.

Does this sound reasonable to you? If not, what do you really want to see changed?

abias:
I have rebased and force-pushed this branch after updates in #202 , but didn't make any changes yet regarding the scope of this PR

db/access.php
```diff
@@ -79,4 +79,14 @@
             'manager' => CAP_ALLOW
         ),
     ),
+
+    'mod/zoom:alternativehost' => array(
```
rlorenzo, line 83: @abias Just noticed that this is missing a lang string entry.

Also, to clarify the capability can we call it eligiblealternativehost? Because just having the capability will not automatically add them as an alternative host, right?

rlorenzo:
@abias Okay, those are good reasons. I do like to simplify the number of settings... but I guess it is okay.

Just an edge case question, for this restricted new, autocomplete feature, what happens if someone enters an email on the Zoom end that is not enrolled in the course.

Does this new change show that email address? Or does it ignore it and when the user updates the meeting it will remove the email address that was added on the Zoom end?

Still feel if we can combine both adding emails ad-hoc and from the course users it can better handle all edges.

Also, if the alternative host feature is disabled, the same question on what happens if the alternative host gets added on Zoom's end. Does it get erased?

db/access.php
abias: Yes, this makes sense. I will rename the capability before the next force-push of the branch.

abias:
> Just an edge case question, for this restricted new, autocomplete feature, what happens if someone enters an email on the Zoom end that is not enrolled in the course.

You are right to ask about this edge case. I did not have this in mind.
With the current implementation, an alternative host who is added in Zoom directly and who a member of the Moodle course would appear correctly in Moodle (after the mod_zoom\task\update_meetings task has run).
But an alternative host who is added in Zoom directly but does not a Moodle account or is not enrolled into the Moodle course would be dropped from the meeting silently when it is saved once more in Moodle.

Thinking about this further, I would offer to rework the patch into this direction:
* I will rework the admin setting to support three cases: ZOOM_ALTERNATIVEHOSTS_DISABLE, ZOOM_ALTERNATIVEHOSTS_COURSEMEMBERSONLY, ZOOM_ALTERNATIVEHOSTS_ALLOWADDITIONALEMAILS.
* The first case would be the same as before: Moodle does not allow the teacher to add / modify alternative hosts. Moodle also does neither show nor modify alternative hosts who are set directly in Zoom.
* The second case would offer the autocomplete widget to the teacher as it is implemented now. There would not be a possibility to add additional alternative hosts who are not enrolled in the course or do not have an appropriate Moodle role as it is implemented now. If the host adds an additional host directly in Zoom who is enrolled in the Moodle course, Moodle would pick up this additional host and would handle it just like it would have been added in Moodle.
In addition to that, I would cover the given edge case in this way: If the host adds an additional host directly in Zoom who is not enrolled in the Moodle course, Moodle would pick up this additional host, would store it in Moodle, would make sure that it is not dropped if the Moodle activity is stored again, but would just show this additional host to the teacher as "readonly information".
* The third case would then be the same as the second case with the difference that the teacher could also add additional email addresses directly in Moodle. With this case, we would always have the benefit of the autocomplete widget available but would also support additional flexibility.

How does this sound to you, do you want me to go ahead into this direction?

Thanks,
Alex

rlorenzo:
@abias Thanks for considering the edge case. We had people complain about changes not syncing back to Moodle before.

Please go ahead with the improvements outlined above.

abias:
I just rebased this branch onto the latest master and force-pushed this branch . The discussed changed for this PR are still pending.

abias:
I just rebased this branch onto the latest master and force-pushed this branch . The discussed changed for this PR are still pending.

rlorenzo:
@abias I am making some changes here for wording and placements.

https://github.com/ucla/moodle-mod_zoom/tree/LMS-5432

After taking it over with some people here we just want to leave the alternative host to plain text entry of the picker option is not chosen.

abias:
@rlorenzo , I have incorporated your language changes into my branch and will now implement the remaining discussed improvements.

abias:
@rlorenzo , as you told me that you, at UCLA, would like the keep the existing input field, I went with the easy way and just added the edge case which shows non-Moodle members' email addresses as read-only information, but didn't implement the possibility to add additional mail addresses in addition to the user picker (yet).

I hope this compromise is ok for you.

From my point of view, this would now be ready for final review and merge.

Issue #180, opened 2021-01-19
Differing pluginname prefix for setting zoom | requirepasscode


abias:
Hi all,

I noticed that the plugin admin setting 'zoom | requirepasscode' (see https://github.com/ucla/moodle-mod_zoom/blob/master/settings.php#L106) does not use the same prefix 'mod_zoom' like all other admin settings of the plugin. In fact, it's the only setting which is labeled correctly as for activity modules, the "mod_" prefix is omited for historic reasons.

You might want to fix that in settings.php and provide an upgrade steps in db/upgrade.php for existing installations.

Cheers,
Alex

jrchamp:
Interesting find! I wonder if that's why this property_exists check was added:
https://github.com/ucla/moodle-mod_zoom/commit/12058f39721ca2fc9333a5ba1d29af06af0f05bc#diff-8f5ff72ef1cf59199cfc4bbf16f2475d09e448601242f0d2c51bb94bd4aa1395L124-R128

rlorenzo:
Closing because https://github.com/ucla/moodle-mod_zoom/pull/197 has been merged

PR #207, opened 2021-02-19
Make activity overview page more usable

abias:
*Cluster and improve activity overview page*
* Move the "Meeting not found" notifications below the main page heading
* Split the large meeting information table and add headings between the table rows. This was done similar to what was done before on the activity instance settings page (with the only exception of the host information which was added to the "Schedule" section)
* Improve the appearance of the join button / unavailability notification

*Add admin setting to hide the all meetings link*
This patch adds an admin setting which simply lets the admin decide if he wants to hide or show the all meetings link at the bottom of each Zoom activity overview page.

*Add admin setting to display the lead time*
This patch adds an admin setting to let admins display the leadtime to the users. With this option, users are informed that / when they can join the meeting before the scheduled start time. This might keep users from constantly reloading the page until they can join.

Please note:
This patch does not address the mobile view of the plugin as I didn't have the time and experience to apply the same additions to the mobile view like I have done to view.php.
This should be done by a more knowledgable person in a subsequent patch.


rlorenzo:
@abias I wonder if having the setting to show lead times and meeting links at the site level is a bit restricting. Why not have it at the meeting option? What if some instructors want to show that information?

The site admin configs should be the site defaults, but not the strict setting for every meeting

abias:
> @abias I wonder if having the setting to show lead times and meeting links at the site level is a bit restricting. Why not have it at the meeting option? What if some instructors want to show that information?
> 
> The site admin configs should be the site defaults, but not the strict setting for every meeting

I basically agree with you, @rlorenzo , with regard to the lead times. There might be cases when the teacher would want to selectively disable / enable this information and it's the admins job to provide reasonable defaults.

I could re-implement this as lockable default setting, similar to the requirepasscode setting (see https://github.com/ucla/moodle-mod_zoom/blob/master/settings.php#L135). However, we should then solve #180 first so that the plugin prefixes are all aligned and the set_locked_flag_options() function can do its magic.

What do you think, would this be a good way forward?

In the meantime, I have rebased and force-pushed this branch after updates in #198 , but didn't make any changes yet regarding the scope of this PR

rlorenzo:
@abias Yes a lockable default at the site admin level that allows users to control it at the meeting level is good.

abias:
> @abias Yes a lockable default at the site admin level that allows users to control it at the meeting level is good.

OK. I am afraid I will then have to have a look at #180 first.
Please stay tuned while I tackle this.

rlorenzo:
@abias If you can fix up the commit history and remove 'Add capability to suppress "Download iCal" button - Fixes #192' I can merge this as is

abias:
@rlorenzo , as my other PRs get merged into master more and more and there is no real need to have a stream of PRs anymore for me, I have rebased this on latest master as discrete PR as requested.

I would be really grateful if you could merge this into master as is now as you have proposed.

I am just unsure about your initial feedback that the two settings for the "all meetings" link and the lead time warning should rather be per-activity settings instead of admin kill switches. Did you forget that or did you change your mind?

rlorenzo:
@abias I want a new version pushed out soon. The per activity setting is a nice to have

abias:
@rlorenzo Awesome, thank you very much!

> The per activity setting is a nice to have

All right. Just for the sake of completeness: Regarding our contribution from Ulm university, we would not tackle this per-actitvity setting as we are happy with the kil switch.

PR #214, opened 2021-03-02
Align all plugin settings prefixes - Fixes #180

### LINKED ISSUE #180: Differing pluginname prefix for setting zoom | requirepasscode

abias:
This patch changes all plugin settings prefixes to use the 'zoom' prefix instead of 'mod_zoom'.
To make this transition as transparent as possible for the admin, existing settings are rewritten during an upgrade step.

### DIFF
--- a/classes/task/get_meeting_reports.php
+++ b/classes/task/get_meeting_reports.php
@@ -86,7 +86,7 @@ class get_meeting_reports extends \core\task\scheduled_task {
     public function execute($paramstart = null, $paramend = null, $hostuuids = null) {
         global $CFG, $DB;
 
-        $config = get_config('mod_zoom');
+        $config = get_config('zoom');
         if (empty($config->apikey)) {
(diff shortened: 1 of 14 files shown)


abias:
@rlorenzo , as discussed in #207, I have built this fix for #180 and would be grateful if you could double-check and merge it.

Thank,
Alex

rlorenzo (approved):

PR #216, opened 2021-03-04
Deny direct access if the meeting is not available

abias:
Up to this patch, the check if a meeting is available or not was ineffective. If the user just replaced /mod/zoom/view.php with /mod/zoom/loadmeeting.php in the URL, he could join the meeting anytime. This patch solves this minor security flaw by doing another availability check on loadmeeting.php.
If the meeting is not available, the user is redirected to the activity overview page again.


loadmeeting.php
```diff
@@ -47,6 +47,22 @@
 $PAGE->set_context($context);
 
 require_capability('mod/zoom:view', $context);
+
+// Get meeting state from Zoom.
+list($inprogress, $available, $finished) = zoom_get_state($zoom);
+
+// If the meeting is not yet available, deny access.
+if ($available != true) {
```
jrchamp, line 55:
```suggestion
if ($available !== true) {
```

locallib.php
```diff
@@ -656,3 +656,52 @@ function zoom_get_nonusers_from_alternativehosts(array $alternativehosts) {
...
+
+        // Otherwise we add some more information to the unavailable string.
+    } else {
+        // If we don't have the finished information yet, get it with a small overhead.
+        if ($finished == null) {
+            list($inprogress, $available, $finished) = zoom_get_state($zoom);
+        }
+
+        // If this meeting is still pending.
+        if ($finished != true) {
```
jrchamp, line 689:
```suggestion
        if ($finished !== true) {
```

locallib.php
```diff
@@ -656,3 +656,52 @@ function zoom_get_nonusers_from_alternativehosts(array $alternativehosts) {
...
+ */
+function zoom_get_unavailability_note(object $zoom, $finished = null) {
+    // Get config.
+    $config = get_config('zoom');
+
+    // Get the plain unavailable string.
+    $strunavailable = get_string('unavailable', 'mod_zoom');
+
+    // If this is a recurring meeting, just use the plain unavailable string.
+    if ($zoom->recurring == true) {
```
jrchamp, line 678:
```suggestion
    if ($zoom->recurring === true) {
```

locallib.php
```diff
@@ -656,3 +656,52 @@ function zoom_get_nonusers_from_alternativehosts(array $alternativehosts) {
...
+        $unavailabilitynote = $strunavailable;
+
+        // Otherwise we add some more information to the unavailable string.
+    } else {
+        // If we don't have the finished information yet, get it with a small overhead.
+        if ($finished == null) {
```
jrchamp, line 684:
```suggestion
        if ($finished === null) {
```

locallib.php
```diff
@@ -660,6 +660,55 @@ function zoom_get_nonusers_from_alternativehosts(array $alternativehosts) {
...
+ */
+function zoom_get_unavailability_note(object $zoom, $finished = null) {
+    // Get config.
+    $config = get_config('zoom');
+
+    // Get the plain unavailable string.
+    $strunavailable = get_string('unavailable', 'mod_zoom');
+
+    // If this is a recurring meeting, just use the plain unavailable string.
+    if ($zoom->recurring === true) {
```
jrchamp, line 681: Sorry! It looks like this is coming from the database and may not be a boolean, thus type coercion is probably acceptable. This format mirrors other pre-existing code:
```suggestion
    if ($zoom->recurring) {
```
Alternatively, we could be explicit that we expect type coercion:
```suggestion
    if (!empty($zoom->recurring)) {
```

locallib.php
```diff
@@ -660,6 +660,55 @@ function zoom_get_nonusers_from_alternativehosts(array $alternativehosts) {
...
+    } else {
+        // If we don't have the finished information yet, get it with a small overhead.
+        if ($finished === null) {
+            list($inprogress, $available, $finished) = zoom_get_state($zoom);
+        }
+
+        // If this meeting is still pending.
+        if ($finished !== true) {
+            // If the admin wants to show the leadtime.
+            if ($config->displayleadtime == true && $config->firstabletojoin > 0) {
```
jrchamp, line 694: I would expect that the configuration is a string and not a boolean, so explicit type coercion makes sense here too:
```suggestion
            if (!empty($config->displayleadtime) && $config->firstabletojoin > 0) {
```

jrchamp (commented):
I scanned a bit too quickly. Hopefully this is not too late.

abias:
Hey @jrchamp ,

thanks for your review!
Basically, the code for composing the unavailability note was already merged and I am just moving it within this commit to locallib.php. But I am always happy to make things more secure.

Do I understand it correctly now that you will be directly comming your proposals to the PR? Or should I handle them and update my branch based on your suggestions?

jrchamp:
I think @rlorenzo applied some of the changes directly and is probably trying to get the next release version ready very soon. You might want to wait to see what @rlorenzo does, because the code might be merged soon. My suggestions were mainly to highlight optional changes that caught my eye.

abias:
> I think @rlorenzo applied some of the changes directly

Ah, ok. I have to admit that I am not fully familiar with the change suggestion feature on Github and didn't realize who did commit what.

> You might want to wait to see what @rlorenzo does, because the code might be merged soon. 

OK, I won't update anything unless asked to do so.

PR #224, opened 2021-03-10
String 'join_link' was erroneously hidden from language customization

abias:
This patch overcomes the problem that the string 'join_link' was not translatable within the Moodle language customization by changing the identifier 'joinlink'.

All *_link strings are excluded from translations because they are considered to have special meaning as per https://docs.moodle.org/dev/String_API#Help_strings.
In this case, the given string isn't a special string and thus should not use the special identifier.

Credits go to David Mudrák for analyzing and explaining this problem.


abias:
As a side note:
The AMOSscript in the commit message will not be run automatically in AMOS as this is only possible for Moodle core strings.
But I will ask the AMOS maintainer to run it manually as soon as the PR is merged and released within a new vesion.

PR #229, opened 2021-03-12
Fix a fatal regression on settings.php for Moodle core < 3.7 - Fixes #227

### LINKED ISSUE #227: Version 3.5 not compatible on Moodle versions lower than 3.7
Release 3.5 of this plugin not compatible on any Moodle version lower than 3.7 due to hide_if method not being added until MDL-52167 (https://github.com/moodle/moodle/commit/8ed3671d4cd883bf13754234d339243ec2c4be57) despite plugin version compatibility supporting all versions (https://moodle.org/plugins/pluginversions.php?plugin=mod_zoom), which appears during an upgrade.

Exception - Call to undefined method admin_settingpage::hide_if()

Stack trace:

    line 115 of /mod/zoom/settings.php: Error thrown
    line 89 of /lib/classes/plugininfo/mod.php: call to include()
    line 47 of /admin/settings/plugins.php: call to core\plugininfo\mod->load_settings()
    line 7999 of /lib/adminlib.php: call to require()
    line 22 of /admin/upgradesettings.php: call to admin_get_root()

abias:
This patch fixes a regression which was introduced in c22c108e9c36ebf7e716939a1de8616fed800a41 but which did not handle the fact that the hide_if function does not exist before Moodle core 3.7

### DIFF
--- a/settings.php
+++ b/settings.php
@@ -108,11 +108,20 @@ if ($ADMIN->fulltree) {
             15, $jointimeselect);
     $settings->add($firstabletojoin);
 
-    $displayleadtime = new admin_setting_configcheckbox('zoom/displayleadtime',
-            get_string('displayleadtime', 'mod_zoom'),
-            get_string('displayleadtime_desc', 'mod_zoom'), 0, 1, 0);
-    $settings->add($displayleadtime);
-    $settings->hide_if('zoom/displayleadtime', 'zoom/firstabletojoin', 'eq', 0);
+    if ($CFG->branch >= 37) {
+        $displayleadtime = new admin_setting_configcheckbox('zoom/displayleadtime',
+                get_string('displayleadtime', 'mod_zoom'),
+                get_string('displayleadtime_desc', 'mod_zoom'), 0, 1, 0);
+        $settings->add($displayleadtime);
+        $settings->hide_if('zoom/displayleadtime', 'zoom/firstabletojoin', 'eq', 0);
+    } else {
+        $displayleadtime = new admin_setting_configcheckbox('zoom/displayleadtime',
+                get_string('displayleadtime', 'mod_zoom'),
+                get_string('displayleadtime_desc', 'mod_zoom').'<br />'.
+                        get_string('displayleadtime_nohideif', 'mod_zoom', get_string('firstjoin', 'mod_zoom')),
+                0, 1, 0);
+        $settings->add($displayleadtime);
+    }
 
     $displaypassword = new admin_setting_configcheckbox('zoom/displaypassword',
             get_string('displaypassword', 'mod_zoom'),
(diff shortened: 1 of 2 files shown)


settings.php
```diff
@@ -108,11 +108,20 @@
             15, $jointimeselect);
     $settings->add($firstabletojoin);
 
-    $displayleadtime = new admin_setting_configcheckbox('zoom/displayleadtime',
-            get_string('displayleadtime', 'mod_zoom'),
-            get_string('displayleadtime_desc', 'mod_zoom'), 0, 1, 0);
-    $settings->add($displayleadtime);
-    $settings->hide_if('zoom/displayleadtime', 'zoom/firstabletojoin', 'eq', 0);
+    if ($CFG->branch >= 37) {
```
jrchamp, line 111: $CFG->release seems to be the only machine-safe and human-readable way to check which Moodle version we're on. Moodle adds some extra characters, so they want you to pass it through `normalize_version()` first. Last, the appropriate PHP function for version comparisons is `version_compare()`.

```suggestion
    if (version_compare(normalize_version($CFG->release), '3.7.0', '>=')) {
```

Moodle 3.10 broke the unsafe assumption that the minor version would always be a single digit. The Moodle core code itself makes this assumption here: [convert_branch_numbering_format()](https://github.com/moodle/moodle/blob/6b804f9e1f8271adc9fe64a18fd318ca9537c674/lib/classes/update/api.php#L303-L312), [test_convert_branch_numbering_format()](https://github.com/moodle/moodle/blob/6b804f9e1f8271adc9fe64a18fd318ca9537c674/lib/tests/update_api_test.php#L47-L57)

@mudrd8mz I just noticed this issue with $CFG->branch "parsing" and tests, but I need to run so don't have time to open a tracker.

settings.php
```diff
@@ -108,11 +108,20 @@
...
+        $displayleadtime = new admin_setting_configcheckbox('zoom/displayleadtime',
+                get_string('displayleadtime', 'mod_zoom'),
+                get_string('displayleadtime_desc', 'mod_zoom'), 0, 1, 0);
+        $settings->add($displayleadtime);
+        $settings->hide_if('zoom/displayleadtime', 'zoom/firstabletojoin', 'eq', 0);
+    } else {
+        $displayleadtime = new admin_setting_configcheckbox('zoom/displayleadtime',
+                get_string('displayleadtime', 'mod_zoom'),
+                get_string('displayleadtime_desc', 'mod_zoom').'<br />'.
```
jrchamp, line 120:
```suggestion
                get_string('displayleadtime_desc', 'mod_zoom') . '<br />'.
```

abias:
Thank you for doing these refinements to this PR and merging it.

PR #232, opened 2021-03-15
Fix regression which turned E2E encryption not selectable

abias:
There was a regression in commit 05092c242f669f383825d01e285a6e5e39b07307 which set a parameter type of PARAM_ALPHAEXT for the option_encryption_type parameter. However, the E2E encryption has a parameter value of 'e2ee'. As a result, if a user selected E2E encryption during an activity creation, the form submission was denied and the 'The submitted encryption type is not valid' validation error was given.

This patch fixes this regression by simply changing the parameter type to PARAM_ALPHANUMEXT.


jrchamp (approved):
This makes sense to me. There is a number in `e2ee` and there is an underscore in `enhanced_encryption`, so PARAM_ALPHANUMEXT is somewhat of a minimum.

PR #234, opened 2021-03-19
Invite Text setting

nstefanski:
This addresses issue #193; rather than creating new permissions, it creates an admin setting as in the accepted fix for issue #192


jrchamp:
I'm a little concerned about this comparison: `$config->showinvitationtext != ZOOM_INVITATIONTEXT_DISABLE`. mdl_config entries seem to be strings, so an exact comparison probably won't work unless we use strings throughout (though that might be better). Unfortunately, we have this issue in several places already and this code works as long as the values are `'0'` or `'1'`. However, it is good to remember that in PHP `'a1' == 0` returns true, which has gotten some in hot water recently.

lang/en/zoom.php
```diff
@@ -107,6 +107,10 @@
 $string['indicator:socialbreadth_help'] = 'This indicator is based on the social breadth reached by the student in a Zoom activity.';
 $string['invalidscheduleuser'] = 'You cannot schedule for the specified user.';
 $string['invalid_status'] = 'Status invalid, check the database.';
+$string['invitationtext'] = 'Display invitation text';
+$string['invitationtext_desc'] = 'This setting allows you to choose whether to display the meeting invitation text from Zoom. By default, this text contains call in phone numbers and direct links to meetings in Zoom. If enabled, this text will be displayed on the activity page and in the iCal description.';
+$string['invitationtext_disable'] = 'Disable invitation text';
+$string['invitationtext_enable'] = 'Enable invitation text';
```
jrchamp, line 113: Given that translations of these strings will occur, would "Show/Hide" be a more direct translation?

Also, the use of underscores in the string keys should probably be avoided (see #224).

abias:
Hi @nstefanski ,

thank you for creating this patch similar to the other kill switch implementations which I have done in the last weeks for particular features of mod_zoom.

But as I have just written in #193, Ulm university has contracted Moodle partner Catalyst IT some days ago to implement a solution for #193 based on the specification given there.

In our point of view, a kill switch isn't enough here. Instead, a more fine-granular and configurable control which roles will see what will be helpful for admins.

I would be grateful if this PR could be deferred until our solution to #193 is submitted so that the plugin maintainers could device which solution they prefer.

Thanks,
Alex

nstefanski:
@jrchamp I updated the language strings per your recommendations. On the comparison issue, what would you suggest? I could change the constants to be strings, but that would mean that it would be out of line with the rest of the constants, and this isn't the only commit that uses such a comparison. Perhaps there should be a separate issue created for string to int comparisons?

@abias I'd thought from the conversation on #192 that the maintainers wan't to avoid adding more permissions, but they can certainly hold off on this pull until you've completed yours.

jrchamp:
> Perhaps there should be a separate issue created for string to int comparisons?

I think so. Overall, integers seem like they should make for "cleaner" code, but that doesn't seem to hold true. Considering that the Moodle config values are being stored and returned as strings in addition to the issues with PHP's loose comparisons, using strings throughout would allow for less ambiguity. If I was starting from scratch, I might do something similar to Apache's "On"/"Off" model to avoid the problem completely (avoids the potential risk of a future version of Moodle "helpfully" typecasting integer strings to true integers).

jrchamp:
I'm just now realizing that when #235 is merged, this one will need to be adjusted to resolve merge conflicts. April has been a difficult year. I think the two can work together, providing both technical and administrative controls at multiple levels. Thus, across-the-board and targeted (via capability or Zoom meeting configuration) controls.

abias:
Hi @jrchamp and @nstefanski ,

I am unsure if this needs to be merged at all anymore now that #235 is basically there.
After #235 has landed, the admin can achieve the same result by just retracting mod/zoom:viewdialin from every role.

However, I do not want to have the last say here. It's up to the plugin maintainers to decide.

Thanks,
Alex

rlorenzo:
@nstefanski Closing because of @abias comment above. I hope that solution works.

jrchamp:
@nstefanski If you do decide that you still want Moodle administrators to be able to lock this off globally without setting then to "Prohibit" for all roles, feel free to reopen this and rebase and address the merge conflicts.

PR #235, opened 2021-03-22
Issue #193: Hide/show parts of zoom meeting invitation based on capabilities

andrewmadden:
* Add two new caps: zoom/viewjoinurl, zoom/viewdialin
* Add settings to define regex to pull out key information
* Add toggle to turn this feature on/off
* Remove sections of message user does not have the capability to view
* Add setting to toggle using regex pattern to remove the generic invite sentence.


classes/invitation.php
```diff
@@ -0,0 +1,162 @@
...
+     */
+    public function get_display_string(int $courseid, int $userid = null): string {
+        $displaystring = $this->invitation;
+        try {
+            // If setting enabled, strip the invite message.
+            if (get_config('zoom', 'invitationremoveinvite')) {
+                $displaystring = $this->remove_element($displaystring, 'invite');
+            }
+            // Check user capabilities, and remove parts of the invitation they don't have permission to view.
+            if (!has_capability('mod/zoom:viewjoinurl', \context_course::instance($courseid), $userid)) {
```
abias, line 65: I think, in line 65 and 68, you should check the capability in the module context, not in the course context. Otherwise, you won't cover roles which have just been assigned at the module level.

This would then mean that you would have to pass the module ID to this function, not the course ID.

settings.php
```diff
@@ -275,4 +275,26 @@
             get_string('option_mute_upon_entry_help', 'mod_zoom'),
             1, 1, 0);
     $settings->add($defaultmuteuponentryoption);
+
+    $settings->add(new admin_setting_heading('zoom/invitationregex',
+            get_string('invitationregex', 'mod_zoom'),
+            get_string('invitationregex_help', 'mod_zoom')));
+
+    $settings->add(new admin_setting_configcheckbox('zoom/invitationregexenabled',
+            get_string('invitationregexenabled', 'mod_zoom'),
```
abias, line 284: It's great that you have built a kill-switch setting for this feature here, but the `invitationregexenabled` config isn't used anywhere, especially not in the `invitation` class.

I would like to propose to fix this gap in a way that the `get_display_string()` function in the `invitation` class respects this setting.

If the setting is false, then the plain invitation text which is fetched from Zoom is returned.
If the setting is true, then the processed invitation text is returned.

settings.php
```diff
@@ -275,4 +275,26 @@
...
+            get_string('invitationregexenabled_help', 'mod_zoom'),
+            1, 1, 0));
+
+    $settings->add(new admin_setting_configcheckbox('zoom/invitationremoveinvite',
```
abias, line 288: As this setting is depending on the `invitationregexenabled` setting, I would like to propose to add a hide_if rule to this setting.

However, as this plugin supports older Moodle core versions as well, you will have to check the Moodle core version. Please see https://github.com/ucla/moodle-mod_zoom/blob/master/settings.php#L112-L125 as a blueprint how to realize that.

The same proposal applies to the settings in line 293 to 299.

abias (changes requested):
Hi @andrewmadden ,

thanks for working on this one and for submitting the PR.

I had a look through the code and have made some minor proposals here on Github.
Regarding the general functionality of the patch, some major things seem to be missing or are not functional yet. I will come back to you privately.

@rlorenzo @jrchamp - Please wait with your official review until we have sorted these things out.

Cheers,
Alex

classes/invitation.php
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /**
+     * Get default regex patterns to find the different zoom invitation elements.
+     *
+     * @return string[]
+     */
+    public static function get_default_invitation_regex(): array {
+        return [
+            'invite' => '/^.+is inviting you to a scheduled zoom meeting.+$/mi',
```
rlorenzo, line 156: @andrewmadden  This is only going to work for Zoom that’s in English right? What about Zoom in other languages?

settings.php
andrewmadden: Hide-if settings have now been added for Moodle versions 3.7.0+ with a nohideif message for lower versions.

settings.php
andrewmadden: This has been implemented along with a unit test to ensure the behaviour remains.

classes/invitation.php
andrewmadden: `get_display_message` now expects a course module id to check the module context for the capabilities. Unit tests updated to reflect the change.

settings.php
```diff
@@ -289,12 +293,24 @@
...
         $settings->add(new admin_setting_configtext($name, $visiblename, $description, $pattern));
+        if (version_compare(normalize_version($CFG->release), '3.7.0', '>=')) {
+            $settings->hide_if('zoom/' . \mod_zoom\invitation::PREFIX . $element,
+                    'zoom/invitationregexenabled', 'eq', 0);
+        }
+    }
+
+    // Extra hideif for invite element.
+    if (version_compare(normalize_version($CFG->release), '3.7.0', '>=')) {
+        $settings->hide_if('zoom/invitation_invite', 'zoom/invitationremoveinvite', 'eq', 0);
```
abias, line 314: @andrewmadden , I am unsure what this hide_if rule is there for as there isn't any `invitation_invite` setting in the plugin. Or am I missing something?

classes/invitation.php
abias: @rlorenzo you are right to ask this. But I am wondering if we really have a problem here.

Yes, the RegEx is built and shipped by default for english invitations. 
I am aware that a Zoom user's profile can be changed to another language. Having done that, the invitation which can be retrieved manually for a meeting, is presented in the configured language. Example:

grafik

However, up to now, I have _always_ got an english invitation from the Zoom API, regardless of the language of the Zoom user account and the language of the Moodle user account. Being not a Zoom admin myself, I am unsure of the Zoom admin can change the default / API language of the invitations. Do you know that?

Having said that, @andrewmadden proposed to move the zoom phrase snippets such as 'Join zoom meeting' into the plugin's lang file. This is under the assumption that the non-english language invitations are translations and not structurally different.

Alternatively, it would be just the duty of the Moodle plugin admin to adapt the Regex to his local Zoom invitation wordings if he wants to use this feature.

jrchamp: I believe that the Zoom JWT API may depend on the language of the "Owner" of the overall Zoom account. Additionally, the invitation email text can be modified under `Account -> Branding -> Emails`. Here's the default German template as an example:
image

classes/invitation.php
```diff
@@ -0,0 +1,169 @@
...
+    /** @var array $configregex Array of regex patterns defined in plugin settings. */
+    private $configregex;
+
+    /**
+     * invitation constructor.
+     *
+     * @param string|null $invitation Zoom invitation returned from
+     * https://marketplace.zoom.us/docs/api-reference/zoom-api/meetings/meetinginvitation.
+     */
+    public function __construct(?string $invitation) {
```
jrchamp, line 46: `?string` requires [Nullable type](https://www.php.net/manual/en/migration71.new-features.php#migration71.new-features.nullable-types) support which was added in PHP 7.1 and would increase the minimum PHP version required for this plugin.

classes/invitation.php
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    /**
+     * Get the display string to show on the module page.
+     *
+     * @param int $coursemoduleid Course module where the user will view the invitation.
+     * @param int|null $userid Optionally supply the intended user to view the string. Defaults to global $USER.
+     * @return string|null
+     */
+    public function get_display_string(int $coursemoduleid, int $userid = null): ?string {
```
jrchamp, line 57: `?string` requires [Nullable type](https://www.php.net/manual/en/migration71.new-features.php#migration71.new-features.nullable-types) support which was added in PHP 7.1 and would increase the minimum PHP version required for this plugin.

settings.php
andrewmadden: The `invitation_invite` setting gets generated with the rest of the regex patterns. This hideif rule will hide the pattern setting if it's not being used to remove the 'invite' element.

The name of the settings are generated as `'invitation_' . $elementname` , where the element names are the keys of the `get_default_invitation_regex()` array.

abias: Ah, thank you for the explanation, @andrewmadden !

classes/invitation.php
andrewmadden: I've removed the nullable type, to make code php7.0 compliant, thanks!

classes/invitation.php
andrewmadden: I've removed the nullable type, to make code php7.0 compliant, thanks!

classes/invitation.php
abias: Thank you, @jrchamp , for these additions.

Yes, maybe the language depends on the language of the Owner account. I can't verifiy this right now, unfortunately, with our institution account as I do not have access to the owner account.

Having said that, the question is if the implementation can remain this way (putting the duty to adapt the RegEx pattens to localizations of the invitation texts onto the Moodle admin) or not.

I would be happy if it could remain that way. And we could surely add an additional note to the invitationregexenabled_help string which makes this duty clear.

settings.php
rlorenzo: @andrewmadden I see this setting is enabled by default.

Since this is going to be difficult to test in a real-world scenario and can be prone to breakage if Zoom ever changes their invitation text can this be defaulted to off? Then power users, such as yourself, can enable it and they know they have to be aware of any text changes and fix it themselves?

classes/invitation.php
```diff
@@ -0,0 +1,210 @@
...
+     * @return string
+     *
+     * @throws \coding_exception
+     * @throws \dml_exception
+     * @throws \moodle_exception
+     */
+    private function remove_element(string $invitation, string $element): string {
+        global $PAGE;
+        $configregex = $this->get_config_invitation_regex();
+        if (!in_array($element, array_keys($configregex))) {
```
jrchamp, line 107: Option 1: If you want to make sure they have specified a non-empty string before we try to remove the element.
```suggestion
        if (empty($configregex[$element])) {
```
Option 2: If for some reason you want to allow empty string `''` and the string zero `'0'`.
```suggestion
        if (!isset($configregex[$element])) {
```

settings.php
```diff
@@ -275,4 +275,42 @@
             get_string('option_mute_upon_entry_help', 'mod_zoom'),
             1, 1, 0);
     $settings->add($defaultmuteuponentryoption);
+
+    $invitationregexhelp = get_string('invitationregex_help', 'mod_zoom');
+    if (version_compare(normalize_version($CFG->release), '3.7.0', '<')) {
```
jrchamp, line 280: Since we're using this in several places, we might as well move the check to the top of the file and save it to a boolean.

Something like:
```php
$moodlehashideif = version_compare(normalize_version($CFG->release), '3.7.0', '>=');
```

Then this could be:
```suggestion
    if (!$moodlehashideif) {
```

There is one use of this same check that predates this PR, so it would be good to update that one too.

view.php
```diff
@@ -394,7 +394,9 @@
 $table->data[] = array($strmuteuponentry, ($zoom->option_mute_upon_entry) ? $stryes : $strno);
 
 // Show dial-in information.
-if (!empty($meetinginvite)) {
+if (!empty($meetinginvite)
+        && ($zoom->option_audio == ZOOM_AUDIO_BOTH || $zoom->option_audio == ZOOM_AUDIO_TELEPHONY)
```
jrchamp, line 398:
```suggestion
        && ($zoom->option_audio === ZOOM_AUDIO_BOTH || $zoom->option_audio === ZOOM_AUDIO_TELEPHONY)
```

classes/invitation.php
jrchamp: Hmm... Option 1 is very similar to the code in add_paragraph_break_above_element(), so is there a reason that you did not use Option 1?

andrewmadden: This is already handled by `preg_replace` (line 112-114). It will return 'null' if the regex is invalid which includes an empty string.

jrchamp: I guess I'd rather not invoke preg_replace if it's guaranteed to fail. The functions `in_array` and `array_keys` are a longer route to answer a simple question and are less legible about the purpose than the language construct `empty`.

andrewmadden: I can add the empty check as well if you insist. I would point out that empty only checks for one way that preg_replace could fail so it would not remove or replace how the preg_replace is currently handled.

jrchamp: I feel like there is a miscommunication here. I'm sorry for upsetting you - I'm not trying to force my way here.

I'm not suggesting removing/replacing/changing preg_replace and I'm not suggesting adding another check. I'm specifically suggesting to use `empty($configregex[$element])` in place of `!in_array($element, array_keys($configregex))`. My suggestion matches code that you already wrote in a different function, so I asked if there was a reason to use the current more-complex check to achieve the same goal.

andrewmadden: I'm sorry if I came across as upset. I'll try to explain my intention better.

The main reason why consciously chose to only check the keys, and not use `empty()` was I wanted to give as much information back to the user via the debugging commands as getting the regex patterns right can be a bit finicky. 

Doing a simple replacement of checking array_keys with using `empty()` would actually make the error messages not make sense, and it would need to be refactored. If a Moodle user saved an empty string as the regex pattern for example, and we are using the `empty()` command, they would get a coding exception pointing them to look at the hardcoded element options. This is misleading and not the reason the exception should have been thrown.

The coding exception is only meant to be thrown if the element string provided does not match one of the element string in the default array, which means that discreetly checking the keys is the best way to confirm this.

If we also wanted to do a check to validate that the pattern is not empty (not a coding error), we could add an additional if statement to then call another debugging message. I personally think this would be unnecessary duplication of code, and my personal preference is to not duplicate verification of input when it is already handled natively by the language. If we were to add verification that the pattern is not empty, should we also add verification that the string is a valid regex pattern as well?

andrewmadden: I'll also add that `preg_replace` and `preg_match` do not have the same behaviour for handling malformed regex which is why the pattern and the output of the preg_ methods are not checked in the same way.

Additionally, this is the function that is potentially called for every element so should be responsible for providing information back to the user about any errors.

The `add_paragraph_break_above_element` is only called by one element at the moment and would possibly confuse users if it gave as much verbose information as `remove_element` (which would most likely be duplicated).

andrewmadden: I can change it to use your suggestion `isset()` as it would still allow the regex pattern check to be handled by `preg_replace`. I suppose I considered that checking the array keys made it more obvious of what was actually being checked, but it seems I may have been wrong in this regard.

jrchamp: Thank you for explaining, I understand this much better now.

It's not that we're trying to guard against a missing regex, it's that we want to notify a programmer in the future who tries to call remove_element using an element key that isn't defined by our default list of element-keyed regexes.

In fact, when the regex is missing, you're saying that we specifically *want* it to fail using the invitationmodificationfailed message instead and we're relying on the preg_replace function call to do all of that.

These details weren't immediately obvious, but I understand them now.

With all that in mind, I think that Option 2 is viable as long as the configuration values exist. If we're expecting the configuration values might not exist and this configregex might have null values, Option 3 would be:
```suggestion
        if (!array_key_exists($element, $configregex)) {
```

andrewmadden: I think I like option 3 better as it is safest. Should have used it in the first place. I've pushed the change up. Thanks for your help.

settings.php
andrewmadden: Good idea. I've pushed the change.

view.php
andrewmadden: Good point. I've pushed the change.

jrchamp:
@rlorenzo @abias As you both had comments on this, please check if any additional changes are needed.

abias:
Hi @andrewmadden, @jrchamp and @rlorenzo 

I am happy with the latest changes and the implementation as it stands now. I just tested the latest codebase with our local use cases and Zoom instance and it did was it was expected to do.

Thus, from my point of view, this is ready to be squashed and merged.
Thank you all for your work on this piece of code!

PR #236, opened 2021-03-23
Update get_meeting_reports.php

lcollong:
At least in our setup (why ?), the end_time of the meeting object is already a timestamp. Converting it using strtotime generates an empty string (at least with php v7.4.15) thus the following test is always true... never processing meeting. Proposed change should maintain compatibility with other installation while working for our server.

### DIFF
--- a/classes/task/get_meeting_reports.php
+++ b/classes/task/get_meeting_reports.php
@@ -163,7 +163,7 @@ class get_meeting_reports extends \core\task\scheduled_task {
 
         foreach ($allmeetings as $meeting) {
             // Only process meetings if they happened after the time we left off.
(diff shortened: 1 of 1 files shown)


jrchamp:
That's really interesting, because the two API calls both claim to return a date-time instead of a int or timestamp. This isn't the first time that there has been a bug in the Zoom API, but it would be good to determine how many of the 3 API endpoints in use here are handing back a different format than what is documented.

https://marketplace.zoom.us/docs/api-reference/zoom-api/dashboards/dashboardmeetings
https://marketplace.zoom.us/docs/api-reference/zoom-api/dashboards/dashboardwebinars
https://marketplace.zoom.us/docs/api-reference/zoom-api/reports/reportmeetings

Maniuz:
Last update of "get_meeting_reports.php" fix the problem for recurring meetings but not for meetings with specific date and time

Issue #237, opened 2021-03-24
Mobile app does not update "view" completion criteria


nstefanski:
It was brought to my attention that the mobile app does not update the "view" completion criterion in Moodle, it just assigns a grade if the meeting is set to be gradable. The view completion code in `loadmeeting.php` just needs to be copied into the webservice in `external.php`, I'll open a new pull request shortly.

rlorenzo:
Merged in https://github.com/ucla/moodle-mod_zoom/pull/238

PR #238, opened 2021-03-24
add view completion to moodle's webservice

nstefanski:
This resolves issue #237 by setting the module to "viewed" in the webservice used by the mobile app before passing along the join url. (See also issue #43)


jrchamp:
This makes sense to me. If you're saying that it works for you, I'm inclined to believe.

PR #247, opened 2021-04-19
Backup and restore did not include alternative hosts and some other meeting settings

abias:
This patch fixes the bug that alternative hosts and the meeting options mute_upon_entry, waiting_room and authenticated_users were not handled during backup

### DIFF
--- a/classes/external.php
+++ b/classes/external.php
@@ -104,7 +104,7 @@ class mod_zoom_external extends external_api {
         $result['audioopt'] = $zoom->option_audio;
 
         if (!$zoom->recurring) {
-            if (!$zoom->exists_on_zoom) {
+            if ($zoom->exists_on_zoom == ZOOM_MEETING_EXPIRED) {
                 $status = get_string('meeting_nonexistent_on_zoom', 'mod_zoom');
             } else if ($finished) {
                 $status = get_string('meeting_finished', 'mod_zoom');
--- a/view.php
+++ b/view.php
@@ -240,7 +240,7 @@ if ($config->showdownloadical != ZOOM_DOWNLOADICAL_DISABLE && (!($showrecreate |
 
 // Show meeting status.
 if (!$zoom->recurring) {
-    if (!$zoom->exists_on_zoom) {
+    if ($zoom->exists_on_zoom == ZOOM_MEETING_EXPIRED) {
         $status = get_string('meeting_nonexistent_on_zoom', 'mod_zoom');
     } else if ($finished) {
         $status = get_string('meeting_finished', 'mod_zoom');
--- a/backup/moodle2/backup_zoom_stepslib.php
+++ b/backup/moodle2/backup_zoom_stepslib.php
@@ -42,12 +42,12 @@ class backup_zoom_activity_structure_step extends backup_activity_structure_step
     protected function define_structure() {
         // Define the root element describing the zoom instance.
         $zoom = new backup_nested_element('zoom', array('id'), array(
-            'intro', 'introformat', 'grade',
-            'uuid', 'meeting_id', 'start_url', 'join_url',
-            'created_at', 'host_id', 'name', 'start_time', 'timemodified',
-            'recurring', 'webinar', 'duration', 'timezone', 'password', 'option_encryption_type', 'option_jbh',
-            'option_start_type', 'option_host_video', 'option_participants_video',
-            'option_audio', 'status'));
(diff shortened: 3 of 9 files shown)


jrchamp:
These changes make sense, but I noticed two pre-existing concerns: `uuid` and `status`.

`status` was replaced with `exists_on_zoom` in ba8e8c6bf3dd1e822cd9d583cc540da1e0cb1538 but it appears that some backup/restore code still references that field: https://github.com/ncstate-delta/moodle-mod_zoom/blob/f881c4953abc1169f56f74fccda9ec43497af339/backup/moodle2/restore_zoom_stepslib.php#L73

`uuid` was removed by ba8e8c6bf3dd1e822cd9d583cc540da1e0cb1538 but was not removed from the backup definition.

Since you have been testing the backup and restore code, please
- Remove `uuid` and replace `status` with `exists_on_zoom` in `backup/moodle2/backup_zoom_stepslib.php`.
- Replace `status` with `exists_on_zoom` in `backup/moodle2/restore_zoom_stepslib.php`.
- Confirm that backup and restore is still working.

abias:
@jrchamp ,

thank you for this review.

> These changes make sense, but I noticed two pre-existing concerns: uuid and status.

I have to admit that I just added the missing fields and did not mind about the correctness of the existing fields. Shame on me.

> Remove uuid and replace status with exists_on_zoom in backup/moodle2/backup_zoom_stepslib.php.

Done. I especially adjusted the order of the fields in the array to match the order of the fields in /db/install.xml.

> Replace status with exists_on_zoom in backup/moodle2/restore_zoom_stepslib.php.

Done. But I have spotted that there is more:
The constant ZOOM_MEETING_EXPIRED which is used in this file is not defined anywhere.

So I went ahead and cleaned up the code around the exists_on_zoom meeting attribute.
The main issue was that the exists_on_zoom was sometimes filled with booleans, sometimes with integers. I have now changed everything to integers and hope that existing installations always have integers in the database - at least the DB field type is XMLDB_TYPE_INTEGER. I would be grateful if you could peer-review that changes.

> Confirm that backup and restore is still working.

I have successfully tested backup and restore on my dev system and haven't spotted any glitches or debug messages, but I would be grateful if you could double-check as I am not a backup expert.

Cheers,
Alex

classes/external.php
```diff
@@ -104,7 +104,7 @@ public static function get_state($zoomid) {
         $result['audioopt'] = $zoom->option_audio;
 
         if (!$zoom->recurring) {
-            if (!$zoom->exists_on_zoom) {
+            if ($zoom->exists_on_zoom != ZOOM_MEETING_EXISTS) {
```
jrchamp, line 107:
```suggestion
            if ($zoom->exists_on_zoom == ZOOM_MEETING_EXPIRED) {
```

view.php
```diff
@@ -240,7 +240,7 @@
 
 // Show meeting status.
 if (!$zoom->recurring) {
-    if (!$zoom->exists_on_zoom) {
+    if ($zoom->exists_on_zoom != ZOOM_MEETING_EXPIRED) {
```
jrchamp, line 243:
```suggestion
    if ($zoom->exists_on_zoom == ZOOM_MEETING_EXPIRED) {
```

jrchamp (commented):
Amazing work @abias ! Thank you so much for:
- ordering the fields to match the db/install.xml
- identifying the inconsistent use of the exists_on_zoom field
- fixing the uses of exists_on_zoom to use integer values throughout

I have a couple small suggestion to make view.php and classes/external.php use the same positive version of the if condition.

view.php
abias: Oh well, I think I got the logic wrong with my change initially. Thanks for spotting this. 
I will adopt for suggested change into my branch.

classes/external.php
abias: Sure, it makes sense to align the checks.
I will adopt your suggested change into my branch.

abias:
@jrchamp , thank you for your last comments. I have updated the two highlighted places and force-pushed my branch.

jrchamp (dismissed):
This all looks good to me.

abias:
I have rebased this branch onto the latest master branch and have force-pushed the branch.

smbader (approved):
- zoom nested element updated to include all fields in the zoom table.
- status field renamed to exists_on_zoom 
- exists_on_zoom uses one of two new constants.

Good changes.

PR #249, opened 2021-04-20
prechecker: fix coding style issues

jrchamp:
Quick fixes for most of the issues listed here:
https://moodle.org/plugins/mod_zoom/v3.6/23702?smurf=html#js

Remaining:
 - JWT library, which we ignore.
 - Mustache template documentation, which I am not familiar with.


smbader (approved):
Good coding style changes.

PR #250, opened 2021-04-23
Change duration string to what it really should say

abias:
The 'duration' string is currently saying 'Duration (minutes)'.
However, the table rows on view.php and in the Moodle mobile app where this string is used push the meeting duration through format_time() which renders the duration in a nice format like '1 hour'. Thus, the 'minutes' hint is misleading and removed from the string.


jrchamp:
`format_time()` is not used here:
https://github.com/ncstate-delta/moodle-mod_zoom/blob/f881c4953abc1169f56f74fccda9ec43497af339/participants.php#L146-L151

It looks like the code is saying: if the number of seconds isn't a multiple of 60, then add on the number of seconds to make it a multiple of 60. This would be far less confusing if it was using the round-up function `ceil()` after the duration was divided by 60. Overall, I like the idea of using `format_time()` here, but it's a table that can be exported to a spreadsheet format, so that makes everything complicated if we care about it being sortable.

Do we care if it's sortable?

Personally, I feel like showing "160" is not all that helpful when someone was present for "2 hours and 40 minutes".

jrchamp:
This is just a side note: Here's another situation where duration is not passing through `format_time()`: https://github.com/ncstate-delta/moodle-mod_zoom/blob/f881c4953abc1169f56f74fccda9ec43497af339/report.php#L68

The duration in this table comes from Zoom and for the meeting I tested, it had duration set to a string "03:23:34".

jrchamp:
`normalize_meeting()` from the get_meeting_reports task tries to turn those colon separated strings into minutes... 
https://github.com/ncstate-delta/moodle-mod_zoom/blob/f881c4953abc1169f56f74fccda9ec43497af339/classes/task/get_meeting_reports.php#L568-L578

We should probably have it in the same format that is being used for the zoom.duration so that format_time() can do the right thing.

jrchamp:
Here's the point in the upgrade.php where the zoom.duration was converted from minutes to seconds: https://github.com/ncstate-delta/moodle-mod_zoom/blob/f881c4953abc1169f56f74fccda9ec43497af339/db/upgrade.php#L92

If we want to apply a similar change to the zoom_meeting_details and zoom_meeting_participants, we would need to updated the existing table data similar to above and modify the duration "normalization" code for both the meetings (`normalize_meeting()`) and the participants (`format_participant()`).

jrchamp:
Here's more code that assumes (correctly, for now) that the duration is in minutes: https://github.com/ncstate-delta/moodle-mod_zoom/blob/f881c4953abc1169f56f74fccda9ec43497af339/locallib.php#L342-L344

abias:
Oh well, @jrchamp , you are right to spot and highlight all these locations where durations are processed.

I think I was too shortsighted when I just looked at view.php, thought 'Hey, this table row isn't in minutes, it's a formatted time' and wanted to quickly change the table row label.

However, I am not sure how to deal with this PR now. Just merging it is wrong, but leaving the codebase as it is also wrong.

Do you have a proposal how to go ahead without moving too much furniture or even breaking things?

jrchamp:
I like your original plan to remove the 'minutes' part of the translated string, because the units are provided by format_time() in most of these cases. I personally don't believe that sorting is necessary for the spreadsheet export situation, but if people feel otherwise, adding a column that is designed for sorting is far less ambiguous. Moodle likes seconds and Zoom likes minutes. We're storing the information in Moodle and displaying it in Moodle. We already have a set of "normalization" functions so it's probably better to take what Zoom provides, convert it to a Moodle-friendly format, and then convert from what's stored in Moodle to the Zoom-friendly format when we send it back to the API.

This is a little bit of furniture moving, but right now we have incorrect labels on the furniture and we're not even treating them consistently. If someone tried to put together a report that fetched the raw data from the database, they would need to know that some of the columns are measured in seconds and others are measured in minutes. If we add an upgrade task to convert the two duration columns that are measured in minutes and update the code to reflect that change, I would hope that it would not be very risky.

sgrandh3:
Performed regression testing and verified the following :  1,Create and join the meeting as instructor
2. View the meeting reports .
 Duration is being displayed instead of Duration(minutes)

smbader (approved):

PR #252, opened 2021-04-26
Check if a Zoom account is active before offering it as alternative host

abias:
This patch fixes a flaw which existed before when the list of alternative hosts was composed for the alternative hosts user picker.
Adding a pending or inactive user as alternative host will result in a Zoom API error, that's why we check if the user is active before offering it as alternative host.

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -582,8 +582,10 @@ function zoom_get_selectable_alternative_hosts_list(context $context) {
         // At least, Zoom does not care if the user who is the host adds himself as alternative host as well.
 
         // Verify that the user really has a Zoom account.
+        // Furthermore, verify that the user's status is active. Adding a pending or inactive user as alternative host will result
+        // in a Zoom API error otherwise.
         $zoomuser = $service->get_user($u->email);
-        if ($zoomuser !== false) {
+        if ($zoomuser !== false && $zoomuser->status === 'active') {
             // Add user to array of users.
             $selectablealternativehosts[$u->email] = fullname($u);
         }


locallib.php
```diff
@@ -582,8 +582,10 @@ function zoom_get_selectable_alternative_hosts_list(context $context) {
         // At least, Zoom does not care if the user who is the host adds himself as alternative host as well.
 
         // Verify that the user really has a Zoom account.
+        // Furthermore, verify that the user's status is active. Adding a pending or inactive user as alternative host will result
+        // in a Zoom API error otherwise.
         $zoomuser = $service->get_user($u->email);
-        if ($zoomuser !== false) {
+        if ($zoomuser !== false && $zoomuser->status == 'active') {
```
jrchamp, line 588: As 'active' is known to be a string, a strict comparison here is precise, safe and efficient.
```suggestion
        if ($zoomuser !== false && $zoomuser->status === 'active') {
```

jrchamp (commented):
@abias A rebase here (and in the other Pull Requests with failed checks) is the only way to get the fixes from #249 that allow the checks to pass. Sorry about that, I think upstream Moodle changed the code sniffer ruleset.

locallib.php
abias: Thanks for this note (again!). I keep forgetting about strict comparisons when working with PHP.

abias:
I have rebased this branch onto the latest master branch, added the strict comparison and have force-pushed the branch.

jrchamp (approved):
We already have the object, so checking for 'active' to avoid selecting invalid options is a win-win!

smbader (approved):
The change prevents the user from choosing an option that is not helpful.

Good change!

PR #258, opened 2021-05-18
Improvement: Fully support recurring meetings

### LINKED ISSUE #189: Improvement: Fully support recurring meetings
Hi Rex,

mod_zoom currently lets teachers
* either create meetings with a start time and a duration
* or create recurring meetings which are infinitely available.

Only with the first option, students and teachers benefit from Moodle calendar entries and reliable meeting room entry times.
However, teachers tend to create a recurring meeting activity for the whole lecture term so that they don't have to create one activity per lecture day.

Looking at the Zoom management interface directly, you will see that you are able to define a recurrence (daily, weekly, monthly). This is not possible with mod_zoom but would directly 
a) produce benefits for teachers as they still have just to create one meeting activity, but could better control when the meeting room is open and when it is not.
b) produce benefits for students as mod_zoom could then create Moodle calendar entries for each recurrence.

Thus I would like to propose to add support for defining the recurrence to mod_zoom as well, providing the same UI as the Zoom management interface:

* Weekly meetings:
grafik
* Daily meetings:
grafik
* Monthly meetings:
grafik

To realize this goal, these implementastions will have to be done:

* Add a "Recurrence" dropdown to mod_form.php
** Add the values "Daily", "Weekly", "Monthly" and "No fixed time" to this dropdown. The default will be "No fixed time".
** Hook the "Recurrence" dropdown to the "Recurring meeting" checkbox with a hideIf rule so that it is only shown as soon as the "Recurring meeting" checkbox is checked.
* Add the necessary form widgets which will hold the recurrence details to mod_form.php. 
** Ideally, stick 100% with the way how Zoom handles this in their own web interface (see screenshots).
** Hook these widgets to the "Recurrence" dropdown with hideIf rules so that they are only shown as soon as the corresponding value is selected in the "Recurrence" dropdown
* When the form is submitted and the meeting is created or updated, directly pass the values of the new widgets to the Zoom API within the "recurrence" object of the endpoint https://marketplace.zoom.us/docs/api-reference/zoom-api/meetings/meetingcreate and https://marketplace.zoom.us/docs/api-reference/zoom-api/meetings/meetingupdate
* When the form is submitted and the meeting is created / updated or deleted in Zoom, create / update or delete the Moodle calendar events accordingly. There isn't a sophisticated support for recurring events in Moodle calendar. If I go to /calendar/view.php, I just can create recurring events on a weekly basis, nothing more. The approach in this case, in my point of view, would be that mod_zoom interprets the recurrence schedule which is set by the teacher and creates individual events in the Moodle calendar which are not coupled to each other in any way.

As a side note:
In real life, there will be edge cases where the recurring meeting schedule should be skipped in one week because of vacations or something like that. But mod_zoom does not need to be smarter than Zoom itself in my point of view. Teachers have to accept that there will be a meeting scheduled within the vacation in this case. If they do not accept that, they have to schedule the meetings before and after the vacation separately.

Cheers,
Alex

jwalits:
The set of commits contain the patch to improve the support for recurring zoom meetings in Moodle.

Addresses the issues that have been raised by @abias here: https://github.com/ncstate-delta/moodle-mod_zoom/issues/189

Fixes #189


locallib.php
```diff
@@ -346,24 +359,112 @@ function zoom_get_sessions_for_display($meetingid) {
...
+        // Otherwise.
+    } else {
+        // Calculate the time when the meeting becomes available at first,
+        // based on the meeting start time and the general meeting lead time.
+        $firstavailable = $zoom->start_time - ($config->firstabletojoin * 60);
+
+        // Calculate the time when the meeting ends to be available,
+        // based on the meeting start time and the meeting duration.
+        $lastavailable = $zoom->start_time + $zoom->duration;
+    }
```
jrchamp, line 454: Prefer to minimize the number of times that duplicate logic occurs. Please consider this:
```suggestion
        // Get the next occurrence time.
        $starttime = zoom_get_next_occurrence($zoom);
    } else {
        $starttime = $zoom->start_time;
    }

    // Calculate the time when the recurring meeting becomes available next,
    // based on the next occurrence start time and the general meeting lead time.
    $firstavailable = $starttime - ($config->firstabletojoin * 60);

    // Calculate the time when the meeting ends to be available,
    // based on the next occurrence start time and the meeting duration.
    $lastavailable = $starttime + $zoom->duration;
```

exportical.php
```diff
@@ -82,10 +77,60 @@
...
+    $event->add_property('uid', $zoom->meeting_id . '@' . $hostaddress); // A unique identifier.
+    $event->add_property('summary', $zoom->name); // Title.
+    $event->add_property('dtstamp', Bennu::timestamp_to_datetime()); // Time of creation.
+    $event->add_property('last-modified', Bennu::timestamp_to_datetime($zoom->timemodified));
+    $event->add_property('dtstart', Bennu::timestamp_to_datetime($zoom->start_time)); // Start time.
+    $event->add_property('dtend', Bennu::timestamp_to_datetime($zoom->start_time + $zoom->duration)); // End time.
+    $event->add_property('description', $descriptiontext);
+
+    // Add the event to the iCal file.
+    $ical->add_component($event);
```
jrchamp, line 130: A lot of these lines are the same. It would be ideal to reduce the amount of duplication while maintaining legibility, but I know that is a bit more complex with this case.

mod_form.php
```diff
@@ -606,6 +724,42 @@ public function validation($data, $files) {
...
         }
 
+        // Add validation for recurring meeting.
+        if ($data['recurring'] == 1) {
+            if ($data['recurrence_type'] == ZOOM_RECURRINGTYPE_WEEKLY) {
+                $weekdayselected = [];
+                for ($i = 1; $i <= 7; $i++) {
+                    $key = 'weekly_days_' . $i;
+                    if (!empty($data[$key])) {
+                        $weekdayselected[] = $data[$key];
```
jrchamp, line 734: Also here, if we expect that `$i` and `$data[$key]` are the same value, let's not trust user input.

db/upgrade.php
```diff
@@ -493,5 +493,73 @@ function xmldb_zoom_upgrade($oldversion) {
...
+        }
+
+        // Define and conditionally add field end_date_option.
+        $field = new xmldb_field('end_date_option', XMLDB_TYPE_INTEGER, '1', null, null, null, null, 'end_date_time');
+        if (!$dbman->field_exists($table, $field)) {
+            $dbman->add_field($table, $field);
+        }
+
+        // Zoom savepoint reached.
+        upgrade_mod_savepoint(true, 2021031600, 'zoom');
```
jrchamp, line 560: Version should probably match the change to version.php Right?
```suggestion
        upgrade_mod_savepoint(true, 2021041301, 'zoom');
```

db/upgrade.php
```diff
@@ -493,5 +493,73 @@ function xmldb_zoom_upgrade($oldversion) {
         upgrade_mod_savepoint(true, 2021030300, 'zoom');
     }
 
+    if ($oldversion < 2021031600) {
```
jrchamp, line 496:
```suggestion
    if ($oldversion < 2021041301) {
```

view.php
```diff
@@ -239,10 +248,10 @@
 }
 
 // Show meeting status.
-if (!$zoom->recurring) {
-    if (!$zoom->exists_on_zoom) {
-        $status = get_string('meeting_nonexistent_on_zoom', 'mod_zoom');
-    } else if ($finished) {
+if (!$zoom->exists_on_zoom) {
+    $status = get_string('meeting_nonexistent_on_zoom', 'mod_zoom');
+} else if (!($zoom->recurring && $zoom->recurrence_type == ZOOM_RECURRINGTYPE_NOTIME)) {
```
jrchamp, line 253: This file (and a few others) have `$zoom->recurring && $zoom->recurrence_type == ZOOM_RECURRINGTYPE_NOTIME` in them a lot. It might improve legibility to save that to a boolean that can be easily reused.

Something like:
```php
$isrecurringnotime = ($zoom->recurring && $zoom->recurrence_type == ZOOM_RECURRINGTYPE_NOTIME);
```

view.php
```diff
@@ -221,16 +221,25 @@
...
 $numcolumns = 2;
 
-// Show start/end date or recurring flag.
-if ($zoom->recurring) {
+// Show start/end date or recurring meeting information.
+if ($zoom->recurring && $zoom->recurrence_type == ZOOM_RECURRINGTYPE_NOTIME) {
     $table->data[] = array(get_string('recurringmeeting', 'mod_zoom'), get_string('recurringmeetingexplanation', 'mod_zoom'));
+} else if ($zoom->recurring && $zoom->recurrence_type != ZOOM_RECURRINGTYPE_NOTIME) {
+    $table->data[] = array(get_string('recurringmeeting', 'mod_zoom'), get_string('recurringmeetingthisis', 'mod_zoom'));
+    if (($nextoccurrence = zoom_get_next_occurrence($zoom)) > 0) {
```
jrchamp, line 229: Trying to minimize assignments within conditionals.
```suggestion
    $nextoccurrence = zoom_get_next_occurrence($zoom);
    if ($nextoccurrence > 0) {
```

jrchamp (commented):
There is a lot here. I did my best for a first pass. Please also check the prechecker results to see if you've introduced anything new that it checks.

jwalits:
Hey @jrchamp - Thank you for the feedback on the code. I will update those in the next day or so to address the concerns you have rasied


Thanks,
Jwalit

lib.php
```diff
@@ -97,6 +102,24 @@ function zoom_add_instance(stdClass $zoom, mod_zoom_mod_form $mform = null) {
...
+        // Undo everything we can. This is not necessary for databases which
+        // support transactions, but improves consistency for other databases.
+        // Code taken from course/modlib.php (lines: 133-136)
+        context_helper::delete_instance(CONTEXT_MODULE, $zoom->coursemodule);
+        $DB->delete_records('course_modules', array('id'=>$zoom->coursemodule));
+
+        // Trigger a notification, mentioning the error.
+        \core\notification::error(get_string('erroraddinstance', 'zoom'));
+        // Redirect out to course page, without adding the zoom module.
+        redirect(new moodle_url('/course/view.php', ['id' => $zoom->course]));
```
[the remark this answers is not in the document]
jwalits: I did have it set to `return false;` in my initial testing. Result was this:


Screenshot_2021-05-24 Error


I then changed the code to above, just for more graceful handling like so

Screenshot_2021-05-24 Course Test course XS


I'm happy to change it and have it set to `return false` let Moodle handle the error (the developer in me thinks thats the correct option), but from an user point of view - seeing the big error message on screen seems like something is broken? Happy to change it either way.

locallib.php
jwalits: To me it looks logical enough to change, however I'll get @abias (as its his code) to comment if its valid enough to change?

locallib.php
```diff
@@ -346,24 +359,112 @@ function zoom_get_sessions_for_display($meetingid) {
...
+
+            // Otherwise we have a recurring meeting with a recurrence schedule.
+        } else {
+            // Get the calendar event of the next occurrence.
+            $selectclause = "modulename = :modulename AND instance = :instance AND (timestart + timeduration) >= :now";
+            $selectparams = array('modulename' => 'zoom', 'instance' => $zoom->id, 'now' => time());
+            $nextoccurence = $DB->get_records_select('event', $selectclause, $selectparams, 'timestart ASC', 'timestart', 0, 1);
+
+            // If we haven't got a single event.
+            if (count($nextoccurence) != 1) {
```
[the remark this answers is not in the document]
jwalits: I'm not sure `empty($nextocurence)` is correct? @abias thoughts on this change?

amd/src/form.js
```diff
@@ -24,20 +24,136 @@
...
+        });
+    };
+
+    /**
+     * Toggle start time and duration elements.
+     */
+    var toggle_start_time_duration = function () {
+        // Disable start time and duration if "No Fixed Time" recurring meeting/webinar selected.
+        var disabled = false;
+        if ($(SELECTORS.RECURRING).prop('checked') && $(SELECTORS.REPEAT_SELECT).val() == REPEAT_OPTIONS.REPEAT_OPTION_NONE) {
```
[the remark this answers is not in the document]
jwalits: have updated the javascript to be stricter, due to which I had to use the `parseInt` function to get the javascript working correctly.

amd/src/form.js
```diff
@@ -24,20 +24,136 @@
...
+     * To show either Days, Weeks or Months
+     */
+    var toggle_repeat_interval_text = function () {
+        $(SELECTORS.REPEAT_INTERVAL).hide();
+        var repeat_select = $(SELECTORS.REPEAT_SELECT);
+        if (repeat_select.val() == REPEAT_OPTIONS.REPEAT_OPTION_DAILY) {
+            $(SELECTORS.REPEAT_INTERVAL_DAILY).show();
+        } else if (repeat_select.val() == REPEAT_OPTIONS.REPEAT_OPTION_WEEKLY) {
+            $(SELECTORS.REPEAT_INTERVAL_WEEKLY).show();
+        } else if (repeat_select.val() == REPEAT_OPTIONS.REPEAT_OPTION_MONTHLY) {
```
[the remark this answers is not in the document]
jwalits: have updated these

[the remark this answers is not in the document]
jwalits: also updated the coding style to match the updated style in the file

amd/src/form.js
```diff
@@ -24,20 +24,136 @@
...
+     * Limit the options shown in the drop-down based on repeat type selected.
+     * Max value for daily meeting is 90.
+     * Max value for weekly meeting is 12.
+     * Max value for monthly meeting is 3.
+     */
+    var limit_repeat_values = function () {
+        var selectedvalue = $(SELECTORS.REPEAT_SELECT).val();
+        // Restrict options if weekly or monthly option selected.
+        $(SELECTORS.REPEAT_INTERVAL_OPTIONS).each(function() {
+            if (selectedvalue == REPEAT_OPTIONS.REPEAT_OPTION_WEEKLY) {
```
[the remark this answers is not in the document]
jwalits: have updated those

amd/src/form.js
```diff
@@ -24,20 +24,136 @@
...
+                if (this.value > REPEAT_MAX_OPTIONS.REPEAT_OPTION_WEEKLY) {
+                    $(this).hide();
                 }
-            });
-            optionWaitingRoom.change(function() {
-                if (optionWaitingRoom.is(':checked') === true) {
-                    optionJoinBeforeHost.prop('checked', false);
+            } else if (selectedvalue == REPEAT_OPTIONS.REPEAT_OPTION_MONTHLY) {
```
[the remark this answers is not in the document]
jwalits: have updated this

classes/invitation.php
```diff
@@ -204,7 +208,8 @@ public static function get_default_invitation_regex(): array {
             'onetapmobile' => '/^one tap mobile.*(\n\s*\+.+)+$/mi',
             'dialin' => '/^dial by your location.*(\n\s*\+.+)+(\n.*)+find your local number.+$/mi',
             'sip' => '/^join by sip.*\n.+$/mi',
-            'h323' => '/^join by h\.323.*(\n.*)+?(\nmeeting id.+\npasscode.+)$/mi'
+            'h323' => '/^join by h\.323.*(\n.*)+?(\nmeeting id.+\npasscode.+)$/mi',
+            'icallink' => '/^.+download and import the following iCalendar.+$\n.+$/mi'
```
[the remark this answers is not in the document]
jwalits: have added the `,` here

classes/webservice.php
```diff
@@ -540,7 +560,32 @@ protected function _database_to_api($zoom) {
...
-        if ($data['type'] == ZOOM_SCHEDULED_MEETING || $data['type'] == ZOOM_SCHEDULED_WEBINAR) {
+        // Add recurrence object.
+        if ($zoom->recurring && $zoom->recurrence_type != ZOOM_RECURRINGTYPE_NOTIME) {
+            $data['recurrence']['type'] = (int) $zoom->recurrence_type;
+            $data['recurrence']['repeat_interval'] = (int) $zoom->repeat_interval;
+            if ($zoom->recurrence_type == ZOOM_RECURRINGTYPE_WEEKLY) {
+                $data['recurrence']['weekly_days'] = $zoom->weekly_days;
+            }
+            if ($zoom->recurrence_type == ZOOM_RECURRINGTYPE_MONTHLY) {
+                if ($zoom->monthly_repeat_option == ZOOM_MONTHLY_REPEAT_OPTION_DAY){
```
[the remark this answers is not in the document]
jwalits: fixed spacing here

classes/webservice.php
```diff
@@ -540,7 +560,32 @@ protected function _database_to_api($zoom) {
...
+                $data['recurrence']['end_times'] = (int) $zoom->end_times;
+            } else {
+                $data['recurrence']['end_date_time'] = gmdate('Y-m-d\TH:i:s\Z', $zoom->end_date_time);
+            }
+        }
+
+        if ($data['type'] == ZOOM_SCHEDULED_MEETING || 
+            $data['type'] == ZOOM_RECURRING_FIXED_MEETING ||
+            $data['type'] == ZOOM_SCHEDULED_WEBINAR ||
+            $data['type'] == ZOOM_RECURRING_FIXED_WEBINAR) {
```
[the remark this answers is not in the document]
jwalits: added the stricter matching

lib.php
```diff
@@ -97,6 +102,24 @@ function zoom_add_instance(stdClass $zoom, mod_zoom_mod_form $mform = null) {
         $zoom->host_id = $correcthostzoomuser->id;
     }
 
+    if (isset($zoom->recurring) && isset($response->occurrences) && count($response->occurrences) == 0) {
```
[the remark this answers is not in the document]
jwalits: have updated code for to use Option 1

lib.php
```diff
@@ -156,12 +184,67 @@ function zoom_update_instance(stdClass $zoom, mod_zoom_mod_form $mform = null) {
...
 
     return true;
 }
 
+/**
+ * Function to handle selected weekdays, for recurring weekly meeting.
+ * 
+ * @param stdClass $zoom The zoom instance
+ * 
+ * @return string The comma seperated string for selected weekdays
```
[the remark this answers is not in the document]
jwalits: fixed the typo

lib.php
```diff
@@ -156,12 +184,67 @@ function zoom_update_instance(stdClass $zoom, mod_zoom_mod_form $mform = null) {
...
     zoom_grade_item_update($zoom);
 
     return true;
 }
 
+/**
+ * Function to handle selected weekdays, for recurring weekly meeting.
+ * 
+ * @param stdClass $zoom The zoom instance
+ *
```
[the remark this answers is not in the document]
jwalits: removed the extra line

lib.php
```diff
@@ -156,12 +184,67 @@ function zoom_update_instance(stdClass $zoom, mod_zoom_mod_form $mform = null) {
...
+ * 
+ * @return string The comma seperated string for selected weekdays
+ */
+function zoom_handle_weekly_days($zoom) {
+
+    $weekdayselected = [];
+    for ($i = 1; $i <= 7; $i++) {
+        $key = 'weekly_days_' . $i;
+        if (!empty($zoom->$key)) {
+            $weekdayselected[] = $zoom->$key;
```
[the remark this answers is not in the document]
jwalits: have updated to `$i`

lib.php
```diff
@@ -156,12 +184,67 @@ function zoom_update_instance(stdClass $zoom, mod_zoom_mod_form $mform = null) {
...
 
+/**
+ * Function to handle selected weekdays, for recurring weekly meeting.
+ * 
+ * @param stdClass $zoom The zoom instance
+ * 
+ * @return string The comma seperated string for selected weekdays
+ */
+function zoom_handle_weekly_days($zoom) {
+
```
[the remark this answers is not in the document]
jwalits: removed the extra line

lib.php
```diff
@@ -156,12 +184,67 @@ function zoom_update_instance(stdClass $zoom, mod_zoom_mod_form $mform = null) {
...
+        }
+    }
+    return implode(',', $weekdayselected);
+}
+
+/**
+ * Function to unset the weekly options in postprocessing.
+ * 
+ * @param stdClass $data The form data object
+ *
```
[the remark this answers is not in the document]
jwalits: removed the extra line

lib.php
```diff
@@ -156,12 +184,67 @@ function zoom_update_instance(stdClass $zoom, mod_zoom_mod_form $mform = null) {
...
+        unset($data->key);
+    }
+    return $data;
+}
+
+/**
+ * Function to unset the monthly options in postprocessing.
+ * 
+ * @param stdClass $data The form data object
+ *
```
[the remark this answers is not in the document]
jwalits: removed the extra line

lib.php
```diff
@@ -359,33 +451,105 @@ function zoom_calendar_item_update(stdClass $zoom) {
     global $CFG, $DB;
     require_once($CFG->dirroot.'/calendar/lib.php');
 
+    if (!$zoom->recurring) {
+        $eventid = $DB->get_field('event', 'id', array(
+            'modulename' => 'zoom',
+            'instance' => $zoom->id
```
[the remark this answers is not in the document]
jwalits: added the `,` here

lib.php
```diff
@@ -359,33 +451,105 @@ function zoom_calendar_item_update(stdClass $zoom) {
...
+        -1 => get_string('weekoption_last', 'zoom'),
+    ];
+}
+
+/**
+ * Populate the calendar event object, based on the zoom instance
+ * 
+ * @param stdClass $zoom The zoom instance.
+ * @param stdClass $occurrence The occurrence object passed from the zoom api.
+ *
```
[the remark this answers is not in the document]
jwalits: removed the extra line

lib.php
```diff
@@ -359,33 +451,105 @@ function zoom_calendar_item_update(stdClass $zoom) {
...
+/**
+ * Populate the calendar event object, based on the zoom instance
+ * 
+ * @param stdClass $zoom The zoom instance.
+ * @param stdClass $occurrence The occurrence object passed from the zoom api.
+ * 
+ * @return stdClass The calendar event object.
+ */
+function zoom_populate_calender_item(stdClass $zoom, stdClass $occurrence = null) {
+
```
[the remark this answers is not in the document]
jwalits: removed the extra line

lib.php
```diff
@@ -359,33 +451,105 @@ function zoom_calendar_item_update(stdClass $zoom) {
...
-        $event->eventtype = 'zoom';
-        calendar_event::create($event);
+        $event->timestart = strtotime($occurrence->start_time);
+        $event->timeduration = $occurrence->duration * 60;
     }
+
+    // Recurring meetings/webinars with no fixed time are created as invisible events.
+    // For recurring meetings/webinars with a fixed time, we want to see the events on the calendar.
+    if ($zoom->recurring && $zoom->recurrence_type == ZOOM_RECURRINGTYPE_NOTIME) {
+        $event->visible = false;
```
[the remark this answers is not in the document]
jwalits: I have explicitly added a `$event->visible = true`, so as to avoid any confusion. This  will keep the existing functionality of recurring meeting/webinars without a fixed time to create events that are not visible.

lib.php
```diff
@@ -359,33 +451,105 @@ function zoom_calendar_item_update(stdClass $zoom) {
...
+        $event->timestart = $zoom->start_time;
+        $event->timeduration = $zoom->duration;
     } else {
-        $event->courseid = $zoom->course;
-        $event->modulename = 'zoom';
-        $event->instance = $zoom->id;
-        $event->eventtype = 'zoom';
-        calendar_event::create($event);
+        $event->timestart = strtotime($occurrence->start_time);
+        $event->timeduration = $occurrence->duration * 60;
```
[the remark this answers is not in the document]
jwalits: i have moved the logic to `populate_zoom_from_response` function

locallib.php
```diff
@@ -346,24 +359,112 @@ function zoom_get_sessions_for_display($meetingid) {
...
+                // Use 0 as there isn't anything better to return.
+                $cachednextoccurrence = 0;
+            }
+
+            // Use the timestamp of the event.
+            $nextoccurenceobject = reset($nextoccurence);
+            $cachednextoccurrence = $nextoccurenceobject->timestart;
```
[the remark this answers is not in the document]
jwalits: I'm not entirely sure around this code, so will let @abias comment and happy to update the code based on his comments

exportical.php
jwalits: @abias, thoughts on this one?

mod_form.php
jwalits: Have updated this to use `$i` here

db/upgrade.php
jwalits: Thanks for picking that up! I missed that when i merged Alex's code in. Have updated this now

db/upgrade.php
jwalits: Thanks, have updated this

view.php
jwalits: Good suggestion. I have added a new variable to re-use through the page

view.php
jwalits: have updated that

lib.php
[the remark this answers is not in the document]
jrchamp: The debug info and the stack trace tell me that Moodle debugging is active on your installation. You may want to try throwing an exception instead to see if the you can get a more graceful error message.

amd/src/form.js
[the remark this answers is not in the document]
jrchamp: Alternatively, if the constants were strings, parseInt would not be necessary. parseInt will turn values like `'90asdf'` into `90`.

classes/webservice.php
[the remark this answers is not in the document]
jrchamp: This does not appear to be modified.

lib.php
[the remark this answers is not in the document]
jrchamp: This does not appear to be modified.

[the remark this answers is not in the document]
jrchamp: It might be good to give it a meaningful name like `$daynumber`.

lib.php
```diff
@@ -156,12 +184,63 @@ function zoom_update_instance(stdClass $zoom, mod_zoom_mod_form $mform = null) {
...
+/**
+ * Function to unset the weekly options in postprocessing.
+ * 
+ * @param stdClass $data The form data object
+ * @return stdClass $data The form data object minus weekly options.
+ */
+function zoom_remove_weekly_options($data) {
+    // Unset the weekly_days options.
+    for ($i = 1; $i <= 7; $i++) {
+        $key = 'weekly_days_' . $i;
```
jrchamp, line 223: If you do end up changing the variable name for the function above, it should match here.

lib.php
```diff
@@ -195,7 +274,22 @@ function populate_zoom_from_response(stdClass $zoom, stdClass $response) {
     if (isset($response->start_time)) {
         $newzoom->start_time = strtotime($response->start_time);
     }
-    $newzoom->recurring = $response->type == ZOOM_RECURRING_MEETING || $response->type == ZOOM_RECURRING_WEBINAR;
+    $recurringtypes = [
+        ZOOM_RECURRING_MEETING,
+        ZOOM_RECURRING_FIXED_MEETING,
+        ZOOM_RECURRING_WEBINAR,
+        ZOOM_RECURRING_FIXED_WEBINAR
```
jrchamp, line 281:
```suggestion
        ZOOM_RECURRING_FIXED_WEBINAR,
```

lib.php
```diff
@@ -359,33 +453,104 @@ function zoom_calendar_item_update(stdClass $zoom) {
...
-    $event->timesort = $zoom->start_time;
+    $event->modulename = 'zoom';
+    $event->eventtype = 'zoom';
+    $event->courseid = $zoom->course;
+    $event->instance = $zoom->id;
+    $event->visible = true;
+    if (!$occurrence) {
+        $event->timesort = $zoom->start_time;
+    } else {
+        $event->timesort = strtotime($occurrence->start_time);
```
jrchamp, line 532: I think you already perform the strtotime() in populate_zoom_from_response()
```suggestion
        $event->timesort = $occurrence->start_time;
```

lib.php
```diff
@@ -359,33 +453,104 @@ function zoom_calendar_item_update(stdClass $zoom) {
...
+        } else {
+            calendar_event::create($event);
+        }
+    } else {
+        // First remove existing events for recurring meeting.
+        zoom_calendar_item_delete($zoom);
+        // Based on data passed back from zoom, create the calendar events.
+        if (!empty($zoom->occurrences)) {
+            foreach ($zoom->occurrences as $occurence) {
+                $event = zoom_populate_calender_item($zoom, $occurence);
```
jrchamp, line 475:
```suggestion
            foreach ($zoom->occurrences as $occurrence) {
                $event = zoom_populate_calender_item($zoom, $occurrence);
```

locallib.php
```diff
@@ -346,24 +359,112 @@ function zoom_get_sessions_for_display($meetingid) {
...
+        } else if ($zoom->recurrence_type == ZOOM_RECURRINGTYPE_NOTIME){
+            // Use 0 as there isn't anything better to return.
+            $cachednextoccurrence = 0;
+
+            // Otherwise we have a recurring meeting with a recurrence schedule.
+        } else {
+            // Get the calendar event of the next occurrence.
+            $selectclause = "modulename = :modulename AND instance = :instance AND (timestart + timeduration) >= :now";
+            $selectparams = array('modulename' => 'zoom', 'instance' => $zoom->id, 'now' => time());
+            $nextoccurence = $DB->get_records_select('event', $selectclause, $selectparams, 'timestart ASC', 'timestart', 0, 1);
```
jrchamp, line 398:
```suggestion
            $nextoccurrence = $DB->get_records_select('event', $selectclause, $selectparams, 'timestart ASC', 'timestart', 0, 1);
```

locallib.php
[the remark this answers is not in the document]
jrchamp: occurence => occurrence

locallib.php
jrchamp, line 408: Just to fix the typo, does not resolve the "else" conversation.
```suggestion
            $nextoccurrenceobject = reset($nextoccurrence);
            $cachednextoccurrence = $nextoccurrenceobject->timestart;
```

lib.php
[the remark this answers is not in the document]
jwalits: Yes, I have the full debugging enabled on my local dev env. On a production Moodle, we'd expect it will not show the full stack trace.

I have tried to `thrown new moodle_exception` (turned off debugging to check user experience) - the result is this:

chrome-capture (2)

Its slightly better user-experience than `return false`. Only slight downside I see is when I click the 'Continue' button. You get taken to the Moodle homepage, away from the course. If that flow is ok, then I can make that change in an updated commit?

view.php
```diff
@@ -221,16 +222,26 @@
...
+    $table->data[] = array($strduration, format_time($zoom->duration));
 } else {
     $table->data[] = array($strtime, userdate($zoom->start_time));
     $table->data[] = array($strduration, format_time($zoom->duration));
 }
 
 // Display add-to-calendar button if meeting was found and isn't recurring and if the admin did not disable the feature.
-if ($config->showdownloadical != ZOOM_DOWNLOADICAL_DISABLE && (!($showrecreate || $zoom->recurring))) {
+if ($config->showdownloadical != ZOOM_DOWNLOADICAL_DISABLE && !$showrecreate &&
+        !($isrecurringnotime)) {
```
jrchamp, line 244:
```suggestion
if ($config->showdownloadical != ZOOM_DOWNLOADICAL_DISABLE && !$showrecreate && !$isrecurringnotime) {
```

view.php
```diff
@@ -239,10 +250,10 @@
 }
 
 // Show meeting status.
-if (!$zoom->recurring) {
-    if (!$zoom->exists_on_zoom) {
-        $status = get_string('meeting_nonexistent_on_zoom', 'mod_zoom');
-    } else if ($finished) {
+if (!$zoom->exists_on_zoom) {
+    $status = get_string('meeting_nonexistent_on_zoom', 'mod_zoom');
+} else if (!($isrecurringnotime)) {
```
jrchamp, line 255:
```suggestion
} else if (!$isrecurringnotime) {
```

jrchamp (commented):
occurrence spelling and a couple more items I noticed.

lib.php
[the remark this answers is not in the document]
jrchamp: It looks like you can set the redirect target link on the moodle_exception:
https://github.com/moodle/moodle/blob/e29b2564380c8d57a2964f70ae816236fa91b665/lib/setuplib.php#L82-L85

mod_form.php
jrchamp: Does not appear to be updated. If the variable name was changed in the other two places, would be ideal to change it here too.

amd/src/form.js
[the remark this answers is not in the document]
jwalits: I considered that as well, but I couldn't have them as strings as I need to use `>` operator at  couple of places. With strings it doesn't work, so thought `parseInt` might be the best option for stricter comparisions

view.php
jrchamp: I must be missing a commit, because I don't see this one either.

lib.php
```diff
@@ -198,11 +198,9 @@ function zoom_update_instance(stdClass $zoom, mod_zoom_mod_form $mform = null) {
  * Function to handle selected weekdays, for recurring weekly meeting.
  * 
  * @param stdClass $zoom The zoom instance
- * 
- * @return string The comma seperated string for selected weekdays
+ * @return string The comma seprated string for selected weekdays
```
jrchamp, line 201:
```suggestion
 * @return string The comma separated string for selected weekdays
```

amd/src/form.js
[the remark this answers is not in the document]
jrchamp: Good point. Then let's set the second parameter of parseInt to 10 so that it's not trying to parse hexadecimal or octal values.

`parseInt($(SELECTORS.REPEAT_SELECT).val(), 10);`

jrchamp:
@jwalits Can you see this report: https://github.com/ncstate-delta/moodle-mod_zoom/runs/2612496443?check_suite_focus=true

There were 23 errors that the automated tests detected which would prevent merging. I can copy and paste them in the comments if that would be helpful.

jwalits:
Hey @jrchamp 

Thanks, I can see the codechecker errors in the Actions tab. I have updated a fair few of those alrready, are you able to approve the latest workflow please? https://github.com/ncstate-delta/moodle-mod_zoom/actions/runs/869817570 

Some of the line numbers from the previous workflow are not lining up with the updated code and I just want to make sure I dont miss any

classes/webservice.php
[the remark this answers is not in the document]
jwalits: Sorry, I made the change locally, but missed the file in the new commit. I will push those changes with the new commit shortly.

lib.php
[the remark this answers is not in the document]
jwalits: Thanks for that suggestion. I hadn't used a redirect link variable with moodle exception previously! That'll come in handy :-)

jwalits:
Hey @jrchamp 

I have updated most of the concerns the codechecker had. I had to revert and re-add the patch from @abias. Seemed like whilst resolving conflicts locally, I must've missed a couple of changes which were causing the unit tests to fail.

I have tested it locally and those unit tests are passing locally, hopefully Github Actions will also be happy :crossed_fingers:

amd/src/form.js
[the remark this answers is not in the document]
jwalits: have updated to set the radix parameter to 10

classes/webservice.php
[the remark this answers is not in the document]
jwalits: have commited this change as well

lib.php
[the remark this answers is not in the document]
jwalits: Have updated the logic here to throw exception with the "Continue" button going to the course page

lib.php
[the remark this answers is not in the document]
jwalits: Have updated the variable name at the multiple places where its used

locallib.php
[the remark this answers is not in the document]
jwalits: updated the typo

lib.php
jwalits: added the extra `,` here

lib.php
jwalits: Thanks, I have removed the unneccesary `strtotime`

lib.php
jwalits: updated spelling here

locallib.php
jwalits: updated spelling here

locallib.php
jwalits: updated the spelling here

view.php
jwalits: removed the additional brackets

view.php
jwalits: removed the additional brackets

lib.php
jwalits: fixed spelling

lib.php
```diff
@@ -359,33 +445,104 @@ function zoom_calendar_item_update(stdClass $zoom) {
...
+    $event->modulename = 'zoom';
+    $event->eventtype = 'zoom';
+    $event->courseid = $zoom->course;
+    $event->instance = $zoom->id;
+    $event->visible = true;
+    if (!$occurrence) {
+        $event->timesort = $zoom->start_time;
+    } else {
+        $event->timesort = $occurrence->start_time;
+    }
```
jrchamp, line 525: Would be good to move this if block down and merge it with the other `if (!occurrence) {` block.

locallib.php
```diff
@@ -346,24 +359,112 @@ function zoom_get_sessions_for_display($meetingid) {
...
+    // Prepare an ad-hoc request cache as this function could be called multiple times throughout a request
+    // and we want to avoid to make duplicate DB calls.
+    $cacheoptions = array(
+            'simplekeys' => true,
+            'simpledata' => true
+    );
+    $cache = cache::make_from_params(cache_store::MODE_REQUEST, 'zoom', 'nextoccurrence', array(), $cacheoptions);
+
+    // If the next occurrence wasn't already cached, fill the cache.
+    if (($cachednextoccurrence = $cache->get($zoom->id)) === false) {
```
jrchamp, line 382:
```suggestion
    $cachednextoccurrence = $cache->get($zoom->id);
    if ($cachednextoccurrence === false) {
```

locallib.php
```diff
@@ -346,24 +359,112 @@ function zoom_get_sessions_for_display($meetingid) {
...
+
+            // If we haven't got a single event.
+            if (count($nextoccurrence) != 1) {
+                // Use 0 as there isn't anything better to return.
+                $cachednextoccurrence = 0;
+            }
+
+            // Use the timestamp of the event.
+            $nextoccurenceobject = reset($nextoccurrence);
+            $cachednextoccurrence = $nextoccurenceobject->timestart;
```
jrchamp, line 408: Following up on the discussion, here's the else:
```suggestion
            } else {
                // Use the timestamp of the event.
                $nextoccurenceobject = reset($nextoccurrence);
                $cachednextoccurrence = $nextoccurenceobject->timestart;
            }
```

locallib.php
```diff
@@ -346,24 +359,112 @@ function zoom_get_sessions_for_display($meetingid) {
...
+
+            // Otherwise we have a recurring meeting with a recurrence schedule.
+        } else {
+            // Get the calendar event of the next occurrence.
+            $selectclause = "modulename = :modulename AND instance = :instance AND (timestart + timeduration) >= :now";
+            $selectparams = array('modulename' => 'zoom', 'instance' => $zoom->id, 'now' => time());
+            $nextoccurrence = $DB->get_records_select('event', $selectclause, $selectparams, 'timestart ASC', 'timestart', 0, 1);
+
+            // If we haven't got a single event.
+            if (count($nextoccurrence) != 1) {
```
jrchamp, line 401: I do not like using count() if we don't care about the actual count. I have two different ideas.

Option 1: If `$nextoccurrence` is an array, then we can use `empty()`
```suggestion
            if (empty($nextoccurrence)) {
```

Option 2: If `empty()` returns false because `$nextoccurrence` is an object that confuses `empty()`
```suggestion
            if (count($nextoccurrence) === 0) {
```

After all, if there's one event or two events, we're still only going to pick the first one. And in theory, the limit arguments passed to get_records_select will help ensure sure there's only the first one.

locallib.php
```diff
@@ -346,24 +359,112 @@ function zoom_get_sessions_for_display($meetingid) {
...
 
-    $available = $zoom->recurring || $inprogress;
+    // Determine if the meeting is available,
+    // based on the fact if it is recurring or in progress.
+    $available = ($zoom->recurring && $zoom->recurrence_type == ZOOM_RECURRINGTYPE_NOTIME) || $inprogress;
 
-    $finished = !$zoom->recurring && $now > $lastavailable;
+    // Determine if the meeting is finished,
+    // based on the fact if it is recurring or the meeting end time is still in the future.
+    $finished = !($zoom->recurring && $zoom->recurrence_type == ZOOM_RECURRINGTYPE_NOTIME) && $now > $lastavailable;
```
jrchamp, line 465: Might be less confusing to have an "$isrecurringnotime" variable to slightly simplify the logic here.

mod_form.php
```diff
@@ -478,6 +563,39 @@ public function data_postprocessing($data) {
...
+            // If weekly recurring selected, unset monthly options.
+            if ($data->recurrence_type == ZOOM_RECURRINGTYPE_WEEKLY) {
+                // Unset the monthly fields.
+                $data = zoom_remove_monthly_options($data);
+            }
+            // If daily recurring selected, unset weekly and monthly options.
+            if ($data->recurrence_type == ZOOM_RECURRINGTYPE_MONTHLY) {
+                // Unset the weekly fields.
+                $data = zoom_remove_weekly_options($data);
+            }
```
jrchamp, line 597: I read this through a couple times and I think it has the same effect but with less duplication.
```suggestion
            // If "No fixed time" meeting selected, don't need repeat_interval and other options.
            if ($data->recurrence_type == ZOOM_RECURRINGTYPE_NOTIME) {
                unset($data->repeat_interval);
                // Unset end_times and end_date.
                unset($data->end_date_option);
                unset($data->end_times);
                unset($data->end_date_time);
            }
            // If weekly recurring is not selected, unset weekly options.
            if ($data->recurrence_type != ZOOM_RECURRINGTYPE_WEEKLY) {
                $data = zoom_remove_weekly_options($data);
            }
            // If monthly recurring is not selected, unset monthly options.
            if ($data->recurrence_type != ZOOM_RECURRINGTYPE_MONTHLY) {
                $data = zoom_remove_monthly_options($data);
            }
```

lib.php
```diff
@@ -156,12 +176,63 @@ function zoom_update_instance(stdClass $zoom, mod_zoom_mod_form $mform = null) {
...
+ * Function to unset the weekly options in postprocessing.
+ *
+ * @param stdClass $data The form data object
+ * @return stdClass $data The form data object minus weekly options.
+ */
+function zoom_remove_weekly_options($data) {
+    // Unset the weekly_days options.
+    for ($i = 1; $i <= 7; $i++) {
+        $key = 'weekly_days_' . $i;
+        unset($data->key);
```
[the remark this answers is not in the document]
jwalits: have updated that

lib.php
jwalits: have moved that with the other block

locallib.php
jwalits: added the else statement here

locallib.php
jwalits: updated to use `empty()` here

mod_form.php
jwalits: thanks for the suggestion, have refactored this chunk of code

locallib.php
jwalits: have added a new variable to handle this logic

locallib.php
```diff
@@ -346,24 +359,116 @@ function zoom_get_sessions_for_display($meetingid) {
...
+ *             the timestamp of the meeting start date if this isn't a recurring meeting.
+ */
+function zoom_get_next_occurrence($zoom) {
+    global $DB;
+
+    // Prepare an ad-hoc request cache as this function could be called multiple times throughout a request
+    // and we want to avoid to make duplicate DB calls.
+    $cacheoptions = array(
+            'simplekeys' => true,
+            'simpledata' => true
```
jrchamp, line 377:
```suggestion
        'simplekeys' => true,
        'simpledata' => true,
```

mod_form.php
```diff
@@ -606,6 +713,42 @@ public function validation($data, $files) {
...
+
+            if ($data['recurrence_type'] == ZOOM_RECURRINGTYPE_MONTHLY) {
+                // For monthly, max is 3 months.
+                if ($data['repeat_interval'] > 3) {
+                    $errors['repeat_group'] = get_string('err_repeat_monthly_interval', 'zoom');
+                }
+            }
+
+            if ($data['recurrence_type'] != ZOOM_RECURRINGTYPE_NOTIME && $data['end_date_option'] == ZOOM_END_DATE_OPTION_BY) {
+                if ($data['end_date_time'] < strtotime('today')) {
```
jrchamp, line 743:
```suggestion
                if ($data['end_date_time'] < time()) {
```

locallib.php
jwalits: have updated the spacing and added `,`

mod_form.php
jwalits: updated that to use `time()`

jwalits:
Hey @jrchamp 

I have rebased y branch to incorportate the new changes that were merged into the master branch. Would you please be able to give the latest changes another review and enable the latest workflow?

Thanks,
Jwalit

jrchamp (commented):
I'll set the workflow to run.

This seems to be close to complete and it's getting difficult for me to see this with a fresh perspective, so I'll probably need to get someone else to take a look at this.

jwalits:
> I'll set the workflow to run.
> 
> This seems to be close to complete and it's getting difficult for me to see this with a fresh perspective, so I'll probably need to get someone else to take a look at this.

No poblems at all. I found some issues with the code after i had rebased the branch. Could you please run the workflow again to see if all the errors got fixed with my latest change

Thanks,

jwalits:
In regards to this one: https://github.com/ncstate-delta/moodle-mod_zoom/pull/258#discussion_r645118665

I have not heard back yet, but I will take a closer look at the code on Monday next to see if I can clean it up a bit more

exportical.php
abias: I could imagine that you move the logic block to create the iCal event into a helper function and call it in these locations here. But passing the necessary amount of parameters to the function and switching between a recurring and a non-recurring meeting will create some extra lines as well.

So, personally, as both blocks are located directly besides each other and the risk that a future developer modifies one block and forgets to modify the other as well is comparably low, my gut feeling is that we should leave this as it is here.

locallib.php
abias: Just looking at what the code does, this change is fine for me.

But please note the difference in the comments between line 442-447 and line 445-450: In the first block, we are dealing with a recurring meeting and the next occurrence start time. In the second block, we are dealing with a non-recurring meeting and its start time.

If you do this change, you should adapt the comments to reflect this difference as well.

locallib.php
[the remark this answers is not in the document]
abias: I see that you have already made this change after you pinged me and I think that this is ok.

In https://github.com/ncstate-delta/moodle-mod_zoom/pull/258/files#diff-3494b45596f0f648d894a5f22070eaae01f3e0ea664be0430b8d23eb7d93e35aR403, we are limiting the number of results to 1, so `$nextocurence` will either contain 1 or 0 results. There is no need to `count()` the number of results

abias:
Hey @jwalits and @jrchamp ,

I am sorry for not joining the party here earlier. I was on leave when most of the work was done by you and didn't catch up until now.

First of all, thank you both the the amazing and extensive review. This looks as if it's really close to being integrated.

You pinged me in several comments and I have commented now in these comments:
* https://github.com/ncstate-delta/moodle-mod_zoom/pull/258#discussion_r637630620
* https://github.com/ncstate-delta/moodle-mod_zoom/pull/258#discussion_r637630285
* https://github.com/ncstate-delta/moodle-mod_zoom/pull/258#discussion_r634530118

I am looking forward to seeing this pushed over the line!

Cheers,
Alex

locallib.php
jrchamp: Thanks @abias ! I did not looks closely enough at the comments. I've updated the suggestion with language that I believe addresses both situations.

exportical.php
jrchamp: My gut feeling was the same, but then I noticed that the `uid` property was being constructed in a way that could result in collisions and contained domain information that violates the current version of the iCalendar RFC. Here's the quick draft of what I'm proposing, but I'm still open to other ideas.

```php
    /**
     * Creates an iCalendar_event for a Zoom meeting.
     *
     * @param stdClass $zoom The meeting object.
     * @param string $description The event description.
     * @param ?stdClass $occurrence Optionally, the next occurrence of a recurring event.
     * @return iCalendar_event
     */
    function zoom_helper_icalendar_event($zoom, $description, $occurrence = null) {
        if (isset($occurrence)) {
            $starttime = $occurrence->timestart;
        } else {
            $starttime = $zoom->start_time;
        }

        $event = new iCalendar_event;
        $event->add_property('uid', 'zoom://' . $zoom->meeting_id . '/' . $starttime); // A unique identifier.
        $event->add_property('summary', $zoom->name); // Title.
        $event->add_property('dtstamp', Bennu::timestamp_to_datetime()); // Time of creation.
        $event->add_property('last-modified', Bennu::timestamp_to_datetime($zoom->timemodified));
        $event->add_property('dtstart', Bennu::timestamp_to_datetime($starttime)); // Start time.
        $event->add_property('dtend', Bennu::timestamp_to_datetime($starttime + $zoom->duration)); // End time.
        $event->add_property('description', $description);
        return $event;
    }
```

```suggestion
        $event = zoom_helper_icalendar_event($zoom, $descriptiontext, $o);

        // Add the event to the iCal file.
        $ical->add_component($event);
    }
} else {
    // Otherwise, this event isn't for a recurring meeting.
    $event = zoom_helper_icalendar_event($zoom, $descriptiontext);

    // Add the event to the iCal file.
    $ical->add_component($event);
```

classes/task/update_meetings.php
```diff
@@ -108,15 +107,10 @@ public function execute() {
...
-                    foreach ($calendarfields as $field) {
-                        if ($zoom->$field != $newzoom->$field) {
-                            zoom_calendar_item_update($newzoom);
-                            break;
-                        }
-                    }
                 }
+
+                // Update the calendar events.
+                zoom_calendar_item_update($newzoom);
```
jrchamp, line 113: This change causes the calendar items to update every time, even if the Zoom record has not changed in the database. I think it still need to be inside the `if ($changed) {` conditional.

I suggest the following to address the unnecessary updates and resolve the merge conflict:
```suggestion

                    // Update the calendar events.
                    zoom_calendar_item_update($newzoom);

                    // Show trace message.
                    mtrace('  => Updated calendar item for Zoom meeting ID ' . $zoom->meeting_id);
                } else {
                    // Show trace message.
                    mtrace('  => Skipped Zoom meeting activity for Zoom meeting ID ' . $zoom->meeting_id . ' as unchanged');
                }
```

locallib.php
abias: That's fine, @jrchamp !

I would just like to propose to be slightly more verbose in the if-else sections:

```suggestion
        // Get the next occurrence start time.
        $starttime = zoom_get_next_occurrence($zoom);
    } else {
        // Get the meeting start time.
        $starttime = $zoom-&gt;start_time;
    }
          
    // Calculate the time when the meeting will be available to join,
    // subtracting the meeting lead time from the start time.
    $firstavailable = $starttime - ($config-&gt;firstabletojoin * 60);
          
    // Calculate the time when the meeting will stop being available,
    // adding the start time and the meeting duration.
    $lastavailable = $starttime + $zoom->duration;
```

If you agree, @jwalits could merge this change into his working branch from my point of view.

exportical.php
abias: @jrchamp , looking at your quick draft, I have to admit that the solution is shorter than I imagined when writing my response yesterday. So I would vote +1 to changing the duplicated code into this helper function.

-----

Regarding the `uid` property, that's a separate story as it also affects existing iCal events for non-recurring events.
I would like to see this discussed in a different issue, but as we are introducing additional iCal events for recurring events, we can't avoid to discuss it on-the-fly now.

Currently, the `uid` property is filled with `<the Zoom meeting ID number>@<the FQDN of the Moodle instance>`.
Looking at the original RFC on https://datatracker.ietf.org/doc/html/rfc5545#section-3.8.4.7, this structure should be fine.

This patch here will pick up this structure and will extend it to
`<the Zoom meeting ID number>:<the start time of the particular occurrence>@<the FQDN of the Moodle instance>` for recurring event occurrences.
As to my understanding of RFC 5545, this should still be fine.

However, there is an update to the RFC on https://datatracker.ietf.org/doc/html/rfc7986#section-5.3 which discourages the usage of any identifying items like hostnames in the `uid` property. 
Additionally, we seems to have forgotten to check what happens if the start time of a particular occurrence changes and with it the `uid` property of the iCal event changes as well. Will iCal clients still detect this as an update of the original event or will they simply handle this as a new event? I am usure currently.

Looking at your proposal of using `zoom://' . $zoom->meeting_id . '/' . $starttime`, I am wondering 
a) do we really need to build this `uid` property with a URI-scheme including a protocol and slashes? This is not more  RFC-compliant than the original proposal in my point of view.
b) again how clients will behave if we change the `uid` property of existing non-recurring iCal events when the user downloads the iCal file of a meeting once more.

Coming to an end, I would like to propose to stick to the current `uid` property structure for non-recurring events, even if it is not compliant to RFC 7986 anymore.

And, for recurring events within this patch, I would like to propose to switch to
`<the Zoom meeting ID number>:<the Moodle database row id of the particular occurrence in the mdl_event table>@<the FQDN of the Moodle instance>`
where the database row id can be obtained on https://github.com/ncstate-delta/moodle-mod_zoom/pull/258/files#diff-4d74196f65c25fc69b019dde5d459f3b46202eb181c0b55f4251d5f19d3a22c3R89 to use a persistent identifier for the particular occurrence without the risk that clients behave oddly if the starttimes change.

What do you think?

classes/task/update_meetings.php
jwalits: Hey @jrchamp 

The reason I removed that check in this case was due to the one edge case that I found in my testing. If an individual occurrence is modified on Zoom, it would not trigger that code and the events in Moodle would never update. As the occurrence object returned from the API is changed, not the fields we test for in the `$calendarfields` array are not changed.

The way we're currently handling handling events for recurring meetings is by deleting existing events and re-adding when an activity is added/updated in Moodle based on the API response from zoom.

For non-recurring meetings, I think we can keep that change. For recurring meetings, though I feel we should update the calendar events regardless.

Please let me know your thoughts on this?

In the meantime, I'll rebase my branch with the latest changes

jwalits: I've added a slight update (along with the rebasing of my working branch) to the calendar code after i wrote my comment: https://github.com/ncstate-delta/moodle-mod_zoom/pull/258/commits/7b69c3b28537450e3bd4fc8648a3c7842322a22c

Please let me know if that makes sense to you. I have left the calendar update as it is for a non-recurring meeting, but for a recurring meeting it will update the calendar events regardless

exportical.php
jwalits: Thoughts on this one @jrchamp?

locallib.php
jwalits: Does that look suitable to you @jrchamp? Happy to update the code in my working branch if both of you are happy?

jrchamp:
Sorry, I'm a bit behind on this and other tasks. Thank you for your patience.

exportical.php
jrchamp: @abias You make a lot of good points about the `uid`. The URI-scheme was more to try and not have it collide with other things, but you have a good point that we want it to collide with itself. I've very tempted to use the same format that Moodle uses: https://github.com/moodle/moodle/blob/511a87f5fc357f18a4c53911f6e6c7f7b526246e/calendar/export_execute.php#L208

 - Moodle core: `$event->id . '@' . $hostaddress;`
 - Current mod_zoom: `$zoom->meeting_id . '@' . $hostaddress;`
 - PR 258 non-recurring: `$zoom->meeting_id . '@' . $hostaddress;`
 - PR 258 recurring: `$zoom->meeting_id . ':' . $o->timestart . '@' . $hostaddress;`
 - @abias recurring: `$zoom->meeting_id . ':' . $event->id . '@' . $hostaddress;`

Since we have the `$event->id`, can we drop the `$zoom->meeting_id . ': ' . `? That would bring us full circle back to the original event identifiers that Moodle had in the first place. I think this would be an ideal solution, but there is one caveat. For recurring meetings, the event rows are deleted and replaced every time the update_meetings task runs: https://github.com/ncstate-delta/moodle-mod_zoom/blob/7b69c3b28537450e3bd4fc8648a3c7842322a22c/lib.php#L462-L470

locallib.php
jrchamp: Yes, that sounds good. On my screen, it looks like some `&gt;` snuck in where `>` should be in Alex's suggestion, but otherwise, I'm very happy.

classes/task/update_meetings.php
jrchamp: I understand and we can probably make this option work. However, if we're allowing these occurrences to be created through Moodle, it might be fair for us to say to the users "We will not reflect changes that you have made directly in Zoom to individual occurrences." The benefit to us and our server administrators is significant, because it greatly reduces the mostly unnecessary churn that would otherwise occur. It also has some relationship to the iCal uid discussion happening in the other thread.

locallib.php
abias: > On my screen, it looks like some &gt; snuck in where > should be in Alex's suggestion, but otherwise

Oh, you are right. That seems to be some copy & paste issue. I updated my comment above.

@jwalits, could you please proceed with this comment here then?

exportical.php
abias: Thank you @jrchamp for your additions!

You made me think about these points:

* I was not aware / I have never checked how Moodle handles the `uid` attribute itself in its own iCal export. Thank you for checking that. And yes, it would be great to align with this scheme instead of inventing our own one.
* However, I do not think that we can adopt this scheme exactly. The Zoom iCal export would then export iCal events with the same uid as the Moodle calendar export has. What if a user has subscribed to his Moodle calendar with iCal and then downloads a Zoom meeting iCal file? I assume that both events will collide.
* (The same collision would already happen with the current code if a `$event->id` happens to match a `$zoom->meeting_id`. Sure, the risk is really low, but it is now zero).
* You make a very important point that the current code which updates the Moodle events of a recurring meeting delete all events and then creates them from scratch. This was made to circumvent any event synchronization and @jwalits is the better one to explain the details, but this is a show-stopper here now because we can't use the `$event->id` if we have to assume that it changes constantly. @jwalits, do you think we have to redesign this part of the code?

jwalits: Hey @jrchamp @abias 

Let me address the issues about `$event->id` part. When Zoom API returns the `occurrences` array, we map that into the `mdl_event` table in Moodle. This doesn't allow us to store the `occurrence_id` anywhere to make use of next time there is an update. Basically there is no correlation between the entries stored in Moodle event table and occurences array that is passed back from the API.

This makes it difficult to update the events table individually, thus I had to put the delete and re-create code. I'm not too sure how to re-design this part of the code if we can't store the `occurrence_id` somewhere in Moodle

Back to unique identifier, my thought is that using `$event->timestart` as we have currently should be unique enough?

jwalits: Just as I write this, I'm having a look at the `mdl_event` table - there is a `uuid` field which looks like it is related to iCal function. I'm thinking maybe we can store the `occurrence_id` (passed from the API) into that field. Then for the unique identifier for recurring meetings, we can use that uuid field
i.e `$zoom->meeting_id . ':' . $event->uuid . '@' . $hostaddress;`

With this I can possibly update the update events code to make use of the uuid and only update the events when they are actually changed?

Let me know your thoughts around this idea and I can make those changes?

jrchamp: If the events can be updated, then I'm even more interested in using the core event logic as much as possible.

Alex said "The Zoom iCal export would then export iCal events with the same uid as the Moodle calendar export has. What if a user has subscribed to his Moodle calendar with iCal and then downloads a Zoom meeting iCal file? I assume that both events will collide." And I agree - the events would collide. My suggestion is that they *should* collide. The question is: what values are we setting that are different from the core Moodle values? As far as I can tell, there are no valuable differences. In fact, the exportical.php changes in this Pull Request make the code *more* closely aligned with the core functionality. And that is why I'm now able to see that the exportical.php file is probably unnecessary. For now, we can keep it, but we should make it depend completely on the mdl_events so that both: align with the Moodle core events code and we do not need custom logic for the recurring pieces. That way, if we keep the mdl_events table updated (as we're already trying to do), we'll get the rest of the benefits almost for free.

classes/task/update_meetings.php
abias: Hi @jrchamp ,

let me add some detail:
In the first implementation of this feature, @jwalits went the hard way of calculation the occurrences of the recurring meeting to create the calendar event items on the Moodle site. This somehow made sense as, apparently, all information about the recurrence was there in the configuration of the Zoom activity instance and we could save a roundtrip to the Zoom API. However, we had to see that there are several edge cases which Zoom handles differently then we would have expected with the result that Moodle events were created for occurrences which did not exist in Zoom at all and the other way round.
After trying to fix these edge cases for some time, @jwalits changed the approach to using the occurrences attribute from the Zoom API response which cleary states which occurrences Zoom has scheduled based on the given recurrence configuration. With the occurrences attribute, the calendar events are created. However, as Moodle does not need this attribute for its own subsequent operation, it is not stored in Moodle.

Your point that the calendar items are updated again and again when the Zoom meeting information is fetched is valid.  And I think that @jwalits went into the right direction with https://github.com/ncstate-delta/moodle-mod_zoom/commit/7b69c3b28537450e3bd4fc8648a3c7842322a22c for non-recurring events.

Thinking about recurring events, I can imagine two solution approaches now:

1. We extend https://github.com/ncstate-delta/moodle-mod_zoom/commit/7b69c3b28537450e3bd4fc8648a3c7842322a22c in a way that it also applies for recurring events and also checks if any of the recurrence configuration attributes ('recurrence_type', 'repeat_interval', 'weekly_days', 'monthly_day',                'monthly_week', 'monthly_week_day', 'end_times', 'end_date_time') have been changes in Zoom. If any of these have changed, we should update the calendar events for the recurring meeting based on the occurrences attribute  from the Zoom response.
Implementing this should be quite easy and I can't think of side effects right now.

2. We store the occurrences attribute from the Zoom response in Moodle as well. This way, we could compare the stored attribute value with the attribute value from the Zoom response. However, this seems like a data storage overhead to me. And, consequently, we would have to modify existing code like zoom_get_next_occurrence() (https://github.com/ncstate-delta/moodle-mod_zoom/pull/258/files#diff-3494b45596f0f648d894a5f22070eaae01f3e0ea664be0430b8d23eb7d93e35aR374) which currently works on the created calendar events as well.

Against the background that I would like to see this merged now soon, I would be grateful if you could tell us if option 1 would be conducive in your point of view or not so that @jwalits could make these last changes.

Cheers,
Alex

locallib.php
abias: @jwalits, regardless of the other open discussions here, could you please push this small change to the PR so that this discussion could be resolved?

exportical.php
abias: I am reading and commenting this thread just after my comment in https://github.com/ncstate-delta/moodle-mod_zoom/pull/258#discussion_r665914959.

There, we are discussing the question if we can avoid that _all events_ of a recurring meeting get dropped and recreated unnecessarily if nothing has changed on the Zoom side.

Here, we were discussing something different (the uuid parameter of the calendar items in the iCal export) but were seeing now on-the-fly that it might be possible to check if a particular occurrence has changed and only update the corresponding calendar event.

I would say: @jwalits, if you can use the uuid parameter in the mdl_event table to solve the discussion in https://github.com/ncstate-delta/moodle-mod_zoom/pull/258#discussion_r665914959 and can change this in reasonable time, then please go for it.

----

Regarding the question what the uuid attribute of the iCal events should contain, if we strive to let the uuids collide with the Moodle core calendar export or not and if we do breaking changes to non-recurring events' uuids or not, I would like to let you, @jrchamp , as the plugin maintainer have the last say now.

At Ulm university, we have disabled the iCal export in mod_zoom and have just implemented this aspect to make this contribution here in this PR complete.

Cheers,
Alex

jrchamp: While it is a slightly breaking change, I feel very good about it as more of an "un-breaking" change. Particularly with the addition of recurring meetings and the availability of the Moodle core calendar export, we need those events to collide so we don't have tens or hundreds of duplicates.

As far as the code, here's roughly what I think we'll probably see:

```php
$newevents = array();
foreach ($zoom->occurrences as $occurrence) {
    $uuid = $occurrence['occurrence_id'];
    $newevents[$uuid] = zoom_populate_calender_item($zoom, $occurrence);
}

$conditions = array('modulename' => 'zoom', 'instance' => $zoom->id);
$events = $DB->get_records('event', $conditions);
foreach ($events as $event) {
    $uuid = $event->uuid;
    if (isset($newevents[$uuid])) {
        $changed = false;
        $newevent = $newevents[$uuid];
        foreach ($newevent as $key => $value) {
            if ($value != $event->$key) {
                $changed = true;
            }
        }

        if ($changed) {
            \calendar_event::load($event)->update($newevent);
        }

        unset($newevents[$uuid]);
    } else {
        \calendar_event::load($event)->delete();
    }
}

foreach ($newevents as $uuid => $newevent) {
    \calendar_event::create($newevent);
}
```

zoom_populate_calender_item() will also need to add:
```php
$event->uuid = $occurrence->occurrence_id;
```

locallib.php
jwalits: I have now updated this in the latest commit

exportical.php
jwalits: Thanks @jrchamp and @abias,
I have added support for uuid in my latest commit. Please have a look when you get a chance to see if it all makes sense.

In my testing locally it is working as expected and covering the edge cases that I had found previously.

Once both you you are happy, I can rebase my branch to resolve the conflicts

locallib.php
```diff
@@ -350,24 +363,109 @@ function zoom_get_sessions_for_display($meetingid) {
...
+        if (!$zoom->recurring) {
+            // Use the meeting start time.
+            $cachednextoccurrence = $zoom->start_time;
+
+            // Or if this is a recurring meeting without fixed time.
+        } else if ($zoom->recurrence_type == ZOOM_RECURRINGTYPE_NOTIME) {
+            // Use 0 as there isn't anything better to return.
+            $cachednextoccurrence = 0;
+
+            // Otherwise we have a recurring meeting with a recurrence schedule.
```
jrchamp, line 398: We don't need to change anything here, but I want to note that these two `if` blocks are primarily to improve runtime efficiency at the cost of slight code complexity. I believe that the result would be mostly the same either way, with the one exception being that non-recurring meetings which have already occurred would have a `0` next occurrence time instead of a time in the past.

jrchamp (commented):
This is looking really good. I added a couple suggestions that should simplify the code a bit and align with the Moodle uids for the iCal events.

jwalits:
Hey @jrchamp 

Thanks for the suggestions for simplifying the code around the iCal events. I have made those changes in the latest commit. I have also rebased my branch with the latest changes.

Could you please enable Actions and give the changes another review to see if they are ready to be merged in?

locallib.php
```diff
@@ -799,3 +897,30 @@ function zoom_get_alternative_host_array_from_string($alternativehoststring) {
...
+ *
+ * @return iCalendar_event
+ */
+function zoom_helper_icalendar_event($event, $description) {
+    global $CFG;
+
+    // Match Moodle's uid format for iCal events.
+    $hostaddress = str_replace('http://', '', $CFG->wwwroot);
+    $hostaddress = str_replace('https://', '', $hostaddress);
+    $uid = $event->id . '@' .$hostaddress;
```
jrchamp, line 915: My fault. Looks like I forgot a space here.
```suggestion
    $uid = $event->id . '@' . $hostaddress;
```

jrchamp (dismissed):
Other than the one space I mentioned, I'm not seeing any issues. We'll probably want to do some testing on our side to see how this version behaves.

locallib.php
jwalits: Thanks for picking that up and your extensive review @jrchamp 

I have updated the space issue in locallib.php

abias:
Hi @jrchamp ,

I am grateful that we got everything sorted out now and this PR has been finally approved by you. Let's look forward to your local test results and to seeing this finally merged.

At this point, I would like to thank ETH Zürich who co-funded this development together with us at Ulm university as well as @jwalits from Catalyst IT who carried out the development work.

Cheers,
Alex

jrchamp (dismissed):
Pending user testing by @armccoy

abias:
Hi @jrchamp , hi @armccoy ,

may I ask if you have any news from your user testing of this patch? 
I would be grateful if this could be finally merged anytime soon.

If you need anything else from @jwalits or me, please let us know.

Cheers,
Alex

armccoy (dismissed):
Tested successfully. Looks good to me!

abias:
That's great news, @armccoy !
I am then looking forward to seeing this PR merged finally.

Thanks again for all your awesome work!

jrchamp (approved):
Meant to get this released yesterday, so we're going with version 2021081900.

PR #267, opened 2021-06-16
Issue 266

### LINKED ISSUE #266: Removing "viewdialin" permission hides entire invite
As discussed in issue #193, users without the "viewdialin" permission cannot see the meeting invite at all. Now that the feature has been added to strip the dialin info from the invite text, and the permission check is handled separately, we should allow users to see the rest of the invite text on the activity view page.

nstefanski:
This resolves issue #266 by removing the extra permissions check in view.php. The permissions check is now handled in the invitation class.

Fixes #266

### DIFF
--- a/view.php
+++ b/view.php
@@ -397,8 +397,7 @@ $table->data[] = array($strmuteuponentry, ($zoom->option_mute_upon_entry) ? $str
 
 // Show dial-in information.
 if (!$showrecreate
-        && ($zoom->option_audio === ZOOM_AUDIO_BOTH || $zoom->option_audio === ZOOM_AUDIO_TELEPHONY)
-        && ($userishost || has_capability('mod/zoom:viewdialin', $context))) {
+        && ($zoom->option_audio === ZOOM_AUDIO_BOTH || $zoom->option_audio === ZOOM_AUDIO_TELEPHONY) ) {
     // Get meeting invitation from Zoom.
     $meetinginvite = $service->get_meeting_invitation($zoom)->get_display_string($cm->id);
     // Show meeting invitation if there is any.
--- a/view.php
+++ b/view.php
@@ -397,7 +397,7 @@ $table->data[] = array($strmuteuponentry, ($zoom->option_mute_upon_entry) ? $str
 
 // Show dial-in information.
 if (!$showrecreate
-        && ($zoom->option_audio === ZOOM_AUDIO_BOTH || $zoom->option_audio === ZOOM_AUDIO_TELEPHONY) ) {
+        && ($zoom->option_audio === ZOOM_AUDIO_BOTH || $zoom->option_audio === ZOOM_AUDIO_TELEPHONY)) {
     // Get meeting invitation from Zoom.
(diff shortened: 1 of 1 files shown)


view.php
```diff
@@ -397,8 +397,7 @@
 
 // Show dial-in information.
 if (!$showrecreate
-        && ($zoom->option_audio === ZOOM_AUDIO_BOTH || $zoom->option_audio === ZOOM_AUDIO_TELEPHONY)
-        && ($userishost || has_capability('mod/zoom:viewdialin', $context))) {
+        && ($zoom->option_audio === ZOOM_AUDIO_BOTH || $zoom->option_audio === ZOOM_AUDIO_TELEPHONY) ) {
```
jrchamp, line 400: Small space nitpick
```suggestion
        && ($zoom->option_audio === ZOOM_AUDIO_BOTH || $zoom->option_audio === ZOOM_AUDIO_TELEPHONY)) {
```

jrchamp (approved):
This makes sense! The permissions checks for each component of the invitation are being checked by the get_display_string() method.

abias:
Unfortunately, I am too late to speak up here as the PR is already merged.

But anyway, I would like to say that this PR would have got my -1 and I would like to propose to revert this patch.

Here's why:
* The patch is changing the criteria when a user will see the 'Phone/Dial-In info' row in the 'Media' table on the Zoom activity overview page.
* The label 'Phone/Dial-In info' is quite clear: This row is about dial-in information. And the name of the capability `mod/zoom:viewdialin` is also quite clear: If a user does not have the `mod/zoom:viewdialin` capability, he should not see the dial in information.
* Before this patch, users who had the `mod/zoom:viewdialin` capability did see the 'Phone/Dial-In info' row. And the users who did not have this capability did not see it. This was fine and nobody was missing anything.
* After this patch, users who have the `mod/zoom:viewdialin` capability still see the 'Phone/Dial-In info' row and nothing has changed for them. But the users who do not have this capability see it as well. The dial-in snippet is just removed from the invitation text. This is not necessarily bad, but unnecessary
* Because now two cases have to be considered:
1. Users who do have the `mod/zoom:viewjoinurl` capability. These users see the join url in the invitation text. But they also see a row called 'Join link' which provides the same information. There is no benefit for them.
2. Users who do not have the `mod/zoom:viewjoinurl` capability. These users just see an invitation text without any useful information, just the topic and the date of the meeting. There is no benefit at all.

Coming to an end, from my point of view, the 'Phone/Dial-In info' should really just be shown to users who have the `mod/zoom:viewdialin` capability and to nobody else as it was contributed in #235 

Cheers,
Alex

jrchamp:
If we were to revert, I'd probably suggest removing the `$userishost` portion of the check. This would make the behavior rely more on the Moodle permissions, which should improve consistency.

@nstefanski What part of the invitation do you want them to see that is not already available in other sections? Alex contends that in the two cases that "There is not benefit [to the user]", so I want to understand. 😄

nstefanski:
Honestly, there is nothing specifically in the invitation that I need users to see, I discovered this when investigating the iCal issue (#270), it seemed like a bug so I flagged it, but I see @abias point and I'm fine with reverting it.

PR #274, opened 2021-06-30
Fix/meeting invitation issues

### LINKED ISSUE #270: iCal error if SIP or H.323 not found in invite
As mentioned in #193, if SIP or H.323 information are not found in the invite, and debugging messages are turned on, then the iCal download fails. This should generally only affect development environments when the institution's Zoom meeting plan does not include SIP or H.323, but can be recreated with the following steps:

1. In Site administration -> Development -> Debugging, set "Debug messages" to "DEVELOPER"
2. In Site administration -> Plugins -> Activity modules -> Zoom, set "SIP pattern" and/or "H.323 message pattern" to blank
3. Create a new Zoom meeting activity
4. Log in as a student or switch role to student and view the Zoom meeting activity
5. Click the "Download iCal" button and attempt to save the .ics file

Presumably this could be resolved by adding admin settings for whether SIP and H.323 are expected to be found in the meeting invite.

jrchamp:
If the regular expression is set to blank by the administrator, we should take it to mean that the Meeting Invitation template does not contain the target string and return successfully without modifying the invitation.

Fixes #270 
Also, reverts the change from #267 after the full discussion. The determination was that the purpose of that section is to display the dial-in information that is available from the meeting invitation.


abias:
I have skimmed over these changes and think they are all fine to revert #267 / solve #266 again as well as help to solve #270.

The only thing which came to my mind is that the proposed solution for #270 will be quite opaque to admins as 1. the regular expression for each invitation element is configured by default to contain the default regex pattern and 2. there is no information that the admin should clear the regex pattern to skip the cleaning.

How about adding something like 'If you want to skip a particular invitation element and output it as-is, please clear the corresponding pattern. This especially applies to setups where SIP or H.263 in Zoom' to the invitationregexenabled string?

andrewmadden:
I thought I'd just dump a little context about the debugging messages. As regex can be quite delicate to get working, I wanted to give some power to admins to be able to test if the regex is behaving as expected via debugging messages.

I had missed the case of the debugging messages preventing the ical download.

Having the field set to 'empty' to 'disable' checking for an element isn't the best UX as an empty field could result in any behaviour. It could mean that the element was not checked, or it could mean that it would match any string. It's not immediately obvious.

As an alternative, I propose adding an admin setting 'debugmode' that is checked to be true before logging the debug messages. This would also be an opportunity to explain that if enabled and using DEBUG_DEVELOPER, any errors or missing elements would prevent the ical download. If default to off, and hidden unless invitationregexenabled is true, this would prevent the debug messages interfering with functionality by default and still allow debug messages to help the admin if required.

Just my two cents, but I'll leave the decision to @abias and @jrchamp

jrchamp:
@armccoy For now, let's move forward with Alex's suggestion about adding some text to the description. Thanks @abias for the suggestion.

Thank you Andrew for the UX perspective. I'd like to move a little faster on this by only changing the text at this moment. I'm not opposed to a developer-friendly experience, but I'm not finding a simple solution that I feel good about.

rickbeasley (approved):
Looks good.

Issue #268, opened 2021-06-17
Make the identifying attribute configurable


abias:
Currently, the plugin uses the `email` user profile field to identify the Moodle user towards the Zoom API. This is currently hardcoded and probably fine for 99% of the usage scenarios.

However, especially in SSO setups and decentralized institution structures, it might happen that the `email` user profile field in Moodle does not contain the email address which is used to authenticate the user in Zoom. 

For scenarios like these, the Moodle administrator should able to select the identifying attribute from a list of Moodle user profile fields. This configured field will then be used for the communication with Zoom instead of the currently hardcoded `email` field.

Possible options for this new admin setting would be:
* Email address field (which is the status quo and the default)
* Moodle username field
* ID number field
* All custom profile fields of type 'text input' (i.e. fields which hold a string value)

jrchamp:
This makes sense for users who do not use the default unique identifier convention for Zoom (email). I've added the "help wanted" tag to indicate that we're open to assistance from the community as we do not use (and therefore have a way to test) this functionality.

abias:
Thank you, @jrchamp , for approving this enhancement proposal.

We have already contracted Catalyst IT to have a look into this issue and provide a pull request. We will also be able to test the code within a working Zoom instance.

PR #277, opened 2021-07-01
Making the identifying attribute configurable

jwalits:
Adding a patch to address issue raised in #268


abias:
Just for the record for the plugin maintainers:
The PR was submitted too early. We have to finish our internal review first.

PR #279, opened 2021-07-08
Issue #255. Removed timemodified update from populate_zoom_from_respo…

### LINKED ISSUE #255: populate_zoom_from_response() should not always return a modified timemodified value
In #253 , it has been detected that populate_zoom_from_response() always returns a modified timemodified value even if no field is changed.

During the review of #253, @jrchamp stated:

> I think the timemodified value is being set in the populate_zoom_from_response() function to simplify the number of places where the value needs to be set. It's not really a good thing, because it means that we have several places that are directly updating the zoom table. This is dangerous, because we might need to call rebuild_course_cache(), zoom_calendar_item_update() and/or zoom_grade_item_update() depending on what has changed. If you look at the individual cases, you'll see that this is not handled consistently. In fact, you'll see that the PHPUnit tests (which should only be skipping the remote API calls) use a custom code block instead of following the production code path.
> 
> [...] To simplify our ability to continue moving forward with "bite-size" changes, [...] a separate issue that tries to resolve the "consistent updates for the zoom table" problem should be created.

This is the issue to investigate possible improvements.

rickbeasley:
…nse; added to update_meetings task and zoom_add_instance

Fixes #255

### DIFF
--- a/classes/task/update_meetings.php
+++ b/classes/task/update_meetings.php
@@ -136,6 +136,7 @@ class update_meetings extends \core\task\scheduled_task {
                 }
 
                 if ($changed) {
+                    $newzoom->timemodified = time();
                     $DB->update_record('zoom', $newzoom);
 
                     // Show trace message.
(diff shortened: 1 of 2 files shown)


jrchamp (approved):
Looks great! Thank you for making sure to keep the "skip timemodified" check that was added in the update_meetings task, as that check is likely still needed.

Issue #265, opened 2021-06-16
Enabling scheduling privilege - user not able to create meeting


briannwar:
Hi.... I'm testing the latest version of the Zoom plugin (v3.7, 2021041300) on our Moodle 3.10.4 (2020110904) test site. The "Enable scheduling privilege option" is on (this is a new admin setting in 3.7). This is the test I'm conducting where I'm seeing a problem:

-- User A and User B are instructors in the Moodle course.
-- In his Zoom settings, User B has assigned scheduling privileges to User A
-- User A is logged in to Moodle and goes to add a Zoom Meeting instance to the course
-- User A sees that he can schedule the meeting for both User B and User A ("Yourself").
-- When User A tries to schedule for himself (UserA), he get's an error ("You cannot schedule for the specified user") -- see link below for screenshot.

![zoom problem](https://user-images.githubusercontent.com/85974283/122251624-860b7b80-ce7f-11eb-8a5c-0495f1f85b8f.png)

User A is able to create a Zoom meeting instance for User B. And User A is able to schedule a meeting when User B is not an instructor in the course (i.e., the "Schedule meeting for" option is not present). The error is produced when the "Schedule meeting for" option is present, and User A tries tries to create a meeting for himself.

Please let me know if there might be any settings that would address this problem, or if you have any questions. I can also submit an issue to the GitHub repo is that would help.

Thanks much... Brian Warling (UCSF)

briannwar:
Hi..... I'm assuming that this bug has been confirmed. I'm wondering if there are plans to fix this critical issue soon. 

Please let me know if I can provide any more information.  

Thanks... Brian Warling

jrchamp:
Hi @briannwar! Yes, the bug has been assigned. Once it is fixed and merged, we'll be planning a release.

To speed the process, it may help to have one piece of clarity: When does the error message first show up?
 * when the page first loads
 * when you select "Yourself"
 * when you try to save/create the meeting

Also, thank you for reporting this issue! We really appreciate it.

briannwar:
Hi @jrchamp .... Thanks for getting back on this. The error appears when you try to save/create the meeting (with "Yourself" selected). I hope this helps..... Brian

jrchamp:
[...]

PR #280, opened 2021-07-12
Making the identifying attribute configurable

### LINKED ISSUE #268: Make the identifying attribute configurable

jwalits:
Adding a patch that will address the issue raised in #268 

Fixes #268


classes/task/get_meeting_reports.php
```diff
@@ -34,6 +34,8 @@
 
 defined('MOODLE_INTERNAL') || die();
 
+require_once($CFG->dirroot.'/mod/zoom/locallib.php');
```
jrchamp, line 37:
```suggestion
require_once($CFG->dirroot . '/mod/zoom/locallib.php');
```

locallib.php
```diff
@@ -799,3 +799,74 @@ function zoom_get_alternative_host_array_from_string($alternativehoststring) {
...
+    if (isset($userobject->$field)) {
+        // If one of the standard user fields.
+        $identifier = $userobject->$field;
+    } else if (isset($userobject->profile[$field])) {
+        // If one of the custom user fields.
+        $identifier = $userobject->profile[$field];
+    }
+    if (empty($identifier)) {
+        // Fallback to email if the field is not set.
+        $identifier = $userobject->email;
```
jrchamp, line 868: If possible, it would be good to be specific about the user that is being passed in. Thus, I suggest the following:

```suggestion
function zoom_get_api_identifier($user) {
    // Get the value from the config first.
    $field = get_config('zoom', 'apiidentifier');

    $identifier = '';
    if (isset($user->$field)) {
        // If one of the standard user fields.
        $identifier = $user->$field;
    } else if (isset($user->profile[$field])) {
        // If one of the custom user fields.
        $identifier = $user->profile[$field];
    }
    if (empty($identifier)) {
        // Fallback to email if the field is not set.
        $identifier = $user->email;
```

This does require that each location with `zoom_get_api_identifier()` be changed to `zoom_get_api_identifier($USER)`, but I think that's probably a good thing. What are your thoughts?

jrchamp (commented):
I see where you're going with this. We're not going to be able to fully test the code as we do not use non-email identifiers.

There may be some more places, like Alternative Host and schedule_for that might need changes also. I can't know for sure, so we'll be relying on your testing when the identifying attribute is not set to email.

locallib.php
```diff
@@ -799,3 +799,74 @@ function zoom_get_alternative_host_array_from_string($alternativehoststring) {
...
+function zoom_get_api_identifier_fields() {
+    $options = [
+        'email' => get_string('email'),
+        'username' => get_string('username'),
+        'idnumber' => get_string('idnumber'),
+    ];
+
+    $userfields = zoom_get_user_profile_fields();
+    if (!empty($userfields)) {
+        $options = array_merge($options, $userfields);
```
jrchamp, line 834: We probably want to prioritize the core fields and not use any duplication that might exist in the profile fields.
```suggestion
        $options += $userfields;
```

locallib.php
jwalits: Sure, that makes sense.....I have refactored that function and updated all references

locallib.php
jwalits: done, I had the `+=` to start with....but went with `array_merge` - changed it back to `+=` now

classes/task/get_meeting_reports.php
jwalits: updated that

jwalits:
> I see where you're going with this. We're not going to be able to fully test the code as we do not use non-email identifiers.
> 
> There may be some more places, like Alternative Host and schedule_for that might need changes also. I can't know for sure, so we'll be relying on your testing when the identifying attribute is not set to email.

Yes, this part was tested internally before I submitted this PR and its working as expected.

mod_form.php
```diff
@@ -391,7 +392,7 @@ public function definition() {
                     $mform->addElement('checkbox', 'change_schedule_for', get_string('changehost', 'zoom'));
                     $mform->setDefault('schedule_for', strtolower($service->get_user($this->current->host_id)->email));
                 } else {
-                    $mform->setDefault('schedule_for', strtolower($USER->email));
+                    $mform->setDefault('schedule_for', strtolower($zoomapiidentifier));
```
jrchamp, line 395: Are the identifying attribute values case-insensitive? I'm wondering if calling strtolower() is safe.

mod_form.php
```diff
@@ -391,7 +392,7 @@ public function definition() {
                     $mform->addElement('checkbox', 'change_schedule_for', get_string('changehost', 'zoom'));
                     $mform->setDefault('schedule_for', strtolower($service->get_user($this->current->host_id)->email));
```
jrchamp, line 393: This is using the "email" value from Zoom. If the Zoom account is using a non-email identifying attribute, does this value contain that non-email identifier?

settings.php
```diff
@@ -42,7 +42,7 @@
         $errormessage = '';
         try {
             $service = new mod_zoom_webservice();
-            $service->get_user($USER->email);
+            $service->get_user(zoom_get_api_identifier());
```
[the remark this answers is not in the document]
jwalits: Thanks for picking that one up! I have updated that

mod_form.php
jwalits: I'm not sure about that, as I wasn't able to test that part of the code myself. Maybe @abias can shed more light on that from their testing?

mod_form.php
jwalits: Same for this. I wasn't able to test this with the zoom account I have access to. Again @abias might be the best person to confirm this with?

mod_form.php
abias: I am sorry that I missed this question to me up to now.

Good catch, @jrchamp .
I would say that most scenarios, the identifying attribute will be case-insensitive, but we can't be sure that this is a general rule.

My proposal would be to remove strtolower() here and instead enhance zoom_get_api_identifier() in a way that this function checks before returning the identifier if it is returning a valid emailadress as identifier and, if yes, pushes the identifier through strtolower() before returning it.

This way, we would keep the status quo (that email addresses will be lowercase) but will still support case-sensitive identifiers it this might be needed anywhere.

Would this make sense to you?


On another note, I also saw on https://github.com/ncstate-delta/moodle-mod_zoom/pull/280/files#diff-29c6ce88530e509a2308e09f3c5143503ccdbc032b59249a758978bda34703bfR296-R298 that there is a strtoupper() call. I understand this is needed for pure output formatting reasons. Is this right or do we need to have a look at these lines as well?

mod_form.php
abias: Good catch, @jrchamp .
I can't answer this yet, but I will double-check with our colleagues at ETH Zürich who are running this kind of setup.

mod_form.php
jrchamp: It may be that Zoom is always treating the identifying attribute as case-insensitive (which would make sense given the API's behavior when looking up a user).

The strtoupper() is probably a copy-paste from the name above it. It should probably have been a strtolower(). The purpose of both was to make it easier to match Zoom meeting participants with Moodle users. Notice how it's populating an `$emails` array and then comparing it to `user_email` field of the participant. Hopefully the field that comes back from Zoom uses the identifying attribute, but I don't see how that could work for users that are not part of the same Zoom enterprise account. It's probably all going to be best guess and some things simply won't work for "external" participants.

jrchamp: Oops, forgot to respond to your other question. Rather than test if the field "is valid email address" and have the code behave differently, it would be nice if we can continue treating it as case insensitive.

mod_form.php
jonlan: At ETH Zürich we have the following value for "Sign-In Email" in Zoom: In my case it is the "primary" E-Mail-Adress of the institution which is Linked to other E-Mail-Adresses.

Sign-In Email = jlanz@ethz.ch

jrchamp: @jonlan Thank you for sharing these details. I want to make sure that I understand. It sounds like you are saying "The primary email address is linked to other email addresses." Are you saying that multiple Zoom user accounts have the same "Sign-In email" address?

jonlan: No. Each user has his own and unique E-Mail-Adress in Zoom. But each user has several E-Mail-Adresses which are linked with this primary E-Mail-Adresse.
i.e
jlanz@ethz.ch (primary E-Mail-Adress, User ID in Moodle, matching attribute with Zoom)
jonas.lanz@let.ethz.ch (linked Adress, E-Mail-Adress in Moodle)

abias: @jrchamp , may I ask if @jonlan 's answer was helpful to you or if you need anything else from us before this PR can be integrated?

jrchamp: I think I understand: the Zoom email address field is still an email address, but in Moodle - for @jonlon's users - we need to match it against the username instead of the email field (because the email field contains a "friendly" email alias which does not match the Zoom account). If that's correct, it's a lot less complicated than I previously thought.

abias: Yes, that's my understanding as well, @jrchamp .

I also had to learn recently that my sentence 'Furthermore, Zoom can also be configured to process the given identifying attribute string, which isn't a full email address, in a way that it composes the given attribute string to a full email address' which I wrote in #268 was wrong. Such a configuration is not possible in Zoom. Thus I have removed that sentence from #268 .

If our understanding is the same as mine now, I would propose to merge this PR now, especially as it does not break existing installations and the general behaviour of the plugin.

If any hickups with the patch arise during production at ETH Zürich, they will report and fix them in another PR.

jrchamp (commented):
This is in good shape. Let's do some testing to make sure it won't cause issues for existing users, particularly those with scheduling privilege. In fact, it would probably be good for whomever tests #265 to also test this as it modifies the relevant code.

jonlan:
Hello everybody

From my point of view the code seems OK though we run the master already on a test environment.

We really need this plugin until Thursday, 26.08.2021 in a published version.

Could you please merge this pull request.

Thank you
Jonas

jrchamp (approved):
All set. Marking approved.

PR #285, opened 2021-07-20
loadmeeting: Only real user should use start_url

### LINKED ISSUE #284: Revert treating alternative hosts as hosts in loadmeeting.php
Recently, we deployed v3.8 of this plugin which contains the change from #275 / #257.

Today, we received a bug report where a teacher described this scenario:
* User A has created a Zoom activity in Moodle
* User A added user B as alternative host into this Zoom activity
* User B logged into Moodle on his own computer
* User B went to the Zoom activity in Moodle and clicked the 'Start meeting' button
* As a result, the Zoom client opened and user B was shown with the name and avater _of user A_ in the Zoom client.

During a support session, we could reproduce this problem with another meeting and other users.

Digging further, I understood now that the start_url contains a `zak` parameter which basically seems to be a login token for the host user. This assumption is backed by the post on https://devforum.zoom.us/t/once-we-start-zoom-meeting-from-start-url-as-host-we-are-able-to-access-as-host-even-though-from-join-url-automatically/22305/2, for example.

And reading again on https://marketplace.zoom.us/docs/api-reference/zoom-api/meetings/meeting, I read:

> The start_ur of a Meeting is a URL using which a host or an alternative host can start the Meeting.

but I also read

> This URL should only be used by the host of the meeting and should not be shared with anyone other than the host  of the meeting as anyone with this URL will be able to login to the Zoom Client as the host of the meeting.

Having discovered that, I went ahead and reverted the loadmeeting.php part of #275 as a hotfix in production:
```
diff --git a/mod/zoom/loadmeeting.php b/mod/zoom/loadmeeting.php
index 9fa0c07..f97adb1 100644
--- a/mod/zoom/loadmeeting.php
+++ b/mod/zoom/loadmeeting.php
@@ -39,9 +39,7 @@ if ($id) {
 } else {
     print_error('You must specify a course_module ID');
 }
-$userisrealhost = (zoom_get_user_id(false) == $zoom->host_id);
-$alternativehosts = zoom_get_alternative_host_array_from_string($zoom->alternative_hosts);
-$userishost = ($userisrealhost || in_array($USER->email, $alternativehosts, true));
+$userishost = (zoom_get_user_id(false) == $zoom->host_id);
 
 require_login($course, true, $cm);
```
This brought the odd behaviour to an end.

@jrchamp , I would encourage you to revert this part in a hotfix release of the plugin as well as soon as possible.

In a further step, we could investigate if it makes any difference and sense to send the alternative hosts to the start_url *without the zak parameter*, but this will need more time.

jrchamp:
Fixes #284

### DIFF
--- a/loadmeeting.php
+++ b/loadmeeting.php
@@ -39,7 +39,7 @@ if ($id) {
 } else {
     print_error('You must specify a course_module ID');
 }
-$userisrealhost = (zoom_get_user_id(false) == $zoom->host_id);
+$userisrealhost = (zoom_get_user_id(false) === $zoom->host_id);
 $alternativehosts = zoom_get_alternative_host_array_from_string($zoom->alternative_hosts);
 $userishost = ($userisrealhost || in_array($USER->email, $alternativehosts, true));
 
@@ -65,7 +65,8 @@ if ($available !== true) {
     redirect($unavailabilityurl, $unavailabilitynote, null, \core\output\notification::NOTIFY_ERROR);
 }
 
-if ($userishost) {
+if ($userisrealhost) {
+    // Important: Only the real host can use this URL, because it joins the meeting as the host user.
     $nexturl = new moodle_url($zoom->start_url);
 } else {
     // Check whether user had a grade. If no, then assign full credits to him or her.


jrchamp:
Before merging, decide if we want to tag the new version in this PR or a separate one, and make sure to thank `abias` for reporting.

smbader (approved):

PR #290, opened 2021-08-05
Fix/289 zoom intro

### LINKED ISSUE #190: Improvement: Remove HTML tags in meeting description when sending the description to Zoom
Hi Rex,

currently, mod_zoom sends the meeting activity title and the meeting activity description to Zoom and fills the meeting topic and the meeting description with this data.

However, descriptions in Moodle are rich-text fields and in Zoom they are not.
Because of that, you will end up with strange Zoom meeting descriptions as soon as you start using rich text in Moodle meeting descriptions:

grafik

To avoid this, I would like to propose this improvement:
When sending the meeting description from Moodle to Zoom, always remove all HTML tags and just send a plain text description.

Cheers,
Alex

### LINKED ISSUE #196: Description field in a Zoom learning activity and passcode
We experience an error where the description field in a Zoom learning activity, changes over time.

If we use bullet points, hours or days can pass and suddenly they are gone. See pictures of before and after:
image

image

If we insert an image in the description then the picture disappears.

We also experience that some students on a flow cant join Zoom without the passcode, is it possible to do so that we can choose whether a passcode should be required or not?

### LINKED ISSUE #289: Emoji in meeting description get lost
[also see [https://moodle.org/mod/forum/discuss.php?d=425347](https://moodle.org/mod/forum/discuss.php?d=425347)]
I want to use emoji in meeting descriptions (e.g. country flags to indicate different time zones). Everything works fine, until the meetings are updated through the `\mod_zoom\task\update_meetings` scheduled task which replaces the emoji with hyphens:
`=> Field "intro" has changed from "<p dir="ltr" style="text-align: left;"><strong>Introduction</strong></p><p dir="ltr" style="text-align: left;">�&nbsp;17 janvier 2022</p><p dir="ltr" style="text-align: left;">��&nbsp;09:00-11:00</p><p dir="ltr" style="text-align: left;">��&nbsp;10:00-12:00<br></p>" to "<p dir="ltr" style="text-align: left;"><strong>Introduction</strong></p><p dir="ltr" style="text-align: left;">-&nbsp;17 janvier 2022</p><p dir="ltr" style="text-align: left;">--&nbsp;09:00-11:00</p><p dir="ltr" style="text-align: left;">--&nbsp;10:00-12:00<br></p>"`
Apparently Zoom doesn't support emoji in the meeting description and thus they are lost when reading it back from Zoom.
Workaround could be to disable the scheduled task (as long as you don't need to change the meeting information outside of Moodle).
A better solution might be to offer an option not to sync the description with Zoom in the first place (it seems a bit problematic anyway, as Zoom doesn't support any HTML formatting which Moodle inserts by default).

smbader:
The proposal is to not update the Moodle intro when Zoom sends a response. Zoom removes any special characters and styling breaking the Moodle intro created by the instructor. And also, to send Zoom a more friendly agenda.

Fix #289 
Fix #190 
Fix #196

### DIFF
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -507,7 +507,7 @@ class mod_zoom_webservice {
             )
         );
         if (isset($zoom->intro)) {
-            $data['agenda'] = $zoom->intro;
+            $data['agenda'] = content_to_text($zoom->intro, FORMAT_MOODLE);
         }
         if (isset($CFG->timezone) && !empty($CFG->timezone)) {
             $data['timezone'] = $CFG->timezone;
--- a/lib.php
+++ b/lib.php
@@ -190,9 +190,6 @@ function populate_zoom_from_response(stdClass $zoom, stdClass $response) {
     }
     $newzoom->meeting_id = $response->id;
     $newzoom->name = $response->topic;
-    if (isset($response->agenda)) {
-        $newzoom->intro = $response->agenda;
-    }
     if (isset($response->start_time)) {
         $newzoom->start_time = strtotime($response->start_time);
     }


jrchamp:
To tester: Make sure that the user can set and update the intro text in Moodle, verifying each time that the updates are set to Zoom. The other test that I can think of would be to test that the task that updates meetings does not bring in the value from Zoom.

jrchamp (commented):
Code made sense. If someone can test that it all works, should be approval worthy.

abias:
I had a quick loook at this PR and, while I haven't tested it, understand what the code does and what the general approach is.

However, I am wondering if this resulting behaviour is what users will expect. 

Currently, Moodle synchronizes the description to Zoom _and back_ with the result that you can change the description on each side. This is comfortable for hosts who use plain text descriptions, but this is problematic for hosts who use rich text descriptions as they get broken when they are synced back.

With this change, the behaviour would change in a way that
1. you can now use rich text descriptions in Moodle and syncing them to Zoom and back won't break their presentation in Moodle but
2. you can now change the description still in Zoom and you are nowhere informed that the change does not get synced back to Moodle. This will, in the worst case, result in two forks of the description which contain out-of-sync information.

As a slight improvement, I would propose to extend the populate_zoom_from_response() function in a way that it pushes the current description which is stored in Moodle through content_to_text() and compares the input with the output. If they are equal (and I am not wrong), the host did not use special markup in Moodle and thus Zoom didn't have the need to drop any special markup. In this case, it should still be save to update the description in Moodle with the reponse from Zoom if the host did update the description there.

This still won't be perfect, but it will be better than not updating the description at all in Moodle anymore in my point of view.

Cheers,
Alex

jrchamp:
Thank Alex for taking a look. I fully understand your perspective and I largely agree. However...

Philosophically, if the Zoom activity is being created in Moodle, then it should be primarily managed in Moodle. Ideally, I would like to be able to have the Moodle and Zoom descriptions match, but it's more and more clear that Zoom is going to do what Zoom wants as far as allowed characters and HTML tags. So, my concern is that extended characters, like emoji will be removed by Zoom and thus prevent #289 from being fixed. Additionally, even if it works *today*, I can't guarantee that Zoom won't break it *tomorrow* and thus overwrite all affected descriptions across all users of this plugin.

abias:
Thanks for making your point clear, @jrchamp .
I think I could live with the changes of this PR in our own usage, I just wanted to make clear that it introduces a behaviour change which might be surprising for hosts.

jrchamp (approved):
This looks great! Should address a few open issues.

PR #291, opened 2021-08-06
Fix/212 Fix DB schema check error

### LINKED ISSUE #212: Fix DB schema check error
I've first installed Zoom plugin 3.3 on my Moodle sites. Since then, i've updated them to Zoom 3.4.

Now, when i use /admin/cli/check_database_schema.php i see these problems for Zoom :
 * column 'option_host_video' has default '1', expected '0' (I)
 * column 'option_participants_video' has default '1', expected '0' (I)
 * column 'option_mute_upon_entry' has default '0', expected '1' (I)

I think there is a difference between install files, and update files, and db/upgrade.php file should certainly be fixed for that.

It seems to be the opposite of #141# https://github.com/ucla/moodle-mod_zoom/pull/141

rickbeasley:
Fix DB schema check error, tested with 3.3 -> 3.4 upgrade.

Fixes #212


jrchamp (commented):
These changes make sense as far as "what should have been done originally", but my concern is that this may not fix the issue for users who have already upgraded to a newer version. Maybe we should bump the plugin version and use that version "date" for these changes?

rickbeasley:
For users who have already upgraded to a newer version, second commit adds these updates to a version dated for the 8/19/2021 release. Note I didn't bump the version in version.php since last time I think we did that part of the release process, and I wasn't sure if I should.

jrchamp (approved):
The merge conflict delightfully allowed me to fix the plugin version that was used in db/upgrade.php for #258

PR #293, opened 2021-08-17
Make the Zoom API URL configurable in the connection settings - Fixes #261

### LINKED ISSUE #261: Make the Zoom API Url configurable in the Connection settings
Hi,
For GDPR reason, we would like to connect the plugin to European Zoom servers.
https://eu01api-www4local.zoom.us for example.
Is it possible to make this parameter configurable ?
Kind regards,
Didier

abias:
This patch makes the Zoom API configurable in the Zoom activity connection settings.
Admins can choose between the global API endpoint and the EU API endpoint. The URLs of both endpoints remain hardcoded in the plugin code.

Fixes #261


lang/en/zoom.php
```diff
@@ -42,6 +42,10 @@
 $string['allmeetings_enable'] = 'Enable all meetings link';
 $string['apikey'] = 'Zoom API key';
 $string['apikey_desc'] = '';
+$string['apiendpoint'] = 'Zoom API Endpoint';
+$string['apiendpoint_desc'] = 'With this setting, you control to which Zoom API endpoint the Zoom activity will connect to. The global API endpoint is working in all setups. The EU API endpoint is intended for Zoom customers only who have a Zoom license with the EU provisioning option. If you have any doubts, please leave the setting set to the global API endpoint.';
```
jrchamp, line 46: There were a few words that I wanted to adjust. Please check my proposed wording in case I have used words or phrases that do not easily/directly translate.
```suggestion
$string['apiendpoint_desc'] = 'Choose which Zoom API endpoint the Zoom activity will use to connect. The global API endpoint should work for all users. The European Union (EU) API endpoint is only intended for users who have a Zoom license with the EU provisioning option. If you are unsure, use the global API endpoint.';
```

classes/webservice.php
```diff
@@ -176,7 +170,8 @@ protected function _get_curl_object() {
      */
     protected function _make_call($path, $data = array(), $method = 'get') {
         global $CFG;
-        $url = self::API_URL . $path;
+        $baseurl = zoom_get_api_url();
+        $url = $baseurl . $path;
```
jrchamp, line 174: This is called every time now. I'm not sure how I feel about that. It's probably fine, because Moodle probably caches the values in get_config(). I guess I may have expected the `__construct()` method to initialize the `$this->apiurl` protected member variable. What do you think would be best?

jrchamp (commented):
Thanks @abias!

This makes a lot of sense to me. I've added a few comments, but overall this seems straightforward.

locallib.php
```diff
@@ -799,3 +804,28 @@ function zoom_get_alternative_host_array_from_string($alternativehoststring) {
...
+ * Get the configured Zoom API URL.
+ *
+ * @return string The API URL.
+ */
+function zoom_get_api_url() {
+    // Get the API endpoint setting.
+    $apiendpoint = get_config('zoom', 'apiendpoint');
+
+    // Pick the corresponding API URL.
+    switch($apiendpoint) {
```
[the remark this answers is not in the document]
abias: Sure, I will be happy to change that.
It's just interesting that Moodle code checker didn't complain about it in the CI tests.

locallib.php
```diff
@@ -799,3 +804,28 @@ function zoom_get_alternative_host_array_from_string($alternativehoststring) {
...
+        case ZOOM_API_ENDPOINT_EU:
+            $apiurl = ZOOM_API_URL_EU;
+            break;
```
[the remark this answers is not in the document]
abias: I am sorry, but I struggle to recognize what you want to have changed here?

lang/en/zoom.php
abias: As I am not a native speaker, I always appreciate proof-readings.
I will be happy to adopt your changes.

classes/webservice.php
abias: Well, as you said, the values in get_config() are cached, so this shouldn't have a huge performance impact.
However, as this is a class which already stores the API key and secret in a class property, we can do that with the API URL likewise.

I will update the code.

locallib.php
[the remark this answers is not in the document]
abias: Ah, I must have been blind. You want a blank line.
I will add this.

abias:
Hi @jrchamp ,

thanks for your review.
I have made all requested changes, have rebased the branch onto the latest master branch and have force-pushed it.

Cheers,
Alex

jrchamp (approved):
This all looks right to me.

PR #305, opened 2021-09-08
Mobile app patch

### LINKED ISSUE #304: Moodle mobile app does not trigger join_meeting_button_clicked event
Discovered this while testing event listeners, I've created a patch that fixes the issue.

nstefanski:
This addresses issue #304 

I added a host check as well because it is used by the event. However, in testing, I noticed that the start url does not automatically redirect to Zoom's mobile app like the join url does, so I reverted to using the join url for both hosts and participants. I left the two separate in case Zoom modifies the start url behavior in the future so we can more easily make use of it.

### DIFF
--- a/classes/external.php
+++ b/classes/external.php
@@ -228,7 +228,6 @@ class mod_zoom_external extends external_api {
         \mod_zoom\event\join_meeting_button_clicked::create(array('context' => $context, 'objectid' => $zoom->id, 'other' =>
             array('cmid' => (int) $cm->id, 'meetingid' => (int) $zoom->meeting_id, 'userishost' => $userishost)))->trigger();
 
-
         // Pass url to join zoom meeting in order to redirect user.
         $result = array();
         $result['status'] = true;
--- a/classes/external.php
+++ b/classes/external.php
@@ -197,7 +197,8 @@ class mod_zoom_external extends external_api {
         $userishost = (zoom_get_user_id(false) == $zoom->host_id);
 
         if ($userishost) {
-            $joinurl = new moodle_url($zoom->start_url, array('uname' => fullname($USER)));
+            // Start URL will not launch Zoom app, so using join url instead.
+            $joinurl = new moodle_url($zoom->join_url, array('uname' => fullname($USER)));
         } else {
             // Check whether user had a grade. If no, then assign full credits to him or her.
             $gradelist = grade_get_grades($course->id, 'mod', 'zoom', $cm->instance, $USER->id);
--- a/classes/external.php
+++ b/classes/external.php
@@ -175,6 +175,7 @@ class mod_zoom_external extends external_api {
     public static function grade_item_update($zoomid) {
         global $DB, $CFG, $USER;
         require_once($CFG->dirroot . "/mod/zoom/lib.php");
+        require_once($CFG->dirroot . "/mod/zoom/locallib.php");
         require_once($CFG->libdir . '/gradelib.php');
 
         $params = self::validate_parameters(self::get_state_parameters(),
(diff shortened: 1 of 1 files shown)


classes/external.php
```diff
@@ -193,30 +194,41 @@ public static function grade_item_update($zoomid) {
...
-            $grademax = $gradelist->items[0]->grademax;
-            $grades = array('rawgrade' => $grademax,
-                            'userid' => $USER->id,
-                            'usermodified' => $USER->id,
-                            'dategraded' => '',
-                            'feedbackformat' => '',
-                            'feedback' => '');
-            // Call the zoom/lib API.
-            zoom_grade_item_update($zoom, $grades);
+        $userishost = (zoom_get_user_id(false) == $zoom->host_id);
```
jrchamp, line 197: This is the logic for `$userisrealhost`. If we're trying to consistently mark `$userishost`, we'll need to include alternative hosts.

classes/external.php
```diff
@@ -193,30 +194,41 @@ public static function grade_item_update($zoomid) {
...
+
+        if ($userishost) {
+            // Start URL will not launch Zoom app, so using join url instead.
+            $joinurl = new moodle_url($zoom->join_url, array('uname' => fullname($USER)));
```
jrchamp, line 201: This join URL code is the same as the same as the later join URL code.

classes/external.php
```diff
@@ -193,30 +194,41 @@ public static function grade_item_update($zoomid) {
...
+                                'userid' => $USER->id,
+                                'usermodified' => $USER->id,
+                                'dategraded' => '',
+                                'feedbackformat' => '',
+                                'feedback' => '');
+                // Call the zoom/lib API.
+                zoom_grade_item_update($zoom, $grades);
+            }
+
+            $joinurl = new moodle_url($zoom->join_url, array('uname' => fullname($USER)));
```
jrchamp, line 220: Let's move this outside of the conditional block as the code is the same.

classes/external.php
```diff
@@ -193,30 +194,41 @@ public static function grade_item_update($zoomid) {
...
 
         // Track completion viewed.
         $completion = new completion_info($course);
         $completion->set_module_viewed($cm);
 
-        // Pass url to join zoom meeting in order to redirect user.
-        $joinurl = new moodle_url($zoom->join_url, array('uname' => fullname($USER)));
+        // Record user's clicking join.
+        \mod_zoom\event\join_meeting_button_clicked::create(array('context' => $context, 'objectid' => $zoom->id, 'other' =>
+            array('cmid' => (int) $cm->id, 'meetingid' => (int) $zoom->meeting_id, 'userishost' => $userishost)))->trigger();
```
jrchamp, line 229: This code makes sense, but it looks like this is 90% the same as the loadmeeting.php code. Should we perhaps consolidate this into a locallib.php function so that we can reduce inconsistencies in the code as a result of partial duplication?

classes/external.php
```diff
@@ -193,30 +194,41 @@ public static function grade_item_update($zoomid) {
...
+                $grademax = $gradelist->items[0]->grademax;
+                $grades = array('rawgrade' => $grademax,
+                                'userid' => $USER->id,
+                                'usermodified' => $USER->id,
+                                'dategraded' => '',
+                                'feedbackformat' => '',
+                                'feedback' => '');
+                // Call the zoom/lib API.
+                zoom_grade_item_update($zoom, $grades);
+            }
```
jrchamp, line 218: It seems possible that a student or student team might - for a flipped classroom activity - be the/a host. If the activity is a graded activity, is it a problem for the host to get grade credit?

jrchamp (commented):
Thanks @nstefanski for working on this. I agree that this makes the mobile code match more closely with the loadmeeting.php code. However, I think that highlights a few remaining inconsistencies that are likely due to a historical duplication and subsequent diverging of the code.

classes/external.php
nstefanski: I realize I was looking at an older version of loadmeeting.php for this code, I can update this

classes/external.php
nstefanski: that's completely possible, I was just mirroring the loadmeeting.php code which does not give a grade

nstefanski:
@jrchamp your comments make sense, I'll take another crack at this with a shared library function

PR #312, opened 2021-10-01
jwt: v5.4.0

jrchamp:
Pull in the most recent set updates from upstream and use the current method for ignoring files in phpcs


armccoy (approved):
Looks good to me!

PR #316, opened 2021-10-08
start_url: Always use a fresh copy if possible

### LINKED ISSUE #313: Moodle and Zoom disconnected
@jrchamp  Hi, please a need your help!

I can't login through Moodle after a while and it asks me to join again through ZOOM. It seems to have disconnected. How can I solve it?
Apparently it has something to do with the API token.

Thanks a lot!

Syl.

jrchamp:
The start_url expires after a couple hours, so it makes sense to me to get a fresh copy when it is needed.

As long as this works as expected, step two of this pull request would be to remove start_url from the database. At most, it could live in a time-limited cache, but I'm not sure that potential improvement is worth the overhead.

### DIFF
--- a/index.php
+++ b/index.php
@@ -56,7 +56,6 @@ $stractions = get_string('actions', 'mod_zoom');
 $strsessions = get_string('sessions', 'mod_zoom');
 
 $strmeetingstarted = get_string('meeting_started', 'mod_zoom');
-$strstart = get_string('start', 'mod_zoom');
 $strjoin = get_string('join', 'mod_zoom');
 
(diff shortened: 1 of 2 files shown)


jrchamp:
Also, you'll want to make sure that you are not logged in to Zoom via the web and the desktop client. If you are already logged in, it will let you hop in anyway, but a valid start_url should avoid the need to be logged in.

sgrandh3:
When i did not log into Zoom via web/client , on clicking the start meeting, it let me join the meeting, without the need of logging in.

smbader (approved):

PR #319, opened 2021-10-11
calendar: Synchronize events consistently

### LINKED ISSUE #318: Timeline - Events not get removed
Zoom Meeting Event entries are not getting deleted from the moodle timeline block.
1. Create a Zoom Meeting as recurring meeting with recurrence being "no fixed time". Then there wont be any entries made to the timeline block, which is fine.
2. Set Recurrence to weekly e.g. Monday 10:00 ending a few months in the future. Then then timeline block will list those weekly meetings.
3. Set Recurrence back to "no fixed time" and the previously listed meetings are still shown in the timeline block, which is wrong.
4. Deleting the meeting will remove all entries regarding that meeting from the timeline block.

It seems as if the zoom moodle plugin doesnt propagate the events correctly to the timeline block.

jrchamp:
The original code will only process changes to existing events if there are "new" events. Editing a Zoom activity to move it between the three types (single event, recurring no specific occurrences, recurring with specific occurrences) should clean up and synchronize appropriately.

The new code will:
 - Identify all of the events that should exist
 - Look up all of the events that do exist
 - Compare these two lists, creating/updating/deleting as needed

Fixes #318

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -443,63 +443,53 @@ function zoom_get_extra_capabilities() {
  */
 function zoom_calendar_item_update(stdClass $zoom) {
     global $CFG, $DB;
-    require_once($CFG->dirroot.'/calendar/lib.php');
+    require_once($CFG->dirroot . '/calendar/lib.php');
 
+    // Based on data passed back from zoom, create/update/delete events based on data.
+    $newevents = array();
     if (!$zoom->recurring) {
-        $eventid = $DB->get_field('event', 'id', array(
-            'modulename' => 'zoom',
-            'instance' => $zoom->id,
-        ));
-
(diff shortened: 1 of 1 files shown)


sgrandh3:
@jrchamp, i verified it and it looks good. Did add all the 3 kinds of meetings and viewed it in the "All meetings". They were displayed as expected.
image

smbader (approved):

abias:
> @jwalits @abias If you are available to take a look, I would appreciate it. You two are the most familiar with this code outside of our organization.

I am sorry that I didn't respond to that ping. However, seeing that the PR has been merged in the meantime, I assume that everything was ok and the problem was solved correctly.

Thanks for your great work!

PR #328, opened 2021-11-07
restore defunct meetings without failing

jonof:
It has been observed that the `zoom_not_found_exception` (and conceivably others) can be thrown by `mod_zoom_webservice::create_meeting()` during restore of a course backup, so rather than having it unhelpfully fail the whole course restore by the exception not being handled, consider the restored Zoom instance as expired. Additionally, if a meeting was successfully recreated, ensure it is indicated as such rather than keeping its old, possibly expired state.

### DIFF
--- a/backup/moodle2/restore_zoom_stepslib.php
+++ b/backup/moodle2/restore_zoom_stepslib.php
@@ -62,15 +62,17 @@ class restore_zoom_activity_structure_step extends restore_activity_structure_st
         $data->start_time = $this->apply_date_offset($data->start_time);
 
         // Either create a new meeting or set meeting as expired.
-        $updateddata = $service->create_meeting($data);
-        if (!$updateddata) {
-            $updateddata = new stdClass;
-            $updateddata->exists_on_zoom = ZOOM_MEETING_EXPIRED;
+        try {
+            $updateddata = $service->create_meeting($data);
+            $data = populate_zoom_from_response($data, $updateddata);
+            $data->exists_on_zoom = ZOOM_MEETING_EXISTS;
+        } catch (moodle_exception $e) {
+            $data->start_url = '';
+            $data->join_url = '';
+            $data->meeting_id = 0;
+            $data->exists_on_zoom = ZOOM_MEETING_EXPIRED;
         }
 
-        $data->start_url = $updateddata->start_url;
-        $data->join_url = $updateddata->join_url;
-        $data->meeting_id = $updateddata->id;
         $data->course = $this->get_courseid();
 
         if (empty($data->timemodified)) {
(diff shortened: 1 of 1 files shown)


backup/moodle2/restore_zoom_stepslib.php
```diff
@@ -88,8 +90,7 @@ protected function process_zoom($data) {
 
         // Create the calendar events for the new meeting.
         $data->id = $newitemid;
-        $zoom = populate_zoom_from_response($data, $updateddata);
-        zoom_calendar_item_update($zoom);
+        zoom_calendar_item_update($data);
```
jrchamp, line 93: I've not done as much with the restore code. Do you know if `apply_activity_instance` will update the grade item or do we need to call `zoom_grade_item_update($data);` here?

jrchamp (commented):
Thank you for identifying this issue and creating this pull request. I have a couple quick questions.

backup/moodle2/restore_zoom_stepslib.php
```diff
@@ -62,15 +62,17 @@ protected function process_zoom($data) {
...
-            $updateddata = new stdClass;
-            $updateddata->exists_on_zoom = ZOOM_MEETING_EXPIRED;
+        try {
+            $updateddata = $service->create_meeting($data);
+            $data = populate_zoom_from_response($data, $updateddata);
+            $data->exists_on_zoom = ZOOM_MEETING_EXISTS;
+        } catch (moodle_exception $e) {
+            $data->start_url = '';
+            $data->join_url = '';
+            $data->id = 0;
```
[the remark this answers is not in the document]
jonof: Indeed, it is supposed to be.

backup/moodle2/restore_zoom_stepslib.php
jonof: `apply_activity_instance` does course module id old–new mapping. My understanding is grade items are handled elsewhere in the restore machinery, and generally with backup-restore work my guiding principle is: What does Forum do?

jrchamp: Good point. In looking at Forum, grade item titles shouldn't have changed and would be likely be handled separately.

jrchamp:
Thanks @jonof 

This looks great to me. We'll do a manual test and I'm hoping to merge it into the next release.

jonof:
Fault amended. (The perils of transcribing patches over several iterations.)

It's probably also worth stating for the record that I couldn't see a reasonable way for `mod_zoom_webservice::create_meeting()` to do anything other than return a response object or throw an exception, hence why I moved the `(!$updateddata)` business into the catch block, and it seemed safe to set both the join and start URLs to empty strings, and the meeting ID to 0.

You can also fake an error condition by inserting a `throw new zoom_not_found_exception('a','b');` line before the call to _create_meeting_ in the _try_ block.

jrchamp (approved):
This is great! Ready to merge.

PR #330, opened 2021-11-16
Use 'title' as the Zoom event name label

### LINKED ISSUE #310: Redundant column names in "All Meetings"
Both the Moodle course topic and the Zoom meeting topic has the same column name as "Topic"
image

armccoy:
Fixes #310 by using 'title' as the Zoom event name label, since 'topic' is already used throughout Moodle to indicate course sections.


jrchamp (approved):
Looks great! It looks like report.php already uses the title string for the column header, so this makes things more consistent.

PR #332, opened 2021-11-17
Fix for deprecated PHP 8 warning

### LINKED ISSUE #331: PHP 8: Required parameter follows optional parameter
Starting in PHP >= 8.0, a warning is thrown:

> ( ! ) Deprecated: Required parameter $datatoget follows optional parameter $data in [...]/mod/zoom/classes/webservice.php on line 302

This is because the function's signatures expects a required parameter after a previous optional parameter. This makes no sense, since the mandatory `$datatoget`param in effect makes all previous optional parameters mandatory.

Offending code is in https://github.com/ncstate-delta/moodle-mod_zoom/blob/master/classes/webservice.php#L302

ndunand:
Hello, this fixes #331

### DIFF
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -299,7 +299,7 @@ class mod_zoom_webservice {
      * @see make_call()
      * @link https://zoom.github.io/api/#list-users
      */
-    private function make_paginated_call($url, $data = array(), $datatoget) {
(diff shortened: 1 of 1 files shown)


jrchamp (approved):
Thanks @ndunand for bringing this to our attention! All of the calls to this private method are passing a value, so removing the optional nature of the `$data` parameter looks good to us!

PR #335, opened 2021-11-25
Fix issue 334 key for uuid

### LINKED ISSUE #334: Database schema is not aligned - index for uuid in zoom_meeting_details table
I found 'Database schema is not aligned.' error in 'Database schema check' in 'Performance overview' from one of our client. 
zoom_meeting_details
Unexpected index 'mdl_zoommeetdeta_uui_uix'.

I found the codes the index included in upgrade.php, but not install.xml
https://github.com/ncstate-delta/moodle-mod_zoom/blob/3a2d1c2784c63ed83881954bd2e79b50de265c9f/db/install.xml#L70
https://github.com/ncstate-delta/moodle-mod_zoom/blob/3a2d1c2784c63ed83881954bd2e79b50de265c9f/db/upgrade.php#L215

So it probably like:
The mod_zoom is installed before 15 Sep 2018.
Upgrade script added the index.
Schema check process finds difference between install.xml and db schema.

Do we need the index for uuid in zoom_meeting_details table?
If yes, can you add it into install.xml?
If no, I'll remove the index from the our db.

TomoTsuyuki:
Fix the index issue #334 
An index `uuid_unique` was added at ver 2018091200 in `upgrade.php`, but not in `install.xml`.
This patch provides the index in the `install.xml`, and `upgrade.php` in case the plugin is already installed.


jrchamp (commented):
The code looks great to me. Thank you for getting this done so quickly.

TomoTsuyuki:
Thanks for your review @jrchamp .
Let me know if I need to do something else.
I'm waiting for being integrated with the master.

sgrandh3:
Tested by hosting a meeting with Jonathan(Thank you so much for walking through @jrchamp )and was able to view the session report. As expected, sessions did not show up when there are no participants.

jrchamp (approved):
Testing succeeded. I have resolved the conflicts, so this is ready for merge!

PR #337, opened 2021-11-26
tracking fields: always synchronize by coalescing to empty array

### LINKED ISSUE #336: PHP notices/warnings after  the integration of the tracking fields feature
After having upgraded to the latest release of mod_zoom which has #308 / #208 integrated, we are getting these debug warnings in the output of our Moodle cron:

```
PHP Notice:  Undefined property: stdClass::$tracking_fields in /var/www/html/moodle_prod/mod/zoom/classes/task/update_meetings.php on line 164
PHP Warning:  Invalid argument supplied for foreach() in /var/www/html/moodle_prod/mod/zoom/locallib.php on line 1172
```
We do not plan to use the tracking feature and haven't configured anything for this feature during the upgrade.

I would be grateful if you could have a look to see if you can get rid of these debug warnings.

jrchamp:
@haietza Please check to make sure I'm not forgetting something. This may also resolve an issue where a tracking field is added to a meeting (and synchronized to Moodle) and then later removed (but not synchronized to Moodle because there are no fields).

Fixes #336

### DIFF
--- a/classes/task/update_meetings.php
+++ b/classes/task/update_meetings.php
@@ -161,7 +161,7 @@ class update_meetings extends \core\task\scheduled_task {
 
                 // Update tracking fields for meeting.
                 mtrace('  => Updated tracking fields for Zoom meeting ID ' . $zoom->meeting_id);
-                zoom_sync_meeting_tracking_fields($zoom->id, $response->tracking_fields);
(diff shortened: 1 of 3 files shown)


haietza:
@jrchamp Looks good to me! Thanks for taking care of this!

smbader (approved):
Today I learned about the Null Coalescing Operator

PR #338, opened 2021-11-26
webinar: Treat Zoom Events licenses as webinar licenses

### LINKED ISSUE #322: Zoom webinar unavailable
Hi,
I can't understand why the webinar option is always grayed and unavailable. 
My Zoom subscription include the Webinar&Event plan, as I perform Zoom webinar in-platform on a regular basis.
In Moodle I can create a new Zoom meeting, but it seems "I am not preauthorized" to create webinar... so does anyone know what's the trick?

Thanks,
Giacomo

jrchamp:
Please note that this will be very difficult for us to test the new functionality as we do not have a Zoom Events license. However, as long as it doesn't actively break anything, we should be in good shape. I'll reach out to the requester to see if they can test it fully.

Fixes #322

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -819,32 +819,25 @@ function zoom_get_meeting_capacity(string $zoomhostid, bool $iswebinar = false) 
     // Get the 'feature' section of the user's Zoom settings.
     $userfeatures = $service->get_user_settings($zoomhostid)->feature;
 
+    $meetingcapacity = false;
+
     // If this is a webinar.
-    if ($iswebinar == true) {
-        // Get the 'webinar_capacity' value.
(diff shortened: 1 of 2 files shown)


jrchamp:
@sgrandh3 As we don't have one of these licenses ourselves, please check that adding a new Zoom activity is still behaving as expected. It should not show Webinar as an option unless you have a Webinar license. It should not show a Large Meeting capacity unless you have a Large Meeting license. I may be able to find a Webinar and/or Large Meeting license to test that situation - just let me know when so that I'm not disrupting your testing in the middle.

rickbeasley (approved):
Code looks good, approved.

PR #342, opened 2021-12-09
Zoom meeting sync scheduled task is inefficient #321

### LINKED ISSUE #321: Zoom meeting sync scheduled task is inefficient
We noted that the Zoom meeting sync scheduled task was lasting for about 400 seconds, which is probably a product of how many Zoom activities we have.

As it runs for ~8min every 30min, this consumes a lot of time used by cron. I doubt it's creating huge load during this, but it is causing other cron jobs to back up.

As a workaround, we've reduced the syncing to daily and early in the morning. We assume users are more likely to edit the meeting in Moodle if they've created it there.

rickbeasley:
Run update meeting task once daily at 4:30.


jrchamp (approved):
We'll want to mention in the notes for the new version that users may want to reset the task to the default schedule as this change may not automatically update users who are currently using the default.

PR #345, opened 2021-12-14
Add recordings_view table data to privacy provider

### LINKED ISSUE #344: Recording view data not included in privacy provider definition
We came across this failing unit test test when running the full suite of Moodle unit tests.
```
There was 1 failure:

1) provider_testcase::test_table_coverage
The following tables with user fields must be covered with metadata providers: 
  - zoom_meeting_recordings_view (userid)

/siteroot/privacy/tests/provider_test.php:327
/siteroot/lib/phpunit/classes/advanced_testcase.php:80

To re-run:
 vendor/bin/phpunit "provider_testcase" privacy/tests/provider_test.php
```

Currently working on a PR, should have it up for review shortly.

jwalits:
PR to fix the failing unit test outlined in #344


classes/privacy/provider.php
```diff
@@ -269,5 +305,19 @@ public static function delete_data_for_users(\core_privacy\local\request\approve
         $params = array_merge($inparams, ['contextid' => $context->id]);
 
         $DB->delete_records_select('zoom_meeting_participants', "id $sql", $params);
```
jrchamp, line 307:
```suggestion
        $DB->delete_records_select('zoom_meeting_participants', "id IN ($sql)", $params);
```

classes/privacy/provider.php
```diff
@@ -147,13 +167,16 @@ public static function export_user_data(\core_privacy\local\request\approved_con
...
                   FROM {context} c
             INNER JOIN {course_modules} cm ON cm.id = c.instanceid AND c.contextlevel = :contextlevel
             INNER JOIN {modules} m ON m.id = cm.module AND m.name = :modname
             INNER JOIN {zoom} z ON z.id = cm.instance
             INNER JOIN {zoom_meeting_details} zmd ON zmd.zoomid = z.id
             INNER JOIN {zoom_meeting_participants} zmp ON zmp.detailsid = zmd.id
+            INNER JOIN {zoom_meeting_recordings} zmr ON zmr.zoomid = z.id
+            INNER JOIN {zoom_meeting_recordings_view} zmrv.recordingsid = zmr.id
                  WHERE c.id $contextsql
                        AND zmp.userid = :userid
```
jrchamp, line 181: From a data perspective, adding on to this query seems like we're mixing two different types of information. I think it's probably unclear what the data means here to people who have not read the code. I recommend splitting it into two separate datasets (though I'm also not familiar with how to do this in the Moodle privacy code).

At a minimum, if we are not splitting the data into two different datasets, then from a SQL perspective this may need to be modified to use LEFT JOINs and the same "(zmp.userid = :userid OR zmrv.userid = :userid)" logic from get_contexts_for_userid()

Specifically, it will miss situations where there are either no participant entries or no recording view entries due to the INNER JOIN. Conversely, it will double / triple up on the number of rows where there are multiple participant and recording view entries.

Something like:
```suggestion
             LEFT JOIN {zoom_meeting_details} zmd ON z.id = zmd.zoomid
             LEFT JOIN {zoom_meeting_participants} zmp ON zmd.id = zmp.detailsid
             LEFT JOIN {zoom_meeting_recordings} zmr ON z.id = zmr.zoomid
             LEFT JOIN {zoom_meeting_recordings_view} zmrv ON zmr.id = zmrv.recordingsid
                 WHERE c.id $contextsql
                       AND (zmp.userid = :userid OR zmrv.userid = :userid)
```

classes/privacy/provider.php
```diff
@@ -205,6 +229,12 @@ public static function delete_data_for_all_users_in_context(\context $context) {
                 $DB->delete_records('zoom_meeting_participants', array('detailsid' => $meetingdetail->id));
             }
             $DB->delete_records('zoom_meeting_details', array('zoomid' => $cm->instance));
+
+            $meetingrecordings = $DB->get_records('zoom_meeting_recordings', array('zoomid' => $cm->instance));
+            foreach ($meetingrecordings as $recording) {
+                $DB->delete_records('zoom_meeting_recordings_view', array('recordingsid' => $recording->id));
+            }
+            $DB->delete_records('zoom_meeting_recordings', array('zoomid' => $cm->instance));
```
jrchamp, line 237: The meeting recordings themselves are not personal data from the Moodle perspective, so I don't think we should delete them unless this method is what Moodle uses for "course reset" functionality.
```suggestion
```

classes/privacy/provider.php
```diff
@@ -205,6 +229,12 @@ public static function delete_data_for_all_users_in_context(\context $context) {
                 $DB->delete_records('zoom_meeting_participants', array('detailsid' => $meetingdetail->id));
             }
             $DB->delete_records('zoom_meeting_details', array('zoomid' => $cm->instance));
```
jrchamp, line 231: At this moment, "meeting topic" is - for some reason - marked in this provider class as "personal data". However, I don't think that's right. I would expect meeting topics to be "Week 1" or "Group 3 Project", not "Jonathan 1-on-1 Zoom meeting". Similar to my other comment, unless Moodle is using these privacy function for "course reset" functionality, which I don't think it is, then it doesn't make sense to delete metadata for activities that is not tied to users.

classes/privacy/provider.php
```diff
@@ -60,6 +60,11 @@ public static function get_metadata(\core_privacy\local\metadata\collection $col
         $coll->add_database_table('zoom_meeting_details',
                                         ['topic' => 'privacy:metadata:zoom_meeting_details:topic'],
                                         'privacy:metadata:zoom_meeting_details');
```
jrchamp, line 62: As referenced in the previous comment, I don't understand why we're marking the Zoom meeting topics are privacy metadata. It shouldn't be personal data at all, at least from my understanding. After all, this topic data should be the same as the Zoom activity name. Conversely, if this is only being included so that the data export provides additional context for a user's personal data, then a comment here that clarifies the purpose of this code would help avoid future confusion.

This is another "the original code has confused me" comment that would be awesome to fix as part of this Pull Request, but it's optional and I don't want it to slow our ability to merge the proposed fix.

jrchamp (commented):
Hi @jwalits! Thanks for working to address this issue. It looks like for the most part you have successfully followed the patterns that were used by the original code. Unfortunately, there are several places here where the original code seems wrong, so I've added some suggestions to try and address both the original code and the changes in this Pull Request.

As the original code does not appear to work correctly for delete_data_for_users(), it would be great if you can test that it works after the suggested changes.

jwalits:
Hey @jrchamp 

Thanks for the review on that MR, I will try and get some time today/tomorrow to get those issues you raised fixed up.

jwalits:
Hey @jrchamp 

Sure, you can start working on the changes. I actually haven't got around to making a start on the changes. I was hoping to find some time this week, but week kept getting away :-(

classes/privacy/provider.php
```diff
@@ -173,14 +190,55 @@ public static function export_user_data(\core_privacy\local\request\approved_con
...
+                'userid' => $recordingviewinstance->userid,
+                'viewed' => $recordingviewinstance->viewed,
+                'timemodified' => \core_privacy\local\request\transform::datetime($recordingviewinstance->timemodified),
+            ];
+
+            $contextdata = (object) array_merge((array) $contextdata, $instancedata);
+            \core_privacy\local\request\writer::with_context($context)->export_data(array(), $contextdata);
+        }
+
+        $recordingviewinstances->close();
```
jrchamp, line 241: I don't full understand the export structure for the Privacy API, so this is my attempt to mirror what was already here. The names probably need to be more unique if get_metadata() is used to label them.

jrchamp (commented):
Hi @jwalits! 

I did my best to try and apply my suggestions. This is definitely going to need to be tested by someone who can understand how Moodle Privacy works, because I'm not confident that I've done enough. A definite improvement overall though, because I'm not sure it's even working in the current version.

classes/privacy/provider.php
```diff
@@ -76,15 +81,17 @@ public static function get_contexts_for_userid(int $userid): \core_privacy\local
...
             INNER JOIN {modules} m ON m.id = cm.module AND m.name = :modname
             INNER JOIN {zoom} z ON z.id = cm.instance
-            INNER JOIN {zoom_meeting_details} zmd ON zmd.zoomid = z.id
-             LEFT JOIN {zoom_meeting_participants} zmp ON zmp.detailsid = zmd.id
-                 WHERE zmp.userid = :uclauserid
+            LEFT JOIN {zoom_meeting_details} zmd ON zmd.zoomid = z.id
+            LEFT JOIN {zoom_meeting_participants} zmp ON zmp.detailsid = zmd.id
+            LEFT JOIN {zoom_meeting_recordings} zmr ON zmr.zoomid = z.id
+            LEFT JOIN {zoom_meeting_recordings_view} zmrv ON zmrv.recordingsid = zmr.id
+                 WHERE zmp.userid = :userid OR zmrv.userid = :userid
```
tuanngocnguyen, line 88: Hi @jrchamp 

Since @jwalits  is on leave, I will follow up the PR.
This line cause DB error as it will expect 4 parameters
I have a fix here: https://github.com/tuanngocnguyen/moodle-mod_zoom/commit/1efca8d1e82874e1477853d700d48d5829ea3413
As I cannot added the commit to this repos, please feel free to cherry-pick or fix it in a new commit

tuanngocnguyen:
Hi @jrchamp 

I have done export data testing (after fixing param number) with mock up data (sql insert into zoom table).
It works as expected and show the json data as

{
    "name": "Zoom recording test",
    "intro": "",
    "userid": "2",
    "viewed": "1",
    "timemodified": "Thursday, 3 February 2022, 12:47 AM"
}

zoom

classes/privacy/provider.php
jrchamp: Thank you for suggesting a fix. May I ask, which database are you using? I was under the impression that named parameters could be reused multiple times.

tuanngocnguyen: Hi @jrchamp 

I am using postgres

It causes error when this unit test is run

```
Time: 1.31 minutes, Memory: 138.00 MB

There was 1 error:

1) provider_testcase::test_component_understands_deleted_users with data set "mod_zoom" ('mod_zoom', 'mod_zoom\privacy\provider')
dml_exception: ERROR: Incorrect number of query parameters. Expected 4, got 3.

/var/www/html/lib/dml/moodle_database.php:953
/var/www/html/lib/dml/pgsql_native_moodle_database.php:818
/var/www/html/privacy/classes/local/request/contextlist.php:71
/var/www/html/mod/zoom/classes/privacy/provider.php:98
/var/www/html/lib/moodlelib.php:8114
/var/www/html/privacy/classes/manager.php:578
/var/www/html/privacy/tests/provider_test.php:199
/var/www/html/lib/phpunit/classes/advanced_testcase.php:80
phpvfscomposer:///var/www/html/vendor/phpunit/phpunit/phpunit:60

To re-run:
 vendor/bin/phpunit "provider_testcase" privacy/tests/provider_test.php
```

tuanngocnguyen: The error is thrown at this line (master branch): https://github.com/moodle/moodle/blob/master/lib/dml/moodle_database.php#L945

jrchamp: Thank you @tuanngocnguyen! The code you highlighted makes it clear that Moodle requires unique parameter names and does not allow mixing parameter types. For some reason, I thought that they could be reused, but that must have been for something else.

jrchamp:
Thank you @tuanngocnguyen for doing some testing. I noticed that `intro` was present and I wondered where it was coming from. Of course, it is because the privacy code adds the helper's context data to the instance data. Unfortunately, I didn't pay attention to that as closely as I should have, because I used `name` for one of the instance data elements which is essentially a reserved key: https://github.com/moodle/moodle/blob/c352b700223e534781e8971631c73566be3205e6/privacy/classes/local/request/helper.php#L206-L231

At this time, the reserved keys for activity modules are:
 - name
 - intro
 - completion

I'll add a commit to rename the key.

We should probably add a proper privacy testcase so that we can benefit from continuous testing. The testcase class to extend is `\core_privacy\tests\provider_testcase`

tuanngocnguyen:
Hi @jrchamp,

The core test suite has already cover the privacy test case, so I am not sure if we need additional test for this.
Please let me know if there is any task I can do to fast track this MR.
Thanks

jrchamp:
Hi @tuanngocnguyen 

While the core test suite does seem to test some things, it does not verify that our privacy provider is working. There were at least 2 queries in the original code that would have never returned the correct information. Having a dedicated test would have caught that issue and will hopefully avoid any regressions as the code changes in the future. Thankfully, Moodle core does have many [examples](https://github.com/moodle/moodle/blob/master/mod/choice/tests/privacy/provider_test.php) for their internal modules which can act as a guide for creating a provider_testcase for mod_zoom. @smbader volunteered to implement an initial version so that each of the functions that our privacy provider class defines are tested whenever the CI process runs.

tuanngocnguyen:
Thanks @jrchamp for the info.

mattporritt:
Hi @jrchamp and @smbader ,
Is there an eta on this? If not we may be able to pick it.

Cheers

smbader:
Hi @mattporritt,

While this has been on our radar to work, we haven't been able to allocate the time to dedicate to it. I would be super appreciative if you had the opportunity to work on it.

Steve

marcghaly:
Hello @smbader, @jrchamp,
I have added unit tests covering 7 methods present in your provider in this branch in one commit https://github.com/marcghaly/moodle-mod_zoom/commits/issue344-privacy-provider, over @jwalits 's one.

Please let me know if you have any question.

Marc

jrchamp:
Thanks @marcghaly! The current failures are mostly grunt issues that I'll address in a separate PR. I'll then rebase and push the updated branch and we may then be ready.

jrchamp (approved):
This is great! Thank you all so much for your contributions.

PR #350, opened 2022-01-14
Do not store false in cache, causes error on get

### LINKED ISSUE #349: get_user_id stores false in cache
In zoom_get_user_id, it can (and does) store false in the cache. If the cachestore is set to file (the default) then an exception is thrown on retrieval.

Basically set should only be called if $zoomuserid is set. Also note that $zoomuser isn't known to exist (and likely will not) in the catch.

ericmerrill:
Patch for issue #349

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -477,17 +477,15 @@ function zoom_get_user_id($required = true) {
         $service = new mod_zoom_webservice();
         try {
             $zoomuser = $service->get_user(zoom_get_api_identifier($USER));
-            if ($zoomuser !== false) {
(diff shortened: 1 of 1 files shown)


jrchamp (approved):
Thanks @merrill-oakland! This makes a lot of sense. Even if storing false worked, it would still try and fetch the values fresh from the webservice (meaning: I do not see any downsides here).

In case you are wondering why I rebased your commit: I needed to pick up the moodle-local_codechecker fixes from the other PR so this could pass the automated checks which are required to pass before any code is merged.

merrill-oakland:
@jrchamp No worries, thank for picking this up!

PR #351, opened 2022-01-15
Compatibility with moodle-local_codechecker v3.0.5

jrchamp:
moodle-plugin-ci recently updated its moodle-local_codechecker dependency to v3.0.5. Some of the new checks are failing and it is actively preventing new code from being merged until this is fixed.


smbader (approved):
Thanks for keeping us in check. Looks good.

PR #358, opened 2022-02-23
Fix fields in install.xml to match the upgrade script #357

### LINKED ISSUE #357: Database Schema check error
We are coming across the Database schema check on some client sites

.

This occurs only when upgrading the plugin from version 4.2.1 or below to the latest. This does not happen with a new installation. The fix for it is fairly straight forward. Just need to update the install.xml to correctly match the changes we made in the upgrade.php

I'll submit a PR very soon to fix this

jwalits:
Fixes #357


jrchamp:
Hi @jwalits! While this may correct "new new installations", my concern is that the "almost new installations" will show the same error message that the "upgrade installations" are seeing right now.

Essentially:
 - Install 4.2.1 -> Upgrade to 4.3.2 (shows error message) -> apply this PR (fixed)
 - Install 4.3.2 (has incorrect database values from install.xml) ->apply this PR (shows error message)
 
 So there needs to be an entry in upgrade.php that fixes the NOTNULL and DEFAULT values in the database for people who have the wrong values because they installed versions 4.3 - 4.3.2.
 
Also, I'll fix the JWT issues that are showing up on Moodle 4 in a separate issue.

jwalits:
Hey @jrchamp

Thanks for checking! Totally valid point, I missed the upgrade script as part of the change. I have added it here and tested both scenarios. Its working for me locally. Could you please have a look when you get a chance.....thanks

rickbeasley (approved):
Looks great!

PR #359, opened 2022-02-23
webservice: Specify HS256 when encoding

jrchamp:
This is required for forward-compatibility with Moodle 4.0 which has updated the php-jwt library to version 6.0. Thankfully, the only backward-incompatible change that appears to affect us is that the formerly-optional algorithm parameter - that defaulted to our desired value - is now a required parameter. As Zoom currently only allows one option, it makes sense for us to specify that algorithm in the code.

This is a high priority fix because we test against Moodle's master branch and that automated test is failing. This PR needs to be merged before we will be able to merge any other PRs.

### DIFF
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -213,7 +213,7 @@ class mod_zoom_webservice {
             'iss' => $this->apikey,
             'exp' => time() + 40
         );
-        $token = \Firebase\JWT\JWT::encode($payload, $this->apisecret);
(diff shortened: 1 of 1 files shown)


rickbeasley (approved):
Looks great!

smbader (approved):
Big win for our PHP Unit tests.

This will also assure we are using the right encoding with Zoom

Issue #346, opened 2021-12-14
Activity states that recurring meeting is not found, but it exists and can be used


cdipe:
We run moodle 3.9.7 and recently upgraded to Zoom Meetings v4.3.
Now when entering the activity it states that the meeting cannot be found on Zoom.
But the meeting exists in Zoom and can still be started and joined with the Join link.

![image](https://user-images.githubusercontent.com/14754240/146003072-2d681950-ac2e-417e-8fe7-b108fa67110d.png)

![image](https://user-images.githubusercontent.com/14754240/146003558-15fa37ed-a49f-42e3-84b4-c7d8630e16dc.png)

![image](https://user-images.githubusercontent.com/14754240/146003283-bb4b992c-5cf0-402b-91e2-64c44d9e7977.png)

jrchamp:
Is this an issue for new meetings or is it only a problem for this meeting? I wonder if this was due to the AWS issues that were happening earlier this month. You can try setting the "exists_on_zoom" value to `1` in the database to see if that resolves the problem long term.

emmarichardson:
I have also seen this message suddenly start showing up.  We use a course template with zoom meeting included to create all new courses.  I updated the database as noted above and will report back if the message reappears.

jrchamp:
Thank you @emmarichardson for providing your feedback. I'm very interested to know if you encounter more of these.

I have noticed that some of my Zoom automation scripts have been experiencing timeouts. It's possible that these meetings are being marked as expired when a timeout occurs while contacting Zoom's API. That may mean that we need to more carefully check the response codes to differentiate between "does not exist" and "request timed out". I've checked the places where ZOOM_MEETING_EXPIRED gets set, but all of them appear to already be checking that the response from the API is one of the three "not exist" error codes.

marcellobarile:
I'm also experiencing the same issue on some courses, actually digging into it but would be nice to get some official hints

jrchamp:
Hints? Can do. First, I searched for the code that shows "meeting cannot be found on Zoom". It's coming from the language string `zoomerr_meetingnotfound`, which is used on the view.php page when `$showrecreate` is `true` which is set when `$zoom->exists_on_zoom == ZOOM_MEETING_EXPIRED`. So then I searched for places that set `exists_on_zoom` to `ZOOM_MEETING_EXPIRED`.

I found three ways for the Zoom meeting to be marked as expired.

[...]

PR #362, opened 2022-03-07
Added API call timeout

marcellobarile:
* Added a new setting to set the CURL timeout (useful for #346)
* Added a new setting to enable the "webinar" mode by default when creating a new activity


lang/en/zoom.php
```diff
@@ -44,6 +44,8 @@
 $string['apiidentifier_desc'] = 'The identifier field to use when making a call to the Zoom API';
 $string['apikey'] = 'Zoom API key';
 $string['apikey_desc'] = '';
+$string['apitimeout'] = 'Zoom API timeout';
+$string['apitimeout_desc'] = 'Amount of seconds before to consider an API call expired. Default is 60.';
```
jrchamp, line 48:
```suggestion
$string['apitimeout_desc'] = 'Number of seconds to wait before closing the API connection. Default is 60.';
```

mod_form.php
```diff
@@ -301,7 +301,7 @@ public function definition() {
                     }
                     $mform->addElement('advcheckbox', 'webinar', get_string('webinar', 'zoom'),
                             get_string('webinarthisis', 'zoom'), $webinarattr);
-                    $mform->setDefault('webinar', 0);
+                    $mform->setDefault('webinar', $config->webinardefault );
```
jrchamp, line 304: Extra space
```suggestion
                    $mform->setDefault('webinar', $config->webinardefault);
```

settings.php
```diff
@@ -68,6 +68,10 @@
             get_string('apisecret_desc', 'mod_zoom'), '');
     $settings->add($apisecret);
 
+    $apitimeout = new admin_setting_configtext('zoom/apitimeout', get_string('apitimeout', 'mod_zoom'),
+            get_string('apitimeout_desc', 'mod_zoom'), '');
```
jrchamp, line 72: Seems like this should be a number instead of text and default to 60.

jrchamp (commented):
Were you able to test and confirm that you were getting timeouts? I have some comments, but overall the code makes sense to me.

classes/webservice.php
```diff
@@ -216,6 +222,14 @@ private function make_call($path, $data = array(), $method = 'get') {
         $token = \Firebase\JWT\JWT::encode($payload, $this->apisecret, 'HS256');
         $curl->setHeader('Authorization: Bearer ' . $token);
 
+        $calltimeout = get_config('zoom', 'apitimeout');
+        if (empty($calltimeout)) {
+          $calltimeout = self::CALL_TIMEOUT;
+        }
+
+        $curl->setopt(array('CURLOPT_CONNECTTIMEOUT' => $calltimeout));
+        $curl->setopt(array('CURLOPT_TIMEOUT' => $calltimeout));
```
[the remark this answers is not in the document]
marcellobarile: fixed, silly me :)

lang/en/zoom.php
marcellobarile: updated, thanks for the hint

settings.php
marcellobarile: yup, done

marcellobarile:
> Were you able to test and confirm that you were getting timeouts? I have some comments, but overall the code makes sense to me.

Honestly, I did an empirical test; since I bumped the timeout to 60 seconds the meetings are not marked as expired anymore.
I thought that because the change was not destructive, I could proceed with it in any case.
I know it's suboptimal so I would understand if before proceeding with the merge you would ask for something objective :)

jrchamp:
I'll try and rebase this to avoid the grunt issues that were fixed in the lasted changes.

jrchamp:
I'll try and rebase this to avoid the grunt issues that were fixed in the lasted changes.

jrchamp:
Based on the testing that I've done, cURL doesn't have a low connected timeout (the amount of time after a successful connection is established before it gives up). The default seems to be `0` which means "no limit". The default connection timeout (the amount of time it waits for a connection to be established) seems to be somewhere about 250 seconds. It looks like Moodle's `curl` class that we are using is initialized with a connection timeout of 30 seconds: https://github.com/moodle/moodle/blob/ad6dc71c56280c3aad236eb7b0f6650e00eecf34/lib/filelib.php#L3233

30 seconds should be enough time to establish a connection (most of the places where it gets overridden in Moodle set the value even lower to somewhere between 2 and 10 seconds). The changes suggested here would - by default - increase the connection timeout from 30 seconds to 60 seconds, and reduce the overall timeout from unlimited to 60 seconds.

As an interesting side note: PHP's `default_socket_timeout` ini value (which does default to 60 seconds) is not used by cURL in any of my testing.

jrchamp:
I've opened #367 to split out the functionality that is ready to be merged (the webinar by default option). As the core of this PR is focused on timeouts, it seemed better to leave that functionality here.

sgrandh3:
I did regression testing and nothing seems to break, everything looks good.

jrchamp:
Thank you @marcellobarile for your work on this issue. While the current code doesn't seem to break anything, it also doesn't appear to provide direct benefit towards resolving the the original issue. At this time, let's close the issue until evidence exists.

Issue #355, opened 2022-01-31
Prep for Moodle 4.0


jrchamp:
There are some upcoming enhancements in Moodle 4.0 that we can support: https://docs.moodle.org/dev/Moodle_4.0_developer_update

There may be a way to add that support in a backward compatible way.

As far as discussion points, would you say that the Zoom activity is more "MOD_PURPOSE_COMMUNICATION" or "MOD_PURPOSE_COLLABORATION"? I feel like the _goal_ is learning and collaboration, while the _method_ is communication. Maybe it means something different to each of our users.

armccoy:
I think that while Zoom is certainly a _conduit_ for collaboration (as well as for content and assignments), at its core, it's really a communication tool, much in the same way that the Moodle chat and feedback modules are categorized as communication activities.

jrchamp:
We should make sure to add PHP 8.0 to the matrix we use for our GitHub Action tests.

jrchamp:
If we have any Grunt-related conflicts as a result of changes in Moodle 4.0, we can make sure that we're only performing grunt tests against specific branches. Example: https://github.com/moodlehq/moodle-plugin-ci/pull/158/files

jrchamp:
While we're here, we should probably make sure that the database connections are properly supporting utf8mb4 for MySQL. These lines were recently added to the upstream example https://github.com/moodlehq/moodle-plugin-ci/blob/master/gha.dist.yml
```yml
          MYSQL_CHARACTER_SET_SERVER: "utf8mb4"
          MYSQL_COLLATION_SERVER: "utf8mb4_unicode_ci"
```

jrchamp:
For the module purpose constants, they appear to be defined here:
https://github.com/moodle/moodle/blob/1d99ba19a21d57e9f1ed4211a8eeee00e50b7baf/lib/moodlelib.php#L464-L472

Thus in our `function zoom_supports()` function in lib.php, we could add something before the `switch` that is forward compatible like:
```php
if (defined('FEATURE_MOD_PURPOSE') && $feature === FEATURE_MOD_PURPOSE) {
    return MOD_PURPOSE_COMMUNICATION;
}
```

PR #364, opened 2022-03-09
Fix/grunt amd rebuild

jrchamp:
Related to #355, this has the grunt updates that have been merged into all of the upstream branches as part of MDL-73915. For now, we'll ignore Moodle 4.0's grunt complaints because upstream has decided to drop support for older browsers.

For our code, this is specifically the translation of the `await` keyword: https://github.com/ncstate-delta/moodle-mod_zoom/blob/13a1f5e3e017e5502202eabe92203df7052a1bd7/amd/src/toggle_text.js#L32-L35

At some point, it makes sense to rebuild our JS against Moodle 4.0+, but it's not even released yet. We should probably revisit this when the first LTS version of Moodle 4.x is released.


rickbeasley (approved):
Looks good.

PR #367, opened 2022-04-08
Feature/option webinar by default

jrchamp:
This includes the "webinar by default option" from #362 and some code checker fixes for code coverage that would have prevented a merge.


marcellobarile:
Yeah makes sense 👍

smbader (approved):

PR #369, opened 2022-04-13
Moodle 4.0 modifications

### LINKED ISSUE #355: Prep for Moodle 4.0

smbader:
Start of Moodle 4.0 modifications

- Update icon.svg
- Add FEATURE_MOD_PURPOSE to zoom_supports
- Remove heading string from view/index

Fixes #355

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - php: '8.1'
+          - php: '8.0'
             moodle-branch: 'master'
             database: 'mariadb'
           - php: '8.0'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,9 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          - php: '8.0'
+            moodle-branch: 'master'
+            database: 'mariadb'
           - php: '7.4'
             moodle-branch: 'master'
             database: 'pgsql'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,11 +8,11 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - php: '8.0'
+          - php: '8.1'
             moodle-branch: 'master'
             database: 'mariadb'
-          - php: '7.4'
-            moodle-branch: 'master'
+          - php: '8.0'
+            moodle-branch: 'MOODLE_400_STABLE'
             database: 'pgsql'
           - php: '7.4'
             moodle-branch: 'MOODLE_311_STABLE'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,8 @@ jobs:
         env:
           MYSQL_USER: 'root'
           MYSQL_ALLOW_EMPTY_PASSWORD: "true"
+          MYSQL_CHARACTER_SET_SERVER: "utf8mb4"
+          MYSQL_COLLATION_SERVER: "utf8mb4_unicode_ci"
         ports:
           - 3306:3306
         options: --health-cmd="mysqladmin ping" --health-interval 10s --health-timeout 5s --health-retries 3
--- a/index.php
+++ b/index.php
@@ -65,7 +65,6 @@ $PAGE->set_heading($course->fullname);
 $PAGE->set_pagelayout('incourse');
 
 echo $OUTPUT->header();
-echo $OUTPUT->heading($strname);
 
 if (! $zooms = get_all_instances_in_course('zoom', $course)) {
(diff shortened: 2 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -8,6 +8,9 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          - php: '8.0'
+            moodle-branch: 'master'
+            database: 'mariadb'
           - php: '7.4'
             moodle-branch: 'master'
```
jrchamp, line 15: I don't know if PHP 8.1 will work yet or what they plan to name the 4.0 branch, but these changes might be nice if we can. I can test it elsewhere to avoid muddying this PR.
```suggestion
          - php: '8.1'
            moodle-branch: 'master'
            database: 'mariadb'
          - php: '8.0'
            moodle-branch: 'MOODLE_400_STABLE'
```

lib.php
```diff
@@ -38,6 +38,12 @@
  * @return mixed true if the feature is supported, null if unknown
  */
 function zoom_supports($feature) {
+
```
jrchamp, line 41: Not sure why initial blank lines irk me, but they do?
```suggestion
```

pix/icon.svg
```diff
@@ -1 +1,5 @@
-<svg xmlns="http://www.w3.org/2000/svg" width="32" height="32" viewBox="0 0 24 24"><g fill-rule="evenodd"><path d="M24 12c0 6.629-5.371 12-12 12S0 18.629 0 12 5.371 0 12 0s12 5.371 12 12zm0 0" fill="#3f88ff"/><path d="M16.219 13.77v-3.54l3.238-2.37v8.28zm0 0M15.422 10.813c0-1.739-.965-2.954-2.856-2.954H4.543v5.286c0 1.87 1.148 2.996 2.871 2.996h8.008v-5.328zm0 0" fill="#fff"/></g></svg>
\ No newline at end of file
+<?xml version="1.0" encoding="utf-8"?>
+<svg width="74.4px" height="74.4px" viewBox="0 0 24 24" version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+     style="enable-background:new 0 0 74.4 74.4;" xml:space="preserve" preserveAspectRatio="xMinYMid meet">
+    <g><path fill-rule="evenodd" d="M24 12c0 6.629-5.371 12-12 12S0 18.629 0 12 5.371 0 12 0s12 5.371 12 12zm0 0 M16.219 13.77v-3.54l3.238-2.37v8.28zm0 0 M15.422 10.813c0-1.739-.965-2.954-2.856-2.954H4.543v5.286c0 1.87 1.148 2.996 2.871 2.996h8.008v-5.328zm0 0" fill="#fff"/></g>
+</svg>
```
jrchamp, line 5: I would like to pass this through SVGOMG and confirm with you that the result displays the same way.

index.php
```diff
@@ -65,7 +65,6 @@
 $PAGE->set_pagelayout('incourse');
 
 echo $OUTPUT->header();
-echo $OUTPUT->heading($strname);
```
jrchamp, line 68: Are these changes causing duplication in all version of Moodle or should this only be removed in Moodle 4.0? If it's the latter, we'll probably need a Moodle version check here to conditionally output the heading in older versions.

view.php
```diff
@@ -134,8 +134,6 @@
 // Output starts here.
 echo $OUTPUT->header();
 
-echo $OUTPUT->heading(format_string($zoom->name), 2);
```
jrchamp, line 137: Same comment for this one: Are these changes causing duplication in all version of Moodle or should this only be removed in Moodle 4.0? If it's the latter, we'll probably need a Moodle version check here to conditionally output the heading in older versions.

jrchamp (commented):
These are great! There's a couple of things that I want to test with you once Moodle 4.0 is released and then we can determine which of these suggestions/questions are valid.

.github/workflows/ci.yml
```diff
@@ -8,8 +8,11 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - php: '7.4'
+          - php: '8.1'
```
jrchamp, line 11: Based on the last test, Moodle is definitely not ready for PHP 8.1 at this moment.
```suggestion
          - php: '8.0'
```

smbader: Worth a shot

jrchamp (approved):
Looks great! Nice to maintain backwards compatibility in the same codebase without too much added complexity.

PR #373, opened 2022-04-26
Ci/phpunit fix warnings

jrchamp:
In PHPUnit 9, new aliases were created for some of the assert methods and the originals were deprecated. Unfortunately, forward compatibility was not added to PHPUnit 8 and PHPUnit 10 removes the deprecated methods. The developer has made it very clear that they are not going to change this behavior. This makes it very difficult to support multiple versions without a home-grown compatibility layer.

Hence the following:
- switch all methods to the new names
- override the new method in the test class to determine if we can truly use the new name

The benefit is that when we drop support for PHPUnit 8 (i.e. drop support for PHP 7.2 and 7.3), we will be able to remove the compatibility method and everything should just keep working.


rickbeasley (approved):
Looks good.

PR #378, opened 2022-05-05
restore: Handle exceptions so overall restore can complete

### LINKED ISSUE #377: Exception during restore process
When restoring a course, if there is a Zoom activity but the mod_zoom plugin has not been configured, an Exception will be thrown that halts the course restore process.

```
Zoom webservice error: Zoom API key not found.:
#0 mod/zoom/backup/moodle2/restore_zoom_stepslib.php(60): mod_zoom_webservice->__construct()
#1 backup/util/plan/restore_structure_step.class.php(137): restore_zoom_activity_structure_step->process_zoom()
#2 backup/util/helper/restore_structure_parser_processor.class.php(121): restore_structure_step->process()
```

jrchamp:
This should allow restores to complete even if the plugin has not yet been configured on the system.

Fixes #377

### DIFF
--- a/backup/moodle2/restore_zoom_stepslib.php
+++ b/backup/moodle2/restore_zoom_stepslib.php
@@ -57,13 +57,13 @@ class restore_zoom_activity_structure_step extends restore_activity_structure_st
         global $DB;
 
         $data = (object)$data;
-        $service = new mod_zoom_webservice();
 
         // Update start_time before attempting to create a new meeting.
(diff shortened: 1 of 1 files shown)


smbader (approved):
Inside of the __construct for mod_zoom_webservice, if the key or secret is not set it will throw an errorwebservice error. Since the behavior from the __construct function is to throw an exception when not configured, I think this is the most acceptable fix for this problem.

Issue #381, opened 2022-05-17
Security settings error


nstefanski:
Zoom's webservice to get account meeting security settings is only available for paid accounts. Rather than throw an error when no security settings are returned, could this be changed to a debugging message? This would allow the plugin to be used even with free accounts, which could be helpful for small schools and sites set up for a single classroom.

jrchamp:
This is probably a duplicate of #228 and #251

PR #382, opened 2022-05-17
change security settings error to debugging call

nstefanski:
Possible resolution for issue #381

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -498,12 +498,17 @@ function zoom_get_user_id($required = true) {
  */
 function zoom_get_meeting_security_settings() {
     $cache = cache::make('mod_zoom', 'zoommeetingsecurity');
-    if (!($zoommeetingsecurity = $cache->get('meetingsecurity'))) {
+    try {
+        $zoommeetingsecurity = $cache->get('meetingsecurity');
+    } catch (moodle_exception $error) {
+        debugging($error->getMessage());
+    }
+    if (!$zoommeetingsecurity) {
         $service = new mod_zoom_webservice();
         try {
             $zoommeetingsecurity = $service->get_account_meeting_security_settings();
         } catch (moodle_exception $error) {
-            throw $error;
+            debugging($error->getMessage());
         }
         $cache->set('meetingsecurity', $zoommeetingsecurity);
     }


locallib.php
```diff
@@ -498,12 +498,17 @@ function zoom_get_user_id($required = true) {
  */
 function zoom_get_meeting_security_settings() {
     $cache = cache::make('mod_zoom', 'zoommeetingsecurity');
-    if (!($zoommeetingsecurity = $cache->get('meetingsecurity'))) {
+    try {
+        $zoommeetingsecurity = $cache->get('meetingsecurity');
```
jrchamp, line 502: What would cause `$cache->get()` to throw an exception?

locallib.php
```diff
@@ -498,12 +498,17 @@ function zoom_get_user_id($required = true) {
...
+    }
+    if (!$zoommeetingsecurity) {
         $service = new mod_zoom_webservice();
         try {
             $zoommeetingsecurity = $service->get_account_meeting_security_settings();
         } catch (moodle_exception $error) {
-            throw $error;
+            debugging($error->getMessage());
         }
         $cache->set('meetingsecurity', $zoommeetingsecurity);
```
jrchamp, line 513: We should probably only worry about caching successful values and return a reasonable default on error. With this change, we're probably returning null or false which is not the expected stdClass type.

jrchamp (commented):
Overall, I agree that we should be gracefully handling the exception, but I think we need to make sure to return a valid default response and only cache good values from the API.

The `$service->get_account_meeting_security_settings()` function definition should probably be modified at the same time, because it has the same confusing unconditional sequence: try, catch, throw.

nstefanski:
@jrchamp luckily I have bot ha paid and unpaid account with which to test this. Here are the default values returned that are actually used by the plugin:

`stdClass Object ( 
  [meeting_password_requirement] => stdClass Object ( 
    [length] => 0 
    [have_letter] => 
    [have_number] => 
    [have_special_character] => 
    [have_upper_and_lower_characters] => 
    [only_allow_numeric] => 
    [consecutive_characters_length] => 0 )
  [end_to_end_encrypted_meetings] => `

I can generate this structure instead of just a null value; the null value _works_, but it generates several notices so admittedly not the best solution on my part.

locallib.php
nstefanski: I am not sure, but this did give me an error when it was outside the try loop. Possibly something got messed up because I set the cache without receiving a good value. I can move the set cache within the try loop and see if this works with a clean install.

jrchamp: It's probably the null value being saved with $cache->set() and then later fetched with $cache->get(), which is similar to the issue from #349

nstefanski:
Closing in favor of pull #383

PR #383, opened 2022-05-18
error handling for security settings

### LINKED ISSUE #228: Error after upgrading
Hello!
After upgrading to the latest version of Zoom. I get an error:
Zoom received a bad request: Only available for Paid account.
Moodle 3.9.
Thank you!

### LINKED ISSUE #251: Mod stop working (V3.6) : Only available for Paid account
Hello,
We have an official "K-12 School Time Limit Removal" account for Education Nationale (France). Since v3.6 update, the Zoom plugin do not works any more, the error message is " Zoom received a bad request: Only available for Paid account.
Thank you for your help.
Olivier

### LINKED ISSUE #381: Security settings error

nstefanski:
Cleaner resolution for issues #228, #251, and #381. If the webservice receives the 200 error "Only available for Paid account" when trying to retrieve security settings, the default values are used instead.

Note: the caching error noted in the last pull request is gone, so likely that was caused by accidentally saving a null value in earlier testing.

### DIFF
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -425,7 +425,7 @@ class mod_zoom_webservice {
      */
     public function get_account_meeting_security_settings() {
         $url = 'accounts/me/settings?option=meeting_security';
-        $response = null;
+        $response = (object) array('meeting_security' => (object) array());
         try {
             $response = $this->make_call($url);
             $meetingsecurity = $response->meeting_security;
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -444,7 +444,7 @@ class mod_zoom_webservice {
         }
 
         // Set a default encryption setting if it is not present.
-        if (!isset($meeting_security->end_to_end_encrypted_meetings)) {
+        if (!isset($meetingsecurity->end_to_end_encrypted_meetings)) {
             $meetingsecurity->end_to_end_encrypted_meetings = false;
         }
 
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -429,10 +429,10 @@ class mod_zoom_webservice {
             $response = $this->make_call($url);
             $meetingsecurity = $response->meeting_security;
         } catch (moodle_exception $error) {
+            // Only available for Paid account, return default settings.
             $meetingsecurity = new stdClass();
-            if ($error->zoomerrorcode == 200) {
-                // Only available for Paid account, return default settings.
-            } else {
+            // If some other error, show debug message.
+            if ($error->zoomerrorcode != 200) {
                 debugging($error->getMessage());
             }
         }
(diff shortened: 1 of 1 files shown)


classes/webservice.php
```diff
@@ -428,13 +428,25 @@ public function get_account_meeting_security_settings() {
...
+        if (!isset($response->meeting_security->meeting_password_requirement)) {
+              $response->meeting_security->meeting_password_requirement = (object) self::DEFAULT_MEETING_PASSWORD_REQUIREMENT;
+        }
+
+        // Set a default encryption setting if it is not present.
+        if (!isset($response->meeting_security->end_to_end_encrypted_meetings)) {
+            $response->meeting_security->end_to_end_encrypted_meetings = false;
+        }
+
         return $response->meeting_security;
```
jrchamp, line 450: We only use `$response->meeting_security`, so I'm tempted to change everything a little.

Inside the try:
`$meetingsecurity = $response->meeting_security;`
Inside the catch:
`$meetingsecurity = new stdClass();`

And then everywhere else that says `$response->meeting_security` would become `$meetingsecurity`.

classes/webservice.php
```diff
@@ -428,13 +428,25 @@ public function get_account_meeting_security_settings() {
...
             $response = $this->make_call($url);
-            // Set a default meeting password requirment if it is not present.
-            if (!isset($response->meeting_security->meeting_password_requirement)) {
-                $response->meeting_security->meeting_password_requirement = (object) DEFAULT_MEETING_PASSWORD_REQUIREMENT;
-            }
         } catch (moodle_exception $error) {
-            throw $error;
+            if ($error->zoomerrorcode == 200) {
+                // Only available for Paid account, return default settings.
+                $response = (object) array('meeting_security' => (object) array());
```
jrchamp, line 434: Should probably move this empty object outside of the `if` so that the empty object is already initialized when we try to use it.

jrchamp (commented):
This makes a lot of sense and I had a couple ideas that may help.

classes/webservice.php
nstefanski: added commit

jrchamp (commented):
Looking great! A couple small adjustments and I will happily advocate for approval in tomorrow's meeting.

classes/webservice.php
nstefanski: added commit, replaced `$response = null;` on ln 428 with `$response = (object) array('meeting_security' => (object) array());`, if that's not what you meant let me know

nstefanski: saw your clarification below, adding new commit

smbader (approved):
Thank you Nicholas for the most excellent contribution!

PR #386, opened 2022-05-26
tracking fields: gracefully catch exception on update

### LINKED ISSUE #385: "Zoom API key not found" error while installing the plugin
Hi,

Moodle version: 3.11.6
Plugin version: v4.5.2 (2022051900)

Description of the issue
---------------------------
While installing the plugin in Moodle, at the end of the process, the new settings that the plugin inserted into Moodle are shown.
It seems that if some settings are left empty (at least "Zoom API key" [zoom | apikey]), and then the admin tries to end the install process by clicking on the "Save changes" button at the bottom of the page, the following error page is displayed:

^^^^^^^^^^^^
Zoom webservice error: Zoom API key not found.

×Debug info:
Error code: errorwebservice

×Stack trace:
line 124 of /mod/zoom/classes/webservice.php: moodle_exception thrown
line 1124 of /mod/zoom/locallib.php: call to mod_zoom_webservice->__construct()
line 962 of /mod/zoom/lib.php: call to zoom_list_tracking_fields()
line 2051 of /lib/adminlib.php: call to mod_zoom_update_tracking_fields()
line 8968 of /lib/adminlib.php: call to admin_setting->post_write_settings()
line 26 of /admin/upgradesettings.php: call to admin_write_settings()
^^^^^^^^^^^^

Expected Results
-------------------
An admin could install the plugin without having to fill in required details (like the API key), already during the install process.
It is reasonable to assume that an admin would want to install the plugin at a certain point in time but would like to config it later.

It would be great if you could fix this behavior of the plugin.

Thanks

jrchamp:
An exception can occur when the Zoom plugin is not yet configured and the plugin is being installed. This catches the exception that occurs inside of a settings update callback so that it won't break other Moodle processes. When the Zoom plugin is configured, the scheduled task will automatically refresh the correct values, so this temporary delay in accuracy should be overall better.

Fixes #385

### DIFF
--- a/classes/task/update_tracking_fields.php
+++ b/classes/task/update_tracking_fields.php
@@ -63,7 +63,9 @@ class update_tracking_fields extends \core\task\scheduled_task {
         // Show trace message.
         mtrace('Starting to process existing Zoom tracking fields ...');
 
-        mod_zoom_update_tracking_fields();
+        if (!mod_zoom_update_tracking_fields()) {
+            mtrace('Error: Failed to update tracking fields.');
+        }
 
         // Show trace message.
         mtrace('Finished processing existing Zoom tracking fields');
--- a/lib.php
+++ b/lib.php
@@ -958,35 +958,42 @@ function mod_zoom_get_fontawesome_icon_map() {
 function mod_zoom_update_tracking_fields() {
     global $DB;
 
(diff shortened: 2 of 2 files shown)


smbader (approved):
Thanks for adding the check to the task as well!

PR #388, opened 2022-06-03
release: v4.5.3

jrchamp:
A small bugfix release so that someone who triggers the settings update callback for the tracking fields but has not yet configured the Zoom API credentials will still be able to process the rest of their changes.


rickbeasley (approved):
Looks good. BTW, looks like Steve reviewed 386 so I just checked the changelog and verson.php.

PR #390, opened 2022-06-23
Issue281 - New activity setting: Automatic Recording

### LINKED ISSUE #281: Add the possibility to record a meeting at creation time
Enhancement : When creating a new meeting, have the checkbox to select the option to activate the recording of the meeting.
It corresponds to the API POST/Create a meeting/Settings object/auto_recording which could have three values (none, cloud, local).
This is not related to the ability to read or get a meeting's record within Moodle.

aduranterres:
Resolves #281


jrchamp (commented):
I know you're still working on this, so I only did a brief check to see if I could understand the code.

jrchamp:
@aduranterres I rebased the code and rebuilt the JavaScript files with `grunt amd`. Please make sure I didn't break anything, because the initial rebase lost the JavaScript files...

aduranterres:
Hi @jrchamp Thank you very much for all the code cleaning you did, sorry about the unchanged headers, etc.

I have uploaded some new changes:
- Don't use form reload if allowrecordingchangeoption is not enabled in Zoom settings, as it is not needed
- Don't do anything on definition_after_data if allowrecordingchangeoption is not enabled, as the field option_auto_recording does not exist (it was throwing an error if you disable allowrecordingchangeoption)

The function definition_after_data is always executed after the form first loads, so that's why the second modification was needed.

I see now that grunt is complaining again...I am not sure why  as I have not changed them after I got your changes or run grunt again, so sorry for giving you more work.

jrchamp:
I was able to get a few minutes to apply your changes and I'm fairly sure that I have them all this time. 😄 

@sgrandh3 Thank you for offering to test this. You will probably need one or two people to grant you scheduling privilege, so let me know and I'll grant you access.

mod_form.php
```diff
@@ -50,6 +50,7 @@ public function definition() {
 
         $service = new mod_zoom_webservice();
         $zoomuser = $service->get_user($zoomapiidentifier);
+        $this->zoomuser = $zoomuser;
```
jrchamp, line 53: As `$this->zoomuser` is not a pre-defined part of the class, this code is adding it as a dynamic property. [Dynamic properties are deprecated in PHP 8.2](https://wiki.php.net/rfc/deprecate_dynamic_properties) and that functionality will removed in a future version.

We either need to define the property as being part of the class or to load the data again. If the purpose was to avoid performance problems, then we should seek to address the performance problems at the source in a separate issue.

mod_form.php
```diff
@@ -521,6 +522,32 @@ public function definition() {
         $mform->setDefault('option_mute_upon_entry', $config->defaultmuteuponentryoption);
         $mform->addHelpButton('option_mute_upon_entry', 'option_mute_upon_entry', 'mod_zoom');
 
+        // Add autorecording option if enabled.
+        $allowrecordingchangeoption = $config->allowrecordingchangeoption;
+        $this->allowrecordingchangeoption = $allowrecordingchangeoption;
```
jrchamp, line 527: Same comment about dynamic properties for `$this->allowrecordingchangeoption`

mod_form.php
```diff
@@ -531,6 +558,7 @@ public function definition() {
         $showschedulingprivilege = ($config->showschedulingprivilege != ZOOM_SCHEDULINGPRIVILEGE_DISABLE) &&
                 count($scheduleusers) > 1 && $allowschedule; // Check if the size is greater than 1 because
                                                              // we add the editing/creating user by default.
+        $this->showschedulingprivilege = $showschedulingprivilege;
```
jrchamp, line 561: Same comment about dynamic properties for `$this->showschedulingprivilege`

jrchamp (commented):
Adding some comments about dynamic class properties and adjusting some admin-level setting defaults.

mod_form.php
aduranterres: Thanks for the info, I didn't know about that change in PHP 8.2.
The purpose of this and the other added class properties is to avoid duplicating calls to get_user and get_config functions.
Maybe for get_config is not worth it, and we can call it again inside definition_after_data().
But get_user makes a web service call, so I think it is better not to do it twice for better performance.

jrchamp: We should probably add a helper function that caches the results of get_user() so that it's only ever called once per page load. That's what I was suggesting could be addressed in a separate issue.

aduranterres: Thanks, I created issue #401 to address it.

jrchamp (approved):
The testing revealed a minor issues string issues that were easily addressed. Approving and merging. We'll plan to release this next week, hopefully with a performance improvement PR #401.

PR #392, opened 2022-07-24
Fix JS error: Computer audio only = no button

mwithheld:
In a course > zoom instance
with Audio options=Computer audio only
view the instance (e.g. /mod/zoom/view.php?id=9)
A JS error appears in the console:

```
toggle_text.js:29 Uncaught TypeError: Cannot read properties of null (reading 'addEventListener')
    at _exports.init (toggle_text.js:29:10)
```
This is bc the element "#show-more-button" does not exist in this particular case.

The fix is to modify mod/sazoom/amd/src/toggle_text.js to check the button exists first.


jrchamp:
Thanks @mwithheld! This makes a lot of sense to me. I'll probably need to add a commit that does a grunt rebuild, but this fix should hopefully be able to released soon.

jrchamp:
@sgrandh3 Please take a look to see if you can verify it is working.

sgrandh3:
Verified for new and existing zoom meetings, even for recurring meetings. Unable to reproduce the JS error. Looks good .

jrchamp (approved):
Thanks @mwithheld ! Simple, clean fix for the optional button.

PR #399, opened 2022-08-18
breakout rooms: only send breakout_room setting when data exists

jrchamp:
During restore of a previous backup of a Zoom activity, a warning message would occur:

> PHP Warning:  Undefined property: stdClass::$breakoutrooms in mod/zoom/classes/webservice.php on line 623


rickbeasley (approved):
Looks good.

PR #402, opened 2022-09-02
Enhancement/minimize api get user requests

### LINKED ISSUE #401: Cache the results of get_user()
Create a helper function that caches the results of get_user() so that it's only ever called once per page load.

jrchamp:
There are a lot of get_user() and get_user_settings() calls, which are often fetching the same information multiple times for a given page. As a first step, add some helper functions that avoid duplicating API calls for a single HTTP request. In the future, it would be easy to add a Moodle cache store so multiple HTTP requests can benefit from a single API call result, but cache invalidation is not a challenge for today.

Fixes #401

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,7 +264,7 @@ v2.1
 v2.0.1
 
 - Fixing conflicts with Firebase\JWT library. If more conflicts are found,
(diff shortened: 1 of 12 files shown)


smbader (approved):
Excellent work!

PR #405, opened 2022-09-22
release: v4.7.0

jrchamp:
A new feature (automatic recording) highlighted how some API calls were being duplicated. Rather than trying to solve the problem in isolation, we now provide a basic set of helper functions so the plugin can ask for the needed data it needs while minimizing API calls and overhead.


rickbeasley (approved):
Looks good!

PR #407, opened 2022-09-26
mod_form: Also call moodleform_mod parent methods

### LINKED ISSUE #406: Locked settings are no longer applied in 4.7.0
Starting with version 4.7.0, when the `requirepasscode` setting is locked by an administrator, it is not properly enforced in the Zoom activity creation form. The lock icon does not appear near the label and the user is allowed to uncheck the box.

The new `definition_after_data()` method that was added in `mod_form.php` should call `parent::definition_after_data()`. This is where the locking of form elements was performed in previous versions.

jrchamp:
`moodleform` is extended by `moodleform_mod` to provide baseline functionality for both forms and activity modules. However, when we override the methods, we need to remember to call the parent class method so that we still benefit from the baseline functionality before adding our own.

This regression was unintentionally a result of #390

Fixes #406

### DIFF
--- a/mod_form.php
+++ b/mod_form.php
@@ -670,6 +670,8 @@ class mod_zoom_mod_form extends moodleform_mod {
     public function definition_after_data() {
         global $USER;
 
+        parent::definition_after_data();
+
         // Get config.
         $config = get_config('zoom');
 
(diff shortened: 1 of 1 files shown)


rickbeasley (approved):
Looks good!

PR #408, opened 2022-09-26
api: Correctly use user-level meeting security settings

### LINKED ISSUE #404: Use User-level meeting security configuration instead of Account-level
Current, passcode and end-to-end encryption settings are loaded from an account-level function:
https://github.com/ncstate-delta/moodle-mod_zoom/blob/f2deab3064fda764e2738467c62772474bc4c970/classes/webservice.php#L422-L428

This caused some issues/regressions for us in the past, which were resolved by setting some defaults via #383

Instead, we should likely be using the user-level functions to get these same values. I say that for two reasons:

1. This should mean that the static default values are used less often, resulting in more accurate results.
2. The values returned by the current method only reflect the account-level settings, not the user-level settings. I have confirmed that by testing the differences between my individual user account and our overall account.

This seems like a win-win, but I think we should still keep the static defaults as a fallback, just in case.

### LINKED ISSUE #423: Missing language string
mod_zoom: main branch latest
Moodle: MOODLE_311_STABLE latest

core_cache\administration_helper_test::test_get_summaries
Unexpected debugging() call detected.
Debugging: String does not exist. Please check your string definition for cachedef_oauth/mod_zoom
* line 10499 of /lib/moodlelib.php: call to debugging()
* line 623 of /cache/classes/definition.php: call to lang_string->__construct()
* line 205 of /cache/classes/administration_helper.php: call to cache_definition->get_name()
* line 100 of /cache/tests/administration_helper_test.php: call to core_cache\administration_helper::get_definition_summaries()
* line 1545 of /vendor/phpunit/phpunit/src/Framework/TestCase.php: call to core_cache\administration_helper_test->test_get_summaries()
* line 1151 of /vendor/phpunit/phpunit/src/Framework/TestCase.php: call to PHPUnit\Framework\TestCase->runTest()
* line 80 of /lib/phpunit/classes/advanced_testcase.php: call to PHPUnit\Framework\TestCase->runBare()
* line 726 of /vendor/phpunit/phpunit/src/Framework/TestResult.php: call to advanced_testcase->runBare()
* line 903 of /vendor/phpunit/phpunit/src/Framework/TestCase.php: call to PHPUnit\Framework\TestResult->run()
* line 672 of /vendor/phpunit/phpunit/src/Framework/TestSuite.php: call to PHPUnit\Framework\TestCase->run()
* line 672 of /vendor/phpunit/phpunit/src/Framework/TestSuite.php: call to PHPUnit\Framework\TestSuite->run()
* line 672 of /vendor/phpunit/phpunit/src/Framework/TestSuite.php: call to PHPUnit\Framework\TestSuite->run()
* line 673 of /vendor/phpunit/phpunit/src/TextUI/TestRunner.php: call to PHPUnit\Framework\TestSuite->run()
* line 143 of /vendor/phpunit/phpunit/src/TextUI/Command.php: call to PHPUnit\TextUI\TestRunner->run()
* line 96 of /vendor/phpunit/phpunit/src/TextUI/Command.php: call to PHPUnit\TextUI\Command->run()
* line 97 of phpvfscomposer:///vendor/phpunit/phpunit/phpunit: call to PHPUnit\TextUI\Command::main()
* line 118 of /vendor/bin/phpunit: call to include()

/var/www/html/lib/phpunit/classes/advanced_testcase.php:88
phpvfscomposer:///var/www/html/vendor/phpunit/phpunit/phpunit:97

jrchamp:
Get accurate information for the user themself instead of the default site-level settings. A nice side benefit is that this should reduce the list of authorized scopes so that the `account:read:admin` scope is not needed for full feature support.

Fixes #404 
Fixes #423


jrchamp:
TODO: Remove the `account:read:admin` scope documentation as it stops being used as a result of this PR.

jrchamp:
Okay, the documentation is updated to reflect these changes. And my testing showed that it did correctly reflect the settings at my individual user level. The two pieces that probably should be checked:

- [ ] Does the meeting edit page correctly show the host's settings (instead of the just the current user), for example with end-to-end encryption.
- [ ] If a user's settings are changed (either at the individual or account-wide levels), how long does it take for the cache to get the new values? Does it ever happen?

sgrandh3:
Worked with jonathan and tested the 2 above stated issues.

jrchamp:
The cached data stays around for a while, likely until the caches are automatically purged during a Moodle update (core or plugin). This is still an improvement over the existing cache which completely ignores the user's individual settings. One caveat: when a new meeting is being scheduled for another user, it will not show the other user's security options until after the initial save.

rickbeasley (approved):
Looks good!

PR #410, opened 2022-09-29
release: v4.8.0

jrchamp:
Two completely different, but very valuable examples of community involvement:
- Server-to-Server OAuth is a required transition, but is a great opportunity for us to clarify which scopes we need for the plugin to function (and each organization can customize that list based on the functionality they enable).
- A regression bug report which we would not have caught on our own because the new functionality was our focus. Plus, it led to the identification of other areas where class inheritance were inconsistent.

Thank you everyone who uses the plugin and either lets us know when something isn't working or helps us make the plugin better. We all benefit from this collaboration. ❤️

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,14 @@ server is properly synchronized with the time servers.
 
 ## Changelog
 
+v4.8.0
+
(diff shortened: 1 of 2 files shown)


rickbeasley (approved):
Looks good!

PR #417, opened 2022-10-18
No longer display activity introduction in v4.0 and later

### LINKED ISSUE #414: Moodle description displayed twice at top of activity
When clicking on the zoom activity, you see the description repeated.
This is on Moodle 4.0.3 with all plugins up-to-date. This seems to be true on all of our created zoom activities since the upgrade to Moodle 4.
Zoom_error

smbader:
Moodle 4.0 moved the activity module's title, description, and activity completion into the standard module API (https://tracker.moodle.org/browse/MDL-72413). We identified the need to no longer display the activity title but did not identify the activity introduction (my fault). 

This pull request employees the same backwards compatible check used for the title.

Fixes #414

### DIFF
--- a/view.php
+++ b/view.php
@@ -148,7 +148,7 @@ if ($showrecreate) {
 }
 
 // Show intro.
-if ($zoom->intro) {
+if ($zoom->intro && $CFG->branch < '400') {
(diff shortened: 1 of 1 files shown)


jrchamp (approved):
This change makes sense to me. If we decide in the future that we need to do something more advanced, here's the Moodle 4 developer update documentation for the new activity header class: https://moodledev.io/docs/devupdate#the-activity-header-class

PR #418, opened 2022-10-19
webservice: Use HTTP/1.1 to avoid HTTP/2 OAuth endpoint issue

### LINKED ISSUE #415: trouble switching to OAuth, "Undefined property: moodle_exception::$zoomerrorcode" errors
After looking at the readme, I've tried to switch to server-to-server OAuth on a test server but been unsuccessful. Here are the steps I took:

- created a s2s app in Zoom App Marketplace
- gave it _all_ scopes listed in the readme (I started with just a subset of the ones we need but then added the rest when I began running into problems; the full list did not fix the issue)
- activated it
- pasted the account id, client id, and client secret from the **App credentials** page in Zoom into the plugin's settings

When I save, I always see an error that the HTTP/2 response wasn't closed cleanly:
Screen Shot 2022-10-06 at 10 05 28 AM

We have debugging on and this error is also displayed and in our server logs:

`PHP Notice:  Undefined property: moodle_exception::$zoomerrorcode in /bitnami/moodle/mod/zoom/locallib.php on line 535, referer: MOODLE_URL/my/`

I saw that line 535 is in the `zoom_is_user_not_found_error` function. I don't know why user lookups would fail, the app works fine with JWT authentication. I tried changing the apiidentifier setting from email to username but that didn't fix the issue. Am I doing something wrong? Did I miss something in setting up the OAuth app? I think I probably set something up wrong but there also appears to be another problem in that the error code is not defined.

jrchamp:
As only a single request is made to this endpoint per hour, the simplicity of HTTP/1.1 avoids whatever underlying HTTP/2 bug was occurring in the original report. At some point in the future, HTTP/1.1 may be deprecated and the override should be able to be safely removed then.

Fixes #415

### DIFF
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -983,6 +983,11 @@ class mod_zoom_webservice {
         $curl->setHeader('Authorization: Basic ' . base64_encode($this->clientid . ':' . $this->clientsecret));
         $curl->setHeader('Content-Type: application/json');
 
+        // Force HTTP/1.1 to avoid HTTP/2 "stream not closed" issue.
+        $curl->setopt(array(
+            'CURLOPT_HTTP_VERSION' => CURL_HTTP_VERSION_1_1,
+        ));
+
         $timecalled = time();
         $response = $this->make_curl_call($curl, 'post',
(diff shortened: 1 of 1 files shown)


smbader (approved):
Forcing the http version that we expect to use is a good thing.

PR #421, opened 2022-10-25
task: move all includes to consistent location

### LINKED ISSUE #420: Cron tasks broken in latest version of the code
In the latest release of the Zoom plugin, the scheduled task 'update_meetings' fails with the following error:

Server Time: Thu, 20 Oct 2022 12:18:21 +0100

Execute scheduled task: Update meeting settings from Zoom (mod_zoom\task\update_meetings)
... started 12:18:21. Current memory use 4.8MB.
!!! Exception - Call to undefined function mod_zoom\task\zoom_webservice() !!!

It looks like the scheduled task is trying to use the zoom_webservice() function before the require_once() line that loads the function:

Screenshot 2022-10-21 at 11 16 56

jrchamp:
The includes all should happen together and for some of the tests to work, the includes need to happen before the public methods are called. There may be a better place that doesn't require side effects, but this at least makes all the tasks consistent and should resolve the reported issue.

Fixes #420

### DIFF
--- a/classes/task/get_meeting_recordings.php
+++ b/classes/task/get_meeting_recordings.php
@@ -51,7 +51,6 @@ class get_meeting_recordings extends \core\task\scheduled_task {
     public function execute() {
         global $DB;
 
-        $config = get_config('zoom');
         try {
             $service = zoom_webservice();
         } catch (\moodle_exception $exception) {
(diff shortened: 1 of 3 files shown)


rickbeasley (commented):
Looks good!

smbader (approved):
Good cleanup work 🧹

PR #422, opened 2022-10-26
Performance: Reduce API calls, respect host's settings

### LINKED ISSUE #430: Resource not exists while editing meeting with not owner
When I try to edit a meeting from moodle with an user which not owned the zoom meeting, plugins throw error
```
The resource does not exists
More information about this error
×Debug info:
Error code: errorwebservice_notfound
×Stack trace:
line 276 of /mod/zoom/classes/webservice.php: zoom_not_found_exception thrown
line 456 of /mod/zoom/classes/webservice.php: call to mod_zoom_webservice->make_call()
line 1286 of /mod/zoom/locallib.php: call to mod_zoom_webservice->get_user_settings()
line 536 of /mod/zoom/mod_form.php: call to zoom_get_user_settings()
line 214 of /lib/formslib.php: call to mod_zoom_mod_form->definition()
line 130 of /course/moodleform_mod.php: call to moodleform->__construct()
line 150 of /course/modedit.php: call to moodleform_mod->__construct()
```

this behavior not happens at 4.3.2 but in 4.8.1 and 4.9.0 testing and it occurs even with the admin role

is that the expected behavior?

jrchamp:
I noticed that we had several zoom_get_user() calls for the current user which were only ever using the Zoom user ID. So then I started checking all of the API calls on the view/edit pages to reduce API calls and improve data accuracy.

Fixes #430

### DIFF
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -610,11 +610,11 @@ class mod_zoom_webservice {
             if ($recordingoption === ZOOM_AUTORECORDING_USERDEFAULT) {
                 if (isset($zoom->schedule_for)) {
                     $zoomuser = zoom_get_user($zoom->schedule_for);
+                    $zoomuserid = $zoomuser->id;
                 } else {
-                    $zoomapiidentifier = zoom_get_api_identifier($USER);
-                    $zoomuser = zoom_get_user($zoomapiidentifier);
+                    $zoomuserid = zoom_get_user_id();
                 }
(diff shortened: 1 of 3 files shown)


jrchamp:
We were able to confirm that this correctly reinstates the ability for users without a Zoom account who have edit access to Zoom activities to be able to edit existing activities. This regression caused by #390 which was introduced in v4.7.0.

armccoy (approved):

PR #426, opened 2022-11-04
Fix/webservice error handling

### LINKED ISSUE #424: Bug: Passcode field rules do not reflect passcode policy from Zoom server
In an earlier version there was a bug that the plugin did not use the passcode policy from zoom server and therefore produced invalid passcodes. This issue was fixed #160 

However, it seems that in version 4.8.1 (2022102000) this bug (or a related issue) is back again: if I try to plan a zoom meeting, the generated default passcode does not match our requirements (in my case 8 characters necessary vs default 6 characters).

If I change the pascode manually to 8 characters, it works. I can save the meeting.
 If I do not change the passcode and try to save, I get the following error message:
 
> Zoom received a bad request: {$a}
> 
> [More information about this error](http://docs.moodle.org/311/en/error/zoom/errorwebservice_badrequest)
> 
> Debug info:
> Error code: errorwebservice_badrequest×Dismiss this notification
> Stack trace:
> line 274 of /mod/zoom/classes/webservice.php: zoom_bad_request_exception thrown
> line 717 of /mod/zoom/classes/webservice.php: call to mod_zoom_webservice->make_call()
> line 102 of /mod/zoom/lib.php: call to mod_zoom_webservice->create_meeting()
> line 128 of /course/modlib.php: call to zoom_add_instance()
> line 168 of /course/modedit.php: call to add_moduleinfo()
> ×Dismiss this notification
> Output buffer: <br /> <b>Notice</b>: Trying to get property 'message' of non-object in <b>/moodle/application/htdocs/moodle/eval/mod/zoom/classes/webservice.php</b> on line <b>274</b><br /> <br /> <b>Notice</b>: Trying to get property 'code' of non-object in <b>/moodle/application/htdocs/moodle/eval/mod/zoom/classes/webservice.php</b> on line <b>274</b><br />

This does not seem to be related the the new oAuth method (and related rights/scopes of the app)  At least I get the same errors if I switch back to jwt.

jrchamp:
The passcode rules issue should already be fixed by #404. However, the error responses were coming back as XML instead of JSON in some cases. This change will always request JSON instead of accepting the default response format. We also add the specific error messages, if they exist, so the user has enough information to hopefully address the cause of the issue.

Fixes #424

### DIFF
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -259,18 +259,18 @@ class mod_zoom_webservice {
             $curl->setHeader('Content-Type: application/json');
             $data = is_array($data) ? json_encode($data) : $data;
         }
-        $response = $this->make_curl_call($curl, $method, $url, $data);
+        $rawresponse = $this->make_curl_call($curl, $method, $url, $data);
 
         if ($curl->get_errno()) {
             throw new moodle_exception('errorwebservice', 'mod_zoom', '', $curl->error);
         }
 
-        $response = json_decode($response);
+        $response = json_decode($rawresponse);
 
         $httpstatus = $curl->get_info()['http_code'];
 
         if ($httpstatus >= 400) {
-            switch($httpstatus) {
+            switch ($httpstatus) {
                 case 400:
                     $errorstring = '';
                     if (!empty($response->errors)) {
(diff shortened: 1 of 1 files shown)


rickbeasley (approved):
Looks good!

PR #429, opened 2022-11-09
Setting: 'licensesnumber' will not work with multiple Moodle instances

KepaUrzelai:
Multiple Moodle instances can use the same Zoom license pool when redefining licenses, so only users from the same instance should be considered when checking if the number of licenses limit has been reached.

Also, when recycling licenses, only the ones that users of the same instance are using should be taken.

### DIFF
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -412,13 +412,16 @@ class mod_zoom_webservice {
      * Incrementally retrieves the active paid users and compares against $numlicenses.
      * @see $numlicenses
      * @return bool Whether the paid user license limit has been reached.
+     * @throws dml_exception
      */
     private function paid_user_limit_reached() {
         $userslist = $this->list_users();
         $numusers = 0;
         foreach ($userslist as $user) {
-            if ($user->type != ZOOM_USER_TYPE_BASIC && ++$numusers >= $this->numlicenses) {
-                return true;
+            if (core_user::get_user_by_email($user->email)) {
+                if ($user->type != ZOOM_USER_TYPE_BASIC && ++$numusers >= $this->numlicenses) {
+                    return true;
+                }
             }
         }
         return false;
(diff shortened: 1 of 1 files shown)


classes/webservice.php
```diff
@@ -412,13 +412,16 @@ public function list_users() {
...
+     * @throws dml_exception
      */
     private function paid_user_limit_reached() {
         $userslist = $this->list_users();
         $numusers = 0;
         foreach ($userslist as $user) {
-            if ($user->type != ZOOM_USER_TYPE_BASIC && ++$numusers >= $this->numlicenses) {
-                return true;
+            if (core_user::get_user_by_email($user->email)) {
+                if ($user->type != ZOOM_USER_TYPE_BASIC && ++$numusers >= $this->numlicenses) {
```
jrchamp, line 422: Only use a database lookup for users who might be relevant. Move the `$numusers` increment operation outside of the conditional.
```suggestion
            if ($user->type != ZOOM_USER_TYPE_BASIC && core_user::get_user_by_email($user->email)) {
                $numusers++;
                if ($numusers >= $this->numlicenses) {
```

classes/webservice.php
```diff
@@ -412,13 +412,16 @@ public function list_users() {
      * Incrementally retrieves the active paid users and compares against $numlicenses.
      * @see $numlicenses
      * @return bool Whether the paid user license limit has been reached.
+     * @throws dml_exception
```
jrchamp, line 415: We're not throwing a dml_exception ourselves, so this seems unnecessary.
```suggestion
```

classes/webservice.php
```diff
@@ -428,12 +431,13 @@ private function paid_user_limit_reached() {
      * Gets the ID of the user, of all the paid users, with the oldest last login time.
      *
      * @return string|false If user is found, returns the User ID. Otherwise, returns false.
+     * @throws dml_exception
```
jrchamp, line 434:
```suggestion
```

jrchamp (commented):
Hi @KepaUrzelai 

Thank you for suggesting this change. I can imagine that this will help administrators with multiple separate groups of users with separate Moodle instances. However, I do think that it may not work as efficiently for other setups. It may be worthwhile to add an admin-level setting that allows an administrator to choose whether or not to include the `core_user::get_user_by_email()` check/filter when counting users. To avoid a backward-compatibility issue, the default should not include the check. Perhaps "Only count licensed users on this Moodle instance"?

KepaUrzelai:
Hello again!

Considering @jrchamp proposal, I made a new Pull request with the changes (#436).

With those, an admin would be able to choose if the **Redefine licenses** functionality needs to check only for actual Moodle instance Zoom users rather than checking for every Zoom API user.

Thank you both.
Kepa

PR #436, opened 2022-12-12
Enhanced PR #429

KepaUrzelai:
Following @jrchamp proposal in #429 , I added the 'Check instance users' setting so 'licensesnumber' setting also works when multiple Moodle instances are pointing to the same Zoom licenses pool.

Read pr #429 for the hole discussion.


settings.php
```diff
@@ -115,6 +115,11 @@
             get_string('lowlicenses', 'mod_zoom'), 0, 1);
     $settings->add($utmost);
 
+    $instanceusers = new admin_setting_configcheckbox('zoom/instanceusers',
+                    get_string('instanceusers','mod_zoom'),
+                    get_string('instanceusers_desc','mod_zoom'),0);
```
jrchamp, line 120:
```suggestion
                    get_string('instanceusers', 'mod_zoom'),
                    get_string('instanceusers_desc', 'mod_zoom'), 0);
```

jrchamp (commented):
I had a few suggestions and I'll go ahead and run the automated tests to see if they find anything else. Overall, the code makes sense visually, so I believe this should be a straightforward review once we're back in the office.

settings.php
```diff
@@ -115,6 +115,11 @@
             get_string('lowlicenses', 'mod_zoom'), 0, 1);
     $settings->add($utmost);
 
+    $instanceusers = new admin_setting_configcheckbox('zoom/instanceusers',
+                    get_string('instanceusers', 'mod_zoom'),
+                    get_string('instanceusers_desc', 'mod_zoom'), 0);
```
jrchamp, line 120: These may be the default values, but I think it may be important for us to be clear that we want this off by default.
```suggestion
                    get_string('instanceusers_desc', 'mod_zoom'), 0, 1, 0);
```

jrchamp (commented):
I feel pretty good about this. It should be ready for testing (mainly to make sure everything is still working with it off, that the setting defaults to off and that all the strings are working as expected).

settings.php
jrchamp, line 120: These may be the default values, but I think it may be important for us to be clear that we want this off by default. Also, fixes the extra indentation.

```suggestion
            get_string('instanceusers', 'mod_zoom'),
            get_string('instanceusers_desc', 'mod_zoom'), 0, 1, 0);
```

armccoy (approved):
Looks good to me.

PR #439, opened 2022-12-17
delete_meeting_recordings: Process deletes one meetinguuid at a time

### LINKED ISSUE #437: Delete meeting recordings doesn't  ends
With a site with almost 200k meeting recordings, we cannot finish the task delete_meeting_recordings.

I think  is that for every zoom recording, fetch all recordings for the unique meetinguuid, and de-populate the recording array as check API $service->get_recording_url_list($meetinguuid) get associated recordings.

At the end of this process (that could take easy more than one day in my case), only then starts the deletion process of the remaining zoomrecordings. Of course, database has gone away at this time (extending timeouts is a solution too, but zoom recording management could be better)

There is any chance that the deletion takes action for every meeting and not at the end?
I suggest fetch all local meetingsuuid, iterate for each to compare local recordings for that meetinguuid against the api query to zoom and then process the deletion for the diff.

jrchamp:
Previously the code would loop over all recordings collecting API responses for all related meeting UUIDs. That takes a long time and the process might expire before any cleanup was actually processed. Now, it should fetch one meeting UUID's list of recordings, process any deletes and move on to the next.

Fixes #437

### DIFF
--- a/classes/task/delete_meeting_recordings.php
+++ b/classes/task/delete_meeting_recordings.php
@@ -67,36 +67,23 @@ class delete_meeting_recordings extends \core\task\scheduled_task {
 
         mtrace('Checking if any meeting recordings in Moodle have been removed from Zoom...');
 
-        // Get all recordings stored in Moodle.
-        $zoomrecordings = zoom_get_meeting_recordings();
-        // Fetch all recordings for the unique meetinguuid.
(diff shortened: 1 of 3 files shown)


smbader:
TODO: Test on server with a lot of data/recordings.

smbader (approved):

PR #446, opened 2023-01-11
Grunt rebuild and GitHub Action updates for Moodle 4.1 LTS

### LINKED ISSUE #434: Rebuild JavaScript against Moodle 4.1 (LTS)
As mentioned in #364, when the first version of Moodle 4.1 (LTS) is released, we should rebuild our JS against Moodle 4.1.

Will also need to modify the GitHub Action step for grunt to only worry about Moodle 4.x

jrchamp:
- Add a specific Moodle 4.1 target
- Use the [latest version of GitHub's checkout action](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/)
- Ignore grunt messages from Moodle 3.x (now that we are building against Moodle 4.1)
- Any remaining changes to the workflow are from moodle-plugin-ci's suggested GitHub Action example

Fixes #434


smbader (approved):
👀 on 4.1

PR #447, opened 2023-01-11
moodle-cs: use short array syntax

jrchamp:
In [MDLSITE-4776](https://tracker.moodle.org/browse/MDLSITE-4776), Moodle has decided to use the short array syntax exclusively. Although they won't be making sweeping changes to the Moodle core code, the moodle-cs ruleset that we test against will start reporting problems soon (via moodlehq/moodle-cs#34) that will prevent us from merging code. Thus, this PR proactively attempts to resolve those issues.


smbader (dismissed):
An array of acceptable changes.

smbader (approved):

PR #412, opened 2022-10-03
Add a registration feature to the Zoom plugin

paulandm:
Add a registration feature to the Zoom plugin

Users are allowed to register far in advance of the scheduled meeting/webinar starting date.
Joining the actual meeting still respects the 'First able to join' setting.

Fixes #95
Fixes #233

PR #448, opened 2023-01-12
registration: switch to select (advcheckbox ignores 0; MDL-76095)

jrchamp:
Followup to #412 

I know this was working at some point, because we tested it, but it is definitely not working now using the advcheckbox (even though it does work in the admin settings). Switching to a select field avoids any problems with using zero as the desired value. Also, remove the workaround that was designed to act as a placeholder until MDL-76095 was fixed in Moodle core.


smbader (approved):
Clean resolution and removes a work around.

PR #451, opened 2023-01-17
patch events for My Overview

### LINKED ISSUE #450: Finished Zooms show up in My Overview block
Not sure if this affects other versions, but noticed in 4.1 that Zoom meetings are still shown in the My Overview block after they're no longer available. I'll share a patch momentarily.

nstefanski:
This addresses issue #450 by returning null data if the Zoom meeting is finished, otherwise sharing the event info with the My Overview block. This should be backwards compatible, tested with Moodle 4.1 and 3.10.

Fixes #450

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -676,12 +676,16 @@ function mod_zoom_core_calendar_provide_event_action(calendar_event $event,
     $zoom = $DB->get_record('zoom', ['id' => $cm->instance], '*');
     list($inprogress, $available, $finished) = zoom_get_state($zoom);
 
-    return $factory->create_instance(
-        get_string('join_meeting', 'zoom'),
-        new \moodle_url('/mod/zoom/view.php', ['id' => $cm->id]),
-        1,
-        $available
-    );
+    if ($finished) {
+        return null; // No point to showing finished meetings in overview.
+    } else {
+        return $factory->create_instance(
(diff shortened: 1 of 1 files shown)


jrchamp:
From https://docs.moodle.org/dev/Block_myoverview

> As each item is completed it should be removed from the list.

So this change makes sense to me. Theoretically, we should probably also be checking that the user is allowed to access this activity, but this current change is a definite improvement.

nstefanski:
@jrchamp good point, I added in a capability check as well

jrchamp:
Testing notes:
I believe that the original change [f0ca5c0](https://github.com/ncstate-delta/moodle-mod_zoom/pull/451/commits/f0ca5c0e1d244c0ebf8adbc97067edfc0868510e) may have been sufficient to resolving the reported issue. To test, remove the capability check and verify if a restricted Zoom activity is visible to a student without view access. If so, then the capability check is redundant and get_fast_modinfo() is already performing the necessary checks.

I will then be able to resolve the merge conflict by cherry-picking the original commit onto the end of the main branch.

jrchamp (approved):
Thanks all! Delighted to know that the "get fast modinfo" function is checking for appropriate access (such as capability and group access). Rebased the commit to only have the delightfully straightforward change.

PR #453, opened 2023-01-24
Registration validation based on Zoom account permissions

paulandm:
Registration validation based on Zoom account permissions

### DIFF
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -1037,4 +1037,66 @@ class mod_zoom_webservice {
         $response = $this->make_call($url);
         return $response;
     }
+
+    /**
+     * Get the roles for the Zoom account.
+     */
+    private function get_roles() {
+        $url = '/roles';
+        $response = null;
+        try {
+            $response = $this->make_call($url);
+        } catch (moodle_exception $error) {
+            throw $error;
+        }
+        return $response;
+    }
+
+    /**
+     * Get the role members for the supplied role id.
+     * @param int $id The Zoom Role ID.
+     * @return stdClass The Zoom object containing the role member information.
+     */
+    private function get_role_members($role_id) {
+        $url = '/roles/' . $role_id . '/members';
+        $response = null;
+        try {
+            $response = $this->make_call($url);
+        } catch (moodle_exception $error) {
+            throw $error;
+        }
+        return $response;
+    }
+
+    /**
+     * Returns whether or not the current user is capable of creating a meeting/webinar that requires registration.
+     * @return boolean
+     */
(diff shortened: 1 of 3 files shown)


jrchamp:
Hi @paulandm 

Thank you for opening this pull request. It looks like this code is checking if the current user has the Owner role in Zoom. That may not be the most universal way to check if registration is possible in a user's Zoom meetings. Is this restriction implementation specific to your organization's needs? Or is there a more universal way that this could be implemented?

paulandm:
Hi @jrchamp , thanks for taking a look at this.
It seems like Zoom users who do not have the 'Owner' role don't have the capability of creating meetings/webinars with the registration option (on the Zoom web page itself, not from Moodle).
We had a look through the Zoom permissions, but couldn't find one specifically related to the creating meetings/webinars with a registration option, so we ended up using the 'Owner' role as a workaround. 
So my feeling is this should actually be a global validation, but since there's no specific permission (unless I'm mistaken) available, it makes it a little tricky.

jrchamp:
Hi @paulandm 

I've been researching how registration works and I agree that there does not appear to be specific information in the API or the Zoom settings for who can and cannot register. The best resource that I have found is https://support.zoom.us/hc/en-us/articles/211579443-Scheduling-and-customizing-a-meeting-with-registration

> Prerequisites for scheduling meetings with registration
> 
> - Licensed host
> - The meeting ID must be [generated automatically](https://support.zoom.us/hc/en-us/articles/201362373) instead of using your [PMI](https://support.zoom.us/hc/en-us/articles/203276937)
> - The meeting cannot be a [No Fixed Time recurring meeting](https://support.zoom.us/hc/en-us/articles/214973206)

So the user's `type` must be `2` (Licensed), the meeting's `use_pmi` must be `false`, and the meeting's `type` cannot be `3` (A recurring meeting with no fixed time).

jrchamp:
That makes sense to me. So instead of the current code, we need to make sure that the host is Licensed if we are planning to show the registration option. And for safety, we should set `use_pmi` to `false` in the `database_to_api()` function when registration is requested:
https://github.com/ncstate-delta/moodle-mod_zoom/blob/6673a2d456c10018603a10a35dc4e9afe4054289/classes/webservice.php#L630-L632
```php
            if ($zoom->registration != ZOOM_REGISTRATION_OFF) {
                $data['settings']['use_pmi'] = false;
            }
```

paulandm:
This PR can be closed.

The above-mentioned changes have been implemented in a new PR:
https://github.com/ncstate-delta/moodle-mod_zoom/pull/622

PR #457, opened 2023-02-10
moodle-cs: One space before assignment operator (MDLSITE-6594)

jrchamp:
Moodle will be soon be enforcing the assignment operator spacing that was approved in MDLSITE-6594. This preemptively resolves the issues that exist in our main branch.

Reference: moodlehq/moodle-cs#43


smbader (approved):
All whitespace changes. Looks good.

PR #479, opened 2023-05-11
fix default activity completion editing issue

### LINKED ISSUE #478: Bug: "cannot require a CSS file before <head> has been printed" when editing default activity completion
If I go to a course > course completion > default activity completion > select Zoom and **Edit** its activity completion defaults, I receive this error: `Cannot require a CSS file after &lt;head&gt; has been printed.`

Screenshot 2023-05-10 at 1 46 13 PM

When you click the error's **Continue** button, it takes you back to the default activity completion settings page, so functionally this bug makes it impossible to edit Zoom's defaults.

Moodle: 4.0.5 (Build: 20221114)
Zoom: v4.10.1 2023021400 (I encountered on v4.10.0 also before I realized you had a new release)

karenliulll:
Related to issue #478

The error is a result of line 51 in mod_form.php. The system will automatically include the styles.css file in the plugin, deleting line 51 will keep the CSS styles but resolve the error.

### DIFF
--- a/mod_form.php
+++ b/mod_form.php
@@ -48,7 +48,6 @@ class mod_zoom_mod_form extends moodleform_mod {
     public function definition() {
         global $PAGE, $USER, $OUTPUT;
         $config = get_config('zoom');
-        $PAGE->requires->css(new moodle_url('/mod/zoom/styles.css'));
         $PAGE->requires->js_call_amd("mod_zoom/form", 'init');
 
         $isnew = empty($this->_cm);


jrchamp (approved):
This is a really nice fix. To avoid a regression in older versions of Moodle, we needed to identify which version added this "automatically include a plugin's styles.css" functionality. Going through the source, we confirmed that [MDL-20204](https://github.com/moodle/moodle/commit/78946b9bdb1299a21f00b88d81c73ad8700c951e) added the functionality to Moodle as part of the 2.0 release, which is older than our minimum supported version. Confirmed that the styles were actually being included twice prior to this fix.

PR #482, opened 2023-05-18
errors: add exception parent for cleaner error handling

### LINKED ISSUE #468: PHP 8.0 warning
Warning: Undefined property: moodle_exception:: $zoomerrorcode in mod/zoom/locallib.php on line 537

jrchamp:
The problem was that we were catching `moodle_exception` and assuming that it had the Zoom-specific webservice exception member variables. By creating a layer in between the specialized Zoom exceptions and the generic `moodle_exception`, we can guarantee that the functions will have exceptions which include the expected member variables. This also de-duplicates some of the exception code that was repeated in each of the specialized exceptions.

This also has the benefit of fixing a PHP 8.2 issue related to dynamic member variables on classes (because most of the specialized classes did not pre-define the zoomerrrorcode variable).

Fixes #468

### DIFF
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -297,8 +297,10 @@ class mod_zoom_webservice {
                         }
                     }
                     throw new zoom_bad_request_exception($response->message . $errorstring, $response->code);
+
                 case 404:
                     throw new zoom_not_found_exception($response->message, $response->code);
+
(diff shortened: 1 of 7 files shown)


armccoy (approved):
Looks good to me.

PR #483, opened 2023-05-18
Only reset zoom user data when check box is checked

### LINKED ISSUE #480: Course reset always remove zoom user data.
On the function zoom_reset_userdata doens't check the value of reset_zoom_all. When a user reset a course without marking the checkbox for reset_zoom_all, the call to zoom_reset_userdata remove everything anyway.
Zoom_reset_data function needs somethings like 
if (!empty($data->reset_zoom_all)) {
to check this value to delete user data or not.

smbader:
Fixes #480 

Added a simple check to confirm the user wants zoom user data reset before doing the work.

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -843,28 +843,30 @@ function zoom_reset_userdata($data) {
     $componentstr = get_string('modulenameplural', 'zoom');
     $status = [];
 
-    // Reset tables that record user data.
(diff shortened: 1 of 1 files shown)


jrchamp (approved):
This makes complete sense! This plugin adds a checkbox named `reset_zoom_all` to the reset form, so we should make sure the checkbox was checked before we proceed to delete all of the user data.

PR #484, opened 2023-05-18
mobile: remove legacy add-on; require Moodle 3.5+

### LINKED ISSUE #456: Drop support for legacy Mobile Remote Add-On
This would remove support for Moodle 3.4, because the modern Mobile plugin method support was added in Moodle 3.5

- [ ] Remove mobile/mod_zoom.zip
- [ ] Increase minimum supported Moodle version to 3.5 (2018051700)

jrchamp:
Prior to Moodle 3.5, Mobile support was provided by these "remote add-ons". In effect, this drops support for Moodle 3.4, so we increase the minimum required Moodle version.

Fixes #456


smbader (dismissed):
Out with the old...

PR #485, opened 2023-05-18
jwt: remove support for deprecated token method

### LINKED ISSUE #459: Remove JWT Support (target June 2023)
Remove the library and all related settings, code, strings, documentation, etc.

jrchamp:
Zoom is deprecating the JWT token auth method as of June 2023, so this removes all of the related JWT library code and the JWT credential settings `apikey` and `apisecret`

Fixes #459


aspark21:
This isn’t happening until September so you might want to wait before removing from the plugin

aspark21:
June is no new Apps but September breaks existing Moodle integrations https://developers.zoom.us/docs/internal-apps/jwt-app-migration/

jrchamp:
Thanks @aspark21. New users of the plugin will likely need to use Server-to-Server OAuth apps from June 1 (because they will not be able to create a JWT app). Although September 1 is the updated JWT end-of-life, I do prefer to remove the JWT functionality before the fall semester begins in August/September. Administrators who update the plugin between the v5.0 release and September 1 will have an additional chance to address the situation before they are hit hard in September. I think it would be fine if we wait until July 1 to drop JWT support, but we need to do it soon if we want to have another chance to help users avoid large-scale disruption.

PR #492, opened 2023-06-09
Fix for install.xml

### LINKED ISSUE #432: Inconsistency between the DB and the schema defined in install.xml
There is a small inconsistency between what is defined in install.xml and what is in the database that originated in update `2022031600`.

What happens is that the fields `show_schedule`, `show_security` and `show_media` in the table `zoom` in `install.xml` file are defined as fields that do not allow null as value. However, when these fields were created in `upgrade.php` in the upgrade block `2022031600` the field was left null in the method to define it as non-null. What causes that if the plugin was updated that field remains as not null false and if it is installed as not null true.

I have not found any problems for the user due to this, however it may cause problems in the future. Running `admin/cli/check_database_schema.php` on an updated db shows the difference.

farhan6318:


jrchamp:
What does this fix?

farhan6318:
Hi @jrchamp 

This is mostly just a syntax fix  in the xml file

smbader:
Hey all,

In #505, I fixed some db inconsistencies and also exported the schema using Moodle's db schema tool.  This action should include everything in this pull request as well.

Thank you for reporting!
Steve

PR #494, opened 2023-06-20
OAuth did not use the proxy settings.

adnbes:


jrchamp (approved):
Thanks @adnbes! This looks great and should resolve proxy issues for OAuth users.

PR #499, opened 2023-07-21
form: add required Promise.then() parameter

jrchamp:
eslint's promise/valid-params check was failing. This adds missing null anonymous callback function.


smbader (approved):
🎆

PR #503, opened 2023-07-31
settings: get error response from webservice_exception

### LINKED ISSUE #502: Exception caused on settings page
We are recently seeing this error on the settings page caused by: https://github.com/ncstate-delta/moodle-mod_zoom/blob/main/settings.php#L51

In most cases this line is fine as $errormessage is a string, but in some cases `$error->a` is not a string.

Like in our case where API gets rate limited and an error is thrown after retries, $a is not a string but an object : https://github.com/ncstate-delta/moodle-mod_zoom/blob/main/classes/webservice.php#L271 - https://github.com/ncstate-delta/moodle-mod_zoom/blob/main/locallib.php#L146-L156

This causes the whole settings page to be broken. This ideally should be handled gracefully.

We are running this on Moodle: 3.9.22+, with PHP: 7.4 and MySQL database version 8.0

jrchamp:
Fixes #502

### DIFF
--- a/settings.php
+++ b/settings.php
@@ -36,14 +36,16 @@ if ($ADMIN->fulltree) {
 
     // Test whether connection works and display result to user.
     if (!CLI_SCRIPT && $PAGE->url == $CFG->wwwroot . '/' . $CFG->admin . '/settings.php?section=modsettingzoom') {
-        $status = 'connectionok';
-        $notifyclass = 'notifysuccess';
+        $status = 'connectionfailed';
+        $notifyclass = 'notifyproblem';
         $errormessage = '';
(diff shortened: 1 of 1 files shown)


jrchamp:
CI will fail until #499 is merged and this is rebased.

jwalits:
Tested the patch and we are now seeing `Connection status: Connection failed: You have reached the maximum per-second rate limit for this API. Try again later.` message instead of the PHP error.

Looks to be working fine :+1: 

Thanks for the quick response on this @jrchamp

smbader (approved):
Good "catch"!

PR #470, opened 2023-03-16
Set default for option_auto_record for older restores

smbader:
When restoring a course containing a zoom activity that existed prior to the auto_recording changes ( #390 ), the option_auto_recording property is not set and runs into a database NULL not allowed error.

The fix in restore_zoom_stepslib.php resolves the issue but I thought it may be safe to add the check within populate_zoom_from_response as well.  This second addition may not be needed.

Fixes #455

PR #505, opened 2023-08-03
Database upgrade.php and schema consistency

### LINKED ISSUE #416: Auto recording turned off on all activities
Not sure if the setting changed on the Zoom side or on the Mod side but suddenly all my meetings that were set to auto record stopped recording - I discovered that the setting which was originally must have been pulling from zoom now was showing as None instead of Null...

### LINKED ISSUE #432: Inconsistency between the DB and the schema defined in install.xml

smbader:
Working to Fix #432 and Fix #416

- Changes made to upgrade.php to assure show_schedule, show_security, and show_media are all set to NOTNULL, matching the schema.
- Changes to the schema to remove the default and NOTNULL from option_auto_recording
- Changed the site-level default setting for auto-recording to USERDEFAULT
- Handle USERDEFAULT in mod_form

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -818,7 +818,7 @@ function xmldb_zoom_upgrade($oldversion) {
         $table = new xmldb_table('zoom');
 
         // Define and conditionally add field option_auto_recording.
-        $field = new xmldb_field('option_auto_recording', XMLDB_TYPE_CHAR, '5', null, XMLDB_NOTNULL, null, 'none', 'show_media');
+        $field = new xmldb_field('option_auto_recording', XMLDB_TYPE_CHAR, '5', null, null, null, null, 'show_media');
         if (!$dbman->field_exists($table, $field)) {
             $dbman->add_field($table, $field);
         }
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -735,19 +735,19 @@ function xmldb_zoom_upgrade($oldversion) {
         $table = new xmldb_table('zoom');
 
         // Define and conditionally add field show_schedule.
-        $field = new xmldb_field('show_schedule', XMLDB_TYPE_INTEGER, '1', null, null, null, '1', 'recordings_visible_default');
+        $field = new xmldb_field('show_schedule', XMLDB_TYPE_INTEGER, '1', null, XMLDB_NOTNULL, null, '1', 'recordings_visible_default');
         if (!$dbman->field_exists($table, $field)) {
             $dbman->add_field($table, $field);
         }
 
         // Define and conditionally add field show_security.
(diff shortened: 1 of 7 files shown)


mod_form.php
```diff
@@ -589,8 +589,14 @@ public function definition() {
                 $options[ZOOM_AUTORECORDING_CLOUD] = get_string('autorecording_cloud', 'mod_zoom');
             }
 
+            if ($config->recordingoption === ZOOM_AUTORECORDING_USERDEFAULT) {
+                $defaultsetting = $recordingsettings;
```
jrchamp, line 593:
```suggestion
                $defaultsetting = $recordingsettings->auto_recording;
```

jrchamp (commented):
This looks great overall. I had one change suggestion that I think is correct. After that, if you're feeling confident about the changes, I would be good approving. Maybe @sgrandh3 can go through the upgrade on a test server and verify that the new USERDEFAULT default recording option is working correctly.

mod_form.php
smbader: Ah, great catch!

smbader:
Miss one meeting @sgrandh3 and we assign everything to you!

I mean... can you test this as well?

jrchamp:
Should we also revert #470?

db/upgrade.php
```diff
@@ -817,7 +818,7 @@ function xmldb_zoom_upgrade($oldversion) {
         $table = new xmldb_table('zoom');
 
         // Define and conditionally add field option_auto_recording.
-        $field = new xmldb_field('option_auto_recording', XMLDB_TYPE_CHAR, '5', null, null, null, null, 'show_media');
+        $field = new xmldb_field('option_auto_recording', XMLDB_TYPE_CHAR, '5', null, XMLDB_NOTNULL, null, 'none', 'show_media');
```
jrchamp, line 821: While the current change would fix the inconsistency, any users who have not yet upgraded should be given the new-correct field definition so that they don't end up with rows that have the wrong defaults.
```suggestion
        $field = new xmldb_field('option_auto_recording', XMLDB_TYPE_CHAR, '5', null, null, null, null, 'show_media');
```

jrchamp (approved):
Looks great! Will be included in integration testing for 5.1.0

PR #507, opened 2023-08-03
pre-registration: skip grading/completion

### LINKED ISSUE #490: Grade/Completion happens when registering for the Zoom session
Hello,
I tested this beforehand, and thought it was working properly to issue a completion when the student joins the live Zoom session.  But now I have tested it to issue the completion simply when the student pre-registers for the Zoom session by clicking the "Register" button.  
Is this intended?  For us it is a problem.  I had seen somewhere else in the issue  queue comments that a grade or completion were issued when a student "joins" the session.
Any clarification or help on this is much appreciated.

jrchamp:
Fixes #490

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -1001,6 +1001,7 @@ function zoom_load_meeting($id, $context, $usestarturl = true) {
     list($inprogress, $available, $finished) = zoom_get_state($zoom);
 
     $userisregistered = false;
+    $userisregistering = false;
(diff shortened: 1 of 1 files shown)


smbader (approved):
If the Zoom activity is not available and the user is registering, they can skip the button click event, activity completion, and grading.  

Looks like a good fix.

smbader:
@sgrandh3 would you mind testing?

PR #512, opened 2023-08-24
mod_form: match title length validation to zoom's limits

### LINKED ISSUE #511: modifiying a existing meeting with a title over 200 car shows an unsignificant error
Hi,

The meeting/patch api entry says that "topic" field is limited to 200 car. So when you try to modify an existing meeting changing its title so that this limit is overshooted (french could be long text !), it shows an error message hard to understand : "Impossible de modifier Zoom" ("impossible to modify zoom").

Whereas the actual response from the api contains more useful info :
"response": {
"code": 300,
"message": "Validation Failed.",
"errors": [
{
"field": "topic",
"message": "Invalid field."
}
]
}

It would be nice to show up the "Validation Failed" response-->message content as an extra text to the standard generic "not able to" which would allow the user or the support team to find the origin of the problem faster.

smbader:
Zoom limits the topic length to 200 characters.  Our form's validation is set to 300.

Fixes #511

### DIFF
--- a/mod_form.php
+++ b/mod_form.php
@@ -168,7 +168,7 @@ class mod_zoom_mod_form extends moodleform_mod {
         $mform->addElement('text', 'name', get_string('title', 'zoom'), ['size' => '64']);
         $mform->setType('name', PARAM_TEXT);
         $mform->addRule('name', null, 'required', null, 'client');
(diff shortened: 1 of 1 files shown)


jrchamp (approved):
Great fix! This will help people right away!

PR #517, opened 2023-09-13
get meeting recordings: use complete recording data

jrchamp:
Previously, we were requesting the recordings using the meeting numeric ID, which only returns the recordings for the most recent occurrence of the meeting. Instead, we request the host's recording list, which does not have that limitation. Then we match it up to the Moodle activity and add it to the database.

Fixes #515

PR #521, opened 2023-09-21
PHPCS - use moodle-extra standard

jrchamp:
More strict standard should allow us to more easily review pull requests.

In draft until #517 is merged. I'll rebase at that point.


smbader (approved):
Consistency is key.

PR #522, opened 2023-09-22
tests: define class properties (PHP 8.2)

jrchamp:
When running the tests on a PHP 8.2 system with `display_errors` on, deprecated messages showed up.


smbader (approved):
Remember kids, define your class properties.

PR #530, opened 2023-10-09
external: fix class references

### LINKED ISSUE #527: invalid calls to mod_zoom_external
c522487652eb4087b12d07a9b5070a9edfeeb819 renamed mod_zoom_external to mod_zoom\external

but... events.php still refers to mod_zoom_external, and so does classes/output/mobile.php

jrchamp:
Fixes #527

### DIFF
--- a/classes/output/mobile.php
+++ b/classes/output/mobile.php
@@ -25,7 +25,7 @@
 namespace mod_zoom\output;
 
 use context_module;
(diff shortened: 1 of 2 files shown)


smbader (approved):
Nice find from the larger test suite.
Thank you @jrchamp for the fix.

PR #531, opened 2023-10-09
ci: check Moodle 4.3, PHP 8.2, strict phpdoc

jrchamp:
- Add Moodle 4.3
- Use PHP 8.2 for 4.3 and newer
- Only run moodle-plugin-ci steps when the moodle-plugin-ci installs successfully
- On moodle-plugin-ci v4, set `--max-warnings 0` for phpdoc
- If the job has been cancelled, mark the job as failed so that PR status checks cannot be bypassed
- Use `!cancelled()` instead of `always()` as recommended by GitHub

From https://docs.github.com/en/actions/learn-github-actions/expressions#always

> If you want to run a job or step regardless of its success or failure, use the recommended alternative: `if: ${{ !cancelled() }}`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,12 +30,15 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - php: '8.1'
+          - php: '8.2'
             moodle-branch: 'master'
             database: 'mariadb'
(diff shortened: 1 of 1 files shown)


smbader (approved):
Onward and upward!

PR #537, opened 2023-10-26
Fix: restore try catch namespace

jrchamp:
Fixes a regression in v5.1.1 where an exception during restore was not caught as a result of namespace issues. While we're here, make the namespaced files leverage `use` statements to simplify lengthy class references.


smbader (approved):
Let's use use!

PR #542, opened 2023-11-09
webservice: always initialize the OAuth scopes

### LINKED ISSUE #541: Since v5.1.1, error with PHP 8.0 : "in_array(): Argument #2 ($haystack) must be of type array, null given"
Hi,

I use Zoom plugin on several Moodle platforms, since several years, with no problem.

Since i upgraded from 5.1.0 to 5.1.1, i have errors in cron execution of task mod_zoom\task\get_meeting_reports :

Scheduled task failed: Obtenir un rapport des réunions de Zoom. (mod_zoom\task\get_meeting_reports),in_array(): Argument #2 ($haystack) must be of type array, null given

I have this problem with Moodle 4.1 on RedHat 8, with PHP 8.0.30 and MariaDB 10.6.15. No problem with Moodle 4.0 on RockyLinux, with PHP 7.3.33.

I think it's something related to some code introduced by 5.1.1 (compared to 5.1.0), not fully compatible with PHP 8.0.

Séverin

jrchamp:
The `$this->scopes` array is correctly initialized when `$this->oauth()` completes successfully. However, `has_scopes()` avoids calling that function directly so that the cached token can be reused. Unfortunately, I forgot to make sure that `$this->scopes` was being populated from the cache in those situations where the token was fresh and thus able to be reused. 😓 

Fixes #541

### DIFF
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -1084,6 +1084,8 @@ class webservice {
         $expires = $cache->get('expires');
         if (empty($token) || empty($expires) || time() >= $expires) {
             $token = $this->oauth($cache);
+        } else {
+            $this->scopes = $cache->get('scopes');
         }
 
         return $token;


smbader (approved):
Nice, clean fix. If we don't need to get a new token, then we need to load our scopes from the cache.

PR #544, opened 2023-11-10
Fix: change recurrence type default option to "No Fixed Time"

### LINKED ISSUE #523: Allow Recurring meetings to have state of 'No fixed date'
There is a setting in the plugin (zoom | defaultrecurring) which allows new meetings to default to be recurring. This used to give meetings that had no fixed time i.e. they didn't recur at specific times each day, week, etc. It appears that a change at the Zoom end means that that new meetings with a default of 'recurring' are now created to recur 'Daily'. Can an additional option be added to allow the default recurrence interval to also be set?

Zoom Recurrence

karenliulll:
Fixes #523 

When the Recurring Meeting checkbox is checked, the Recurrence dropdown below should have "No Fixed Time" as the default option. This simple fix should do the trick.

### DIFF
--- a/mod_form.php
+++ b/mod_form.php
@@ -204,12 +204,16 @@ class mod_zoom_mod_form extends moodleform_mod {
 
         // Add options for recurring meeting.
         $recurrencetype = [
-            ZOOM_RECURRINGTYPE_NOTIME => get_string('recurrence_option_no_time', 'zoom'),
             ZOOM_RECURRINGTYPE_DAILY => get_string('recurrence_option_daily', 'zoom'),
             ZOOM_RECURRINGTYPE_WEEKLY => get_string('recurrence_option_weekly', 'zoom'),
             ZOOM_RECURRINGTYPE_MONTHLY => get_string('recurrence_option_monthly', 'zoom'),
+            ZOOM_RECURRINGTYPE_NOTIME => get_string('recurrence_option_no_time', 'zoom'),
         ];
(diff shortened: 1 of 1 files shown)


mod_form.php
```diff
@@ -204,10 +204,10 @@ public function definition() {
 
         // Add options for recurring meeting.
         $recurrencetype = [
+            ZOOM_RECURRINGTYPE_NOTIME => get_string('recurrence_option_no_time', 'zoom'),
             ZOOM_RECURRINGTYPE_DAILY => get_string('recurrence_option_daily', 'zoom'),
             ZOOM_RECURRINGTYPE_WEEKLY => get_string('recurrence_option_weekly', 'zoom'),
             ZOOM_RECURRINGTYPE_MONTHLY => get_string('recurrence_option_monthly', 'zoom'),
-            ZOOM_RECURRINGTYPE_NOTIME => get_string('recurrence_option_no_time', 'zoom'),
```
jrchamp, line 210: This is the opposite of the change in https://github.com/ncstate-delta/moodle-mod_zoom/commit/df76a7267784cbbf5a9209af0e0ccdffbb9789c8. We had made that change to make sure that the Registration checkbox was available by default. However, if the site is configured to use recurring meetings by default, then `$config->defaultrecurring` could be used to set the default value. It may be best to set the default via `$mform->setDefault('recurrence_type', ZOOM_RECURRINGTYPE_NOTIME);` when `$config->defaultrecurring` is "true-ish"

jrchamp (approved):
This will nicely return to the behavior where recurring defaults to No Fixed Time when configure that way at the site level.

PR #545, opened 2023-11-16
Allow editing description for existing meetings

### LINKED ISSUE #297: Ability to edit a zoom event that is in the past
It would be good to be able to edit the description of a zoom meeting that is in the past? On one of my projects, we record the meetings, and then share this back with the users - the easiest way for me to do this, is to simply add a link to the recording back into the description of the item, so it keeps everything in one place - however it won't allow me to save the item, as it recognises that the meeting is scheduled in the past.

armccoy:
Skips "new activity" check to allow updating meeting descriptions for existing and past meetings. 
If necessary, a further edit could skip the "new activity" check if the meeting already exists AND if the meeting date/time hasn't changed.

### DIFF
--- a/mod_form.php
+++ b/mod_form.php
@@ -1011,7 +1011,7 @@ class mod_zoom_mod_form extends moodleform_mod {
         // Only check for scheduled meetings.
         if (empty($data['recurring'])) {
             // Make sure start date is in the future.
-            if ($data['start_time'] < time()) {
+            if ($data['start_time'] < time() && $data['meeting_id'] < 0) {
                 $errors['start_time'] = get_string('err_start_time_past', 'zoom');
             }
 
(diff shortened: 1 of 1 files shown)


jrchamp:
Things to test:

- Edit only the description and make sure it successfully displays in Moodle
  - This must be successful for this PR to have benefit.
- Edit the description and the start time, setting the start time to a different time that is in the past
  - What is the error that occurs?
  - Does the description save in Moodle?
  - Does the description save in Zoom? If not, do we care?

sgrandh3:
Tested the changes for also recurring meetings along with past meetings. The instructor can change the description for existing meetings

jrchamp (approved):
Thanks @sgrandh3 for testing!

PR #553, opened 2023-12-13
Fix #552: hack to prevent moodle_exception when called by defaultcompletion()

### LINKED ISSUE #552: Zoom module breaks Default activity completion with Moodle 4.3+
When mod_zoom is installed into a Moodle 4.3+ and the user has no account on Zoom the 'Default completion activity' page breaks with an error message.

I believe this originates in the change to the defaultcompletion() function in bulk_activity_completion_renderer.php (https://github.com/moodle/moodle/blob/fe8d759b1759c2dfbdc9cf27c8224c4e6746bf05/course/classes/output/bulk_activity_completion_renderer.php#L77) as this now calls core_completion_defaultedit_form() for all installed modules.

To replicate the issue do this:

- Install a vanilla MOODLE_403_STABLE
- Install mod_zoom from main branch.
- Create a new course as site admin. Make sure that Completion tracking is enabled.
- Then go to Course Completion in that course.
- Chose 'Default activity completion' from the drop down menu. You will see an error message:

```
Unable to find your account on Zoom. If you are using Zoom for the first time, you must activate your Zoom account by logging into <a href="" target="_blank" rel="noreferrer noopener"></a>. Once you've activated your Zoom account, reload this page and continue setting up your meeting. Else make sure your email on Zoom matches your email on this system.
```
Clicking the 'Continue' button will only reload the page. Also JavaScript is broken.

opitz:
This will - somehow - fix #552.
Since Moodle 4.3 defaultcompletion() will call all definition() instances from mod_form.php of installed modules.
This has implications that have been addressed here.
There might be better ways to deal with it...

### DIFF
--- a/mod_form.php
+++ b/mod_form.php
@@ -60,8 +60,14 @@ class mod_zoom_mod_form extends moodleform_mod {
 
         $zoomuserid = zoom_get_user_id(false);
 
+        // Dirty hack to prevent throwing a moodle_exception when calling this from defaultcompletion()
+        // since Moodle 4.3.
+        // When being called from defaultcompletion() the output of the page has already begun (state == 2) due to
+        // course/defaultcompletion.php line 97, so only throw a zoom_fatal_error when before any output (state == 0)
+        // as otherwise it will break the 'Default activity completion' page.
+        // TODO: find a better solution for this.
         // If creating a new instance, but the Zoom user does not exist.
-        if ($isnew && $zoomuserid === false) {
+        if ($isnew && $zoomuserid === false && $PAGE->state == 0) {
             // Assume user is using Zoom for the first time.
             $errstring = 'zoomerr_usernotfound';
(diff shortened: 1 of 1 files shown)


jrchamp:
Hi @opitz,

It looks like Moodle renamed a page and broke out previous hack for this issue:
https://github.com/ncstate-delta/moodle-mod_zoom/blob/d4db341b3429c6f3cbd82a3454d852423b7b2d3d/mod_form.php#L51-L54

Try changing `course-editdefaultcompletion` to `course-defaultcompletion` and see if that fixes the issue. We'll still need to keep the "edit" version for backward compatibility until we can get a real fix. I've asked the Moodle team who worked on it for help.

smbader:
Thank you @opitz,

So that we don't break our previous change, we're going to continue to expand our current method.   

Thank you so much for working on this!
Steve

PR #567, opened 2024-01-23
Sort session report data in start_time ascending order

### LINKED ISSUE #409: Session Report - Data is not sorted
Although the data is loaded sequentially, there are situations where the data is loaded out of order. It appears that the data is then displayed in the order in which is was loaded and not in the order in which the sessions occurred (start time ascending).

karenliulll:
Simple fix for #409


jrchamp (approved):
Great simple fix! 🦊

PR #576, opened 2024-03-01
release: v5.2.0

jrchamp:
There's a new duration-based grading method (which may help people who needed automated "attendance"). Philosophically, I don't believe this is the *best* way to evaluate learners, but I'm not the pedagogy expert.

Also, adjusted the heading levels for the CHANGES.md file, because the heading hierarchy on the Moodle plugin page for mod_zoom needed to be improved.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,9 +1,17 @@
-## v5.1.6 ##
+### Releases ###
+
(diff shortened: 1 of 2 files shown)


smbader (approved):
I win!

PR #578, opened 2024-03-04
Bug/differentiate recording versions

### LINKED ISSUE #547: Differentiate between recorded versions of the same meeting?
Hi,

We've started using the record management functionality of this plugin and I noticed that there's no way to know which "version" of the meeting each link is for.
Stuff like "speaker view", "gallery view", etc...
We have around 5-6 versions of each meeting recorded, and in our Moodle all the links show up with the same text:

image

Can you guys add the option to show the version name instead of the "(Video and Audio)" part?
So for example we'll have those 5-6 links like before, but each will have a descriptor at the end like "(Speaker view with shared screen)" etc.
(If possible, please have those as translatable strings)

Would be even better if we could decide system-wide (or even for each Zoom module) which versions we want to show, but I'm sure that'd be a lot more complicated :)

Thanks!

cbounphengsy:
- Updated recording type strings to allow all recording types provided by Zoom.
- Updated allowed recording file types to include TRANSCRIPT, CC, and CHAT in addition to MP4 and M4A files.
- Updated links to recording files to include non-playable file types (TXT files.)
- Updated labels to match their respective recording type strings.

Fixes #547


jrchamp (commented):
Great progress! A few suggestions to help round it out.

classes/task/get_meeting_recordings.php
```diff
@@ -73,8 +73,25 @@ public function execute() {
...
+            'shared_screen' => get_string('recordingtype_shared', 'mod_zoom'),
+            'shared_screen_with_gallery_view' => get_string('recordingtype_shared_gallery', 'mod_zoom'),
+            'shared_screen_with_speaker_view' => get_string('recordingtype_shared_speaker', 'mod_zoom'),
+            'shared_screen_with_speaker_view(CC)' => get_string('recordingtype_shared_speaker_cc', 'mod_zoom'),
+            'sign_interpretation' => get_string('recordingtype_sign', 'mod_zoom'),
+            'speaker_view' => get_string('recordingtype_speaker', 'mod_zoom'),
+            'summary' => get_string('recordingtype_summary', 'mod_zoom'),
+            'summary_next_steps' => get_string('recordingtype_summary_next_steps', 'mod_zoom'),
+            'summary_smart_chapters' => get_string('recordingtype_summary_smart_chapters', 'mod_zoom'),
+            'timeline' => get_string('recordingtype_timeline', 'mod_zoom'),
```
jrchamp, line 94: If creating a reusable function that only translates strings as needed, something like this might be useful:
```php
function zoom_get_recording_type_string($recordingtype) {
    $recordingtypestringmap = [
        // mappings
        'gallery_view' => 'gallery',
       // more mappings
    ];

    // Return some default string in case new recordingtype values are added in the future.
    if (empty($recordingtypestringmap[$recordingtype])) {
        return $recordingtype;
    }

    return get_string($recordingtypestringmap[$recordingtype], 'mod_zoom');
}

File: classes/task/get_meeting_recordings.php [hunk 2]
```diff
@@ -131,7 +148,6 @@ public function execute() {
                 }
```
jrchamp, line 148: When we're checking if the recording exists locally, specifically:
```php
if (isset($localrecordings[$recording->meetinguuid][$recordingid])) {
```

We should check to see if the recording name that is stored locally is the same as the remote recording name. If they are different, we should update the local recording->name and recording->recordingtype. This will update pre-existing data so that it has the correct name and recording type formats (fixing the "My Recording (Audio only) (Audio only)" issue).

jrchamp (commented):
My suggestions don't quite cover everything, so I'll apply these suggestions in a commit.

classes/task/get_meeting_recordings.php
```diff
@@ -109,17 +104,29 @@ public function execute() {
...
+                    $localrecording = $localrecordings[$recording->meetinguuid][$recordingid];
+
+                    if ($localrecording->name !== $zoom->name) {
+                        $updatemeeting = (object) [
+                            'id' => $localrecording->id,
+                            'name' => $zoom->name,
+                        ];
+                        $DB->update_record('zoom_meeting_recordings', $updatemeeting);
+                    }
+
```
jrchamp, line 126: The get_meeting_recordings task only has a one day lookback. So it's unlikely that we're going to get any benefit from this. If anything, the "update the name" code would be better as part of a "fix the existing data" tool that gets run once, manually by an administrator.

```suggestion
                    mtrace('Recording id: ' . $recordingid . ' exists...skipping');
```

classes/task/get_meeting_recordings.php
```diff
@@ -109,17 +104,29 @@ public function execute() {
             $zoomrecordings = $service->get_user_recordings($hostid, $from, $to);
 
             foreach ($zoomrecordings as $recordingid => $recording) {
-                if (isset($localrecordings[$recording->meetinguuid][$recordingid])) {
-                    mtrace('Recording id: ' . $recordingid . ' exists...skipping');
-                    continue;
-                }
-
                 if (empty($meetings[$recording->meetingid])) {
```
jrchamp, line 107: `$meetings` comes from `$hostmeetings` that is built from `$localmeetings`. While this task should try and minimize the amount of API calls that it uses each time, the complete set of `$hostmeetings` would - in some cases - be better than only including meetings that meet the criteria for identifying which hosts to query. Essentially, it would allow us to add recordings to the database for meetings that don't quite meet the criteria, but that - for whatever reason - are still returned by Zoom as part of the get_user_recordings() API call.

This could be as simple as having a $hosts array that only includes host who have meetings which meet the criteria. And then always adding to the hostmeetings array independent of the criteria.

classes/task/get_meeting_recordings.php
jrchamp: All future content should be correct, so the problem is existing content. Recordings do not have their own in the Zoom API, so setting the recording->name to the meeting->name as part of the upgrade.php step would fix all of the entries that we wanted this code to fix in the first place, without adding overhead to regularly scheduled processes.

jrchamp:
Next steps:

- [ ] Do not worry about updating name in task.
- [ ] Update name as part of upgrade.php step.
- [ ] Keep all of a host's meetings available, just in case we stumble upon a recording that should be included.

Future step: Add manually-run fix script to update recodingtype values that are not in the known list (fixes 'audio', 'video' and translated versions of those original strings).

db/upgrade.php
```diff
@@ -965,5 +965,22 @@ function xmldb_zoom_upgrade($oldversion) {
...
+        foreach ($meetings as $meeting) {
+            var_dump($meeting);
+            $value = $meeting->name;
+            $select = "zoomid = ". $meeting->id;
+
+            $DB->set_field_select('zoom_meeting_recordings', 'name', $value, $select);
+        }
+
+        // Zoom savepoint reached.
+        upgrade_mod_savepoint(true, 2024030101, 'zoom');
```
jrchamp, line 982: When we pick a new version number, this should be that number.

classes/task/get_meeting_recordings.php
```diff
@@ -109,17 +104,19 @@ public function execute() {
...
                     // Skip meetings that are not in Moodle.
                     mtrace('Meeting id: ' . $recording->meetingid . ' does not exist...skipping');
                     continue;
                 }
 
+                $zoom = $meetings[$recording->meetingid];
+
+                if (isset($localrecordings[$recording->meetinguuid][$recordingid])) {
+                    mtrace('Recording id: ' . $recordingid . ' exists...skipping');
+                    continue;
```
jrchamp, line 117: In case we load data for a recording and it has a more accurate recording type, update the local recording data to have the better recording type value.
```suggestion
                    mtrace('Recording id: ' . $recordingid . ' exists...skipping');
                    $localrecording = $localrecordings[$recording->meetinguuid][$recordingid];

                    if ($localrecording->recordingtype !== $zoom->recordingtype) {
                        $updatemeeting = (object) [
                            'id' => $localrecording->id,
                            'recordingtype' => $zoom->recordingtype,
                        ];
                        $DB->update_record('zoom_meeting_recordings', $updatemeeting);
                    }
                    continue;
```

db/upgrade.php
```diff
@@ -965,5 +965,22 @@ function xmldb_zoom_upgrade($oldversion) {
...
+
+        // Update existing recording names to default for translatable recordingtype strings.
+        $meetings = $DB->get_records('zoom');
+
+        foreach ($meetings as $meeting) {
+            var_dump($meeting);
+            $value = $meeting->name;
+            $select = "zoomid = ". $meeting->id;
+
+            $DB->set_field_select('zoom_meeting_recordings', 'name', $value, $select);
```
jrchamp, line 978:
```suggestion
            $DB->set_field_select('zoom_meeting_recordings', 'name', $meeting->name, 'zoomid = ?', [$meeting->id]);
```

db/upgrade.php
```diff
@@ -965,5 +965,22 @@ function xmldb_zoom_upgrade($oldversion) {
         upgrade_mod_savepoint(true, 2024030100, 'zoom');
     }
 
+    if ($oldversion < 2024030101) {
+
+        // Update existing recording names to default for translatable recordingtype strings.
+        $meetings = $DB->get_records('zoom');
+
+        foreach ($meetings as $meeting) {
+            var_dump($meeting);
```
jrchamp, line 974:
```suggestion
```

db/upgrade.php
```diff
@@ -965,5 +965,22 @@ function xmldb_zoom_upgrade($oldversion) {
         upgrade_mod_savepoint(true, 2024030100, 'zoom');
     }
 
+    if ($oldversion < 2024030101) {
+
```
jrchamp, line 969:
```suggestion
```

jrchamp (commented):
Very close to the finish!

classes/task/get_meeting_recordings.php
jrchamp: This made sense when we were updating the names, but we have all of the recording information preloaded, so no array juggling is needed

jrchamp (approved):
This is an excellent improvement. Now people can figure out which file is which because the detailed recording type is now stored and only translated on the fly which addresses the "stored admin translation" issue.

PR #584, opened 2024-04-03
Cleanup: move zoom_get_user_display_name function from view.php to locallib

### LINKED ISSUE #580: Cleanup: move zoom_get_user_display_name function out of view.php
Maybe it gets reworked so that it's not a named function. Maybe it's just a complete move from view.php to locallib.php. Either way, it's a bit odd to define a named function inside of what is essentially an output template file.

karenliulll:
Fixes #580


locallib.php
```diff
@@ -1314,3 +1314,30 @@ function zoom_get_registrant_join_url($useremail, $meetingid, $iswebinar) {
 
     return false;
 }
+
+/**
+ * Get the display name for a Zoom user.
+ * This is wrapped in a function to avoid unnecessary API calls.
+ *
+ * @package mod_zoom
```
jrchamp, line 1322: This may be optional now? I'll try and fix the PHPCS issues first before modifying this.

```suggestion
```

jrchamp (approved):
Looks great! Sorry to pollute the commit history with fixes so the tests pass.

PR #585, opened 2024-04-04
Update zoom record after populated from response

### LINKED ISSUE #303: join_url not saved when I change password in activity module
If I change the password in activity module, the password is correctly save in Zoom account and bdd Moodle.
But the join_url isn't saved in bdd moodle.
join_url contains the encrypted password, if a student wants to join a meeting, he will have to enter the password because the one in the url is different than the one registered in the Zoom account.
Otherwise the start_url is always different but no problem. Zoom must surely modify them permanently, maybe a security issue

smbader:
Update the zoom record in Moodle after populating the object from the response. This is good for fields that Zoom may have updated or sent back that aren't a part of the form such as join_url.

Fixes #303

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -202,6 +202,7 @@ function zoom_update_instance(stdClass $zoom, ?mod_zoom_mod_form $mform = null) 
     // Get the updated meeting info from zoom, before updating calendar events.
     $response = zoom_webservice()->get_meeting_webinar_info($zoom->meeting_id, $zoom->webinar);
     $zoom = populate_zoom_from_response($zoom, $response);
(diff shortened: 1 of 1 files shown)


jrchamp (approved):
This extra update will make sure that fields which Zoom is the source of truth for are also updated on the Moodle side. As long as populate_zoom_from_response() does not include any fields for which Moodle should be the source of truth, this should work well.

A future update should be able to reduce the number of update_record calls. For example, if the update_meeting() API call fails, then we don't want to save values locally that Zoom is the source of truth for. However, we would still want to save values locally that Moodle is the source of truth for.

PR #593, opened 2024-05-02
delete recordings: Only delete if API call succeeded

### LINKED ISSUE #539: Recording deleted after upgrading to v5.1.1 (2023101900)
Hi :)

We upgraded last Friday morning to v5.1.1 (2023101900) as we were eager to get the fix https://github.com/ncstate-delta/moodle-mod_zoom/pull/517 as you can imagine.

Unfortunately we had to rollback this morning due to many recordings disappearing from the database during the weekend. They were still properly listed in Zoom, but not in the plugin, so not in Moodle.
We rolled back to the previous 5.1.0 version, and 521 recordings were added back by launching the task 
image

These recordings were not created over the weekend but in the weeks before, hence this post, to hopefully get some help and figure out the issue.

At some point it means the task to delete the recordings did something unexpected, but probably on Friday or during the weekend, and we don't have the full log being "old" so can't point out the moment.

We tried to replicate the issue in a staging area but the task not rolling as often has proven to be difficult.
Still, if we manually delete a recording from weeks ago in staging, it's not fetched by the task, unless we change the delay in get_meeting_recordings.php to something like:
        $from = gmdate('Y-m-d', strtotime('-1 year', $now));
        $to = gmdate('Y-m-d', strtotime('+1 week', $now));
Which makes sense, but doesn't explain the deletions that happened..

Could it be a difference between the API calls get_recording_url_list() and get_user_recordings()? Which are used in the task to delete vs the task to get of the meeting recordings?

Also, are we the only experiencing this issue?

Thanks for your help,

Lucas

jrchamp:
Reviewer: ignoring spaces in the diff will save significant time.

Previously, if the API call failed due to network or rate-limiting issues, the delete recordings task would think that the recording had been deleted from Zoom (and the local entries in Moodle would also be deleted).

One potential concern: will the API return a "that Meeting UUID doesn't exist" error? If it did, then we wouldn't delete the recordings. Maybe we should check that they exist directly? At least now, we're failing into a safer stance.

Fixes #539 (hopefully!)

### DIFF
--- a/classes/task/delete_meeting_recordings.php
+++ b/classes/task/delete_meeting_recordings.php
@@ -72,20 +72,24 @@ class delete_meeting_recordings extends scheduled_task {
         // Get all recordings stored in Moodle, grouped by meetinguuid.
         $zoomrecordings = zoom_get_meeting_recordings_grouped();
         foreach ($zoomrecordings as $meetinguuid => $recordings) {
-            // Now check which recordings still exist on Zoom.
-            $recordinglist = $service->get_recording_url_list($meetinguuid);
-            foreach ($recordinglist as $recordinginfo) {
-                $zoomrecordingid = trim($recordinginfo->recordingid);
-                if (isset($recordings[$zoomrecordingid])) {
-                    mtrace('Recording id: ' . $zoomrecordingid . ' exist(s)...skipping');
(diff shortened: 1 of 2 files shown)


smbader (approved):
Stop doing bad things when bad things happen.

PR #602, opened 2024-07-23
Mobile app: Use module-info instead of module-description 

dpalou:
In this PR I'm removing the ionic3 code (the ionic5 app was released on 2021 so it should be safe to remove it).

I also removed the usage of core-course-module-description component for newer app versions, it is deprecated and will be removed in a future version of the app. It should have been removed already, but we kept it in the app because some plugins still use it.

### DIFF
--- a/classes/output/mobile.php
+++ b/classes/output/mobile.php
@@ -78,6 +78,7 @@ class mobile {
             'option_audio' => $optionaudio,
             'cmid' => $cm->id,
             'courseid' => $args->courseid,
+            'canusemoduleinfo' => $args->appversioncode >= 44000,
         ];
 
         return [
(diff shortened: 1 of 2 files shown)


jrchamp:
Hi @dpalou,

Thank you for opening this PR. The module-info commit is a nice fix for the deprecated module-description. That would qualify as a `bug` fix, so we can add it to a micro/patch release.

Removing support for ionic3 is technically a `bc-break` change, so it would need to wait for the next major release.

I'll move the commits around and rebase on the latest branch that fixes the CI issues, so nothing that you need to do. :heart:

jrchamp (approved):
Looks great! Now newer versions won't be using the deprecated component.

PR #605, opened 2024-07-24
DB Update for Recording Type Limit

### LINKED ISSUE #599: scheduled task "mod_zoom\task\get_meeting_recordings" fails with recordingtype "shared_screen_with_speaker_view'"
We unfortunately have the scheduled task "mod_zoom\task\get_meeting_recordings" failing: If the recording type is "shared_screen_with_speaker_view", we receive the error "ERROR:  value too long for type character varying(30)".

In the database scheme (https://github.com/ncstate-delta/moodle-mod_zoom/blob/main/db/install.xml -> line 127, we see that field recordingtype has 30 characters, "shared_screen_with_speaker_view" however has 31 characters.

cbounphengsy:
- Increased recording type character limit to 50
- Added step within upgrade.php


jrchamp (approved):
Looks great! Thank you!

PR #606, opened 2024-07-25
release: v5.2.3

jrchamp:
Some nice fixes and one crucial regression fix.


smbader (approved):
Excited to have a "good time" with duration language strings.

PR #608, opened 2024-07-29
index: migrate from label to badge (Bootstrap 3 -> 4/5)

### LINKED ISSUE #601: index: In Progress meeting status has low contrast
When a meeting is in progress, the `.zoom-info` CSS class is applied:
https://github.com/ncstate-delta/moodle-mod_zoom/blob/37b958b3e33bad223b64a0649c047e06a66cebd0/index.php#L159

The CSS file applies some padding, a rounded border and makes the text white:
https://github.com/ncstate-delta/moodle-mod_zoom/blob/37b958b3e33bad223b64a0649c047e06a66cebd0/styles.css#L1-L5

In most themes, this will result in white text on a white (or slightly gray) background, which is completely invisible and far from the minimum levels of contrast necessary for accessible content.

This was discovered by @sgrandh3 while testing #250

jrchamp:
`label` was in Bootstrap 3. Since Bootstrap 4 (theme boost; Moodle 3.2+), it's been `badge` and we don't need/want the custom CSS. The suggested change should work in Bootstrap 4.0 - 5.3+ (specifically Moodle 3.2 - 4.6+).

Fixes #601

### DIFF
--- a/index.php
+++ b/index.php
@@ -156,7 +156,7 @@ foreach ($zooms as $z) {
         $oldtable->data[] = $row;
     } else {
         if ($inprogress) {
-            $label = html_writer::tag('span', $strmeetingstarted, ['class' => 'badge bg-info text-dark']);
+            $label = html_writer::tag('span', $strmeetingstarted, ['class' => 'badge bg-dark badge-dark bg-text-dark']);
             $row[2] = html_writer::tag('div', $label);
         } else {
             $row[2] = $displaytime;
--- a/index.php
+++ b/index.php
@@ -156,7 +156,7 @@ foreach ($zooms as $z) {
         $oldtable->data[] = $row;
     } else {
         if ($inprogress) {
-            $label = html_writer::tag('span', $strmeetingstarted, ['class' => 'label label-info zoom-info']);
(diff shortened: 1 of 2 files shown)


smbader:
I like the change, supporting back to 3.2 is great.

As mentioned in our meeting, there's a little bit of contrast back and forth between versions that we need to work out.

jrchamp:
Thanks y'all for testing these with me!

Even with the backward/forward compatibility, `badge bg-info text-dark` was not sufficient contrast for WCAG 2.x AA in some versions. Forcing `text-dark` also had the drawback of not working well in "dark mode" themes. So if we can't use `text-dark`, then Bootstrap 5.0 and 5.1 will need to use a background style that has sufficient contrast with the default text color. That leaves "primary", "secondary", "success", "danger" and "dark". primary and secondary are already used for buttons to Join or Cancel. success did not have sufficient contrast in some of the versions that we tested. danger is specifically designed to scare/alert people, so that felt incorrect. Thus, dark is the only option that has sufficient contrast, isn't designed to scare people, and does not conflict with the mental model of action buttons.

smbader (approved):
A good choice for both visibility and backwards compatibility

PR #611, opened 2024-08-01
CI: Check Moodle 4.4 and PHP 8.3

jrchamp:
Removes Moodle 3.11 and PHP 7.3 from the matrix, but we still test Moodle 3.9 (LTS), Moodle 4.0, PHP 7.2 and PHP 7.4, so it should be okay. 🤞


smbader (approved):
8.3 support, LTS support, and bon voyage, 3.11.

PR #612, opened 2024-08-01
Add IDs to activity details sections and rows

### LINKED ISSUE #597: provide meaningful HTML ids for the details of the activity
Thanks for the great plugin.

It would be very useful if the HTML-page for the meeting details contained ids that I could reference with JavaScript or CSS. There are some examples where we need to modify the page containing the button to enter the meeting dynamically based on specific needs of the teachers

* we want to force the students to enter the meeting via the button so that the activity can complete and unlock other activities in the course. clicking the URL of the meeting does not trigger the activity completion.
* sometimes it is necessary to display the meeting details to students when the button fails to open the meeting and the meeting details have to be entered manually in the zoom app.

For those cases having ids in the HTML (for the tables and the specific rows) would be very helpful
* ids for each section (h3 and the table below) - or adding a div containing the h3 and the subsequent table
* ids for each row of the table (meeting id, link, code, time, etc.)

armccoy:
- Wrap each activity details section (Schedule, Security, Media) in a `<div>` with a unique ID
- Add unique IDs to table rows
- Define labels in left table columns as `<th>` 

Fixes #597

### DIFF
--- a/view.php
+++ b/view.php
@@ -245,6 +245,7 @@ if (!$showrecreate) {
 }
 
 if ($zoom->show_schedule) {
+    echo $OUTPUT->box_start('', 'section-schedule');
     // Output "Schedule" heading.
     echo $OUTPUT->heading(get_string('schedule', 'mod_zoom'), 3);
 
@@ -256,21 +257,42 @@ if ($zoom->show_schedule) {
     $numcolumns = 2;
 
     // Show start/end date or recurring meeting information.
+    $rowmeetingtime = new html_table_row();
+    $rowmeetingtime->id = 'schedule-row-meetingtime';
+    $thmeetingtime = new html_table_cell();
+    $thmeetingtime->header = true;
+    $tdmeetingtime = new html_table_cell();
+
     if ($isrecurringnotime) {
-        $table->data[] = [get_string('recurringmeeting', 'mod_zoom'), get_string('recurringmeetingexplanation', 'mod_zoom')];
+        $thmeetingtime->text = get_string('recurringmeeting', 'mod_zoom');
+        $tdmeetingtime->text = get_string('recurringmeetingexplanation', 'mod_zoom');
     } else if ($zoom->recurring && $zoom->recurrence_type != ZOOM_RECURRINGTYPE_NOTIME) {
-        $table->data[] = [get_string('recurringmeeting', 'mod_zoom'), get_string('recurringmeetingthisis', 'mod_zoom')];
+        $thmeetingtime->text = get_string('recurringmeeting', 'mod_zoom');
+        $tdmeetingtime->text = get_string('recurringmeetingthisis', 'mod_zoom');
         $nextoccurrence = zoom_get_next_occurrence($zoom);
         if ($nextoccurrence > 0) {
-            $table->data[] = [get_string('nextoccurrence', 'mod_zoom'), userdate($nextoccurrence)];
+            $thmeetingtime->text = get_string('nextoccurrence', 'mod_zoom');
+            $tdmeetingtime->text = userdate($nextoccurrence);
         } else {
-            $table->data[] = [get_string('nextoccurrence', 'mod_zoom'), get_string('nooccurrenceleft', 'mod_zoom')];
+            $thmeetingtime->text = get_string('nextoccurrence', 'mod_zoom');
+            $tdmeetingtime->text = get_string('nooccurrenceleft', 'mod_zoom');
         }
-
-        $table->data[] = [$strduration, format_time($zoom->duration)];
     } else {
-        $table->data[] = [$strtime, userdate($zoom->start_time)];
-        $table->data[] = [$strduration, format_time($zoom->duration)];
+        $thmeetingtime->text = $strtime;
+        $tdmeetingtime->text = userdate($zoom->start_time);
+    }
+
+    $rowmeetingtime->cells = [$thmeetingtime, $tdmeetingtime];
+    $table->data[] = $rowmeetingtime;
+
+    // Show meeting duration
+    if (!$isrecurringnotime) {
+        $rowduration = new html_table_row();
+        $rowduration->id = 'schedule-row-duration';
+        $thduration = new html_table_cell($strduration);
+        $thduration->header = true;
+        $rowduration->cells = [$thduration, format_time($zoom->duration)];
+        $table->data[] = $rowduration;
     }
 
     // Show recordings section if option enabled to view recordings.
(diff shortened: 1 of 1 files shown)


view.php
```diff
@@ -256,21 +257,42 @@
...
+        $thmeetingtime->text = get_string('recurringmeeting', 'mod_zoom');
+        $tdmeetingtime->text = get_string('recurringmeetingthisis', 'mod_zoom');
         $nextoccurrence = zoom_get_next_occurrence($zoom);
         if ($nextoccurrence > 0) {
-            $table->data[] = [get_string('nextoccurrence', 'mod_zoom'), userdate($nextoccurrence)];
+            $thmeetingtime->text = get_string('nextoccurrence', 'mod_zoom');
+            $tdmeetingtime->text = userdate($nextoccurrence);
         } else {
-            $table->data[] = [get_string('nextoccurrence', 'mod_zoom'), get_string('nooccurrenceleft', 'mod_zoom')];
+            $thmeetingtime->text = get_string('nextoccurrence', 'mod_zoom');
```
jrchamp, line 277: This "nextoccurrence" string is in the `if` and the `else` probably safe to move it before/outside the `if`.

view.php
```diff
@@ -352,27 +398,36 @@
                 }
 
                 // Output the concatenated string of alternative hosts.
-                $table->data[] = [get_string('alternative_hosts', 'mod_zoom'), $alternativehoststring];
+                $rowshowalternativehosts->cells = [$thalternativehosts, $alternativehoststring];
 
                 // Otherwise we stick with the plain list of email addresses as we got it from Zoom directly.
             } else {
-                $table->data[] = [get_string('alternative_hosts', 'mod_zoom'), $zoom->alternative_hosts];
+                $rowshowalternativehosts->cells = [$thalternativehosts, $zoom->alternative_hosts];
```
jrchamp, line 405: Same idea here. The `$rowshowalternativehosts->cells = [$thalternativehosts, ` part could be moved after the if/else if desired.

view.php
```diff
@@ -407,30 +477,52 @@
...
+        $table->data[] = $rowjoinbeforehost;
     }
 
     // Show authentication.
-    $table->data[] = [$strauthenticatedusers, ($zoom->option_authenticated_users) ? $stryes : $strno];
+    $rowauthenticatedusers = new html_table_row();
+    $rowauthenticatedusers->id = 'security-row-authenticatedusers';
+    $thauthenticatedusers = new html_table_cell($strauthenticatedusers);
+    $thauthenticatedusers->header = true;
+    $rowauthenticatedusers->cells = [$thauthenticatedusers, $zoom->option_authenticated_users];
```
jrchamp, line 516: Seems like it's going to show "1" or "0". The original behavior would probably be:
```suggestion
    $rowauthenticatedusers->cells = [$thauthenticatedusers, $zoom->option_authenticated_users ? $stryes : $strno];
```

The reason this one is different from some of the others is that the Yes/No value is saved into a variable.
```suggestion
    $strauthenticatedusers = $zoom->option_authenticated_users ? $stryes : $strno;
    $rowauthenticatedusers->cells = [$thauthenticatedusers, $strauthenticatedusers];
```

jrchamp (commented):
Lots of nice improvements! Only a few comments/ideas

view.php
```diff
@@ -245,6 +245,7 @@
 }
 
 if ($zoom->show_schedule) {
+    echo $OUTPUT->box_start('', 'section-schedule');
```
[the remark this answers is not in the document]
armccoy: I like this idea. Additionally, it could be used to prefix the row IDs after dropping the "-row" as suggested.

view.php
```diff
@@ -256,21 +257,42 @@
     $numcolumns = 2;
 
     // Show start/end date or recurring meeting information.
+    $rowmeetingtime = new html_table_row();
+    $rowmeetingtime->id = 'schedule-row-meetingtime';
```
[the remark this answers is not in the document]
armccoy: See comment for line 248.

view.php
```diff
@@ -256,21 +257,42 @@
...
+    $thmeetingtime = new html_table_cell();
+    $thmeetingtime->header = true;
+    $tdmeetingtime = new html_table_cell();
```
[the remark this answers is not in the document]
armccoy: I used `th` and `td` to indicate the header text and cell text, but your suggestion is definitely more easily understood.

view.php
```diff
@@ -256,21 +257,42 @@
...
+    $meetingtimetext = new html_table_cell();
+
     if ($isrecurringnotime) {
-        $table->data[] = [get_string('recurringmeeting', 'mod_zoom'), get_string('recurringmeetingexplanation', 'mod_zoom')];
+        $meetingtimeheader->text = get_string('recurringmeeting', 'mod_zoom');
+        $meetingtimetext->text = get_string('recurringmeetingexplanation', 'mod_zoom');
     } else if ($zoom->recurring && $zoom->recurrence_type != ZOOM_RECURRINGTYPE_NOTIME) {
-        $table->data[] = [get_string('recurringmeeting', 'mod_zoom'), get_string('recurringmeetingthisis', 'mod_zoom')];
+        $meetingtimeheader->text = get_string('recurringmeeting', 'mod_zoom');
+        $meetingtimetext->text = get_string('recurringmeetingthisis', 'mod_zoom');
```
jrchamp, line 271: This pair needs to be added to the cells array before the other pair of cells.

jrchamp (commented):
Looks great! One comment and then I believe we can merge.

view.php
jrchamp: I just realized it probably needs its own row and cells, so this might be what I was recommending:
```suggestion
        $meetingrecurringheader = new html_table_cell();
        $meetingrecurringheader->header = true;
        $meetingrecurringheader->text = get_string('recurringmeeting', 'mod_zoom');
        $meetingrecurringtext = new html_table_cell();
        $meetingrecurringtext->text = get_string('recurringmeetingthisis', 'mod_zoom');
        $rowmeetingrecurring = new html_table_row();
        $rowmeetingrecurring ->id = 'zoom_schedule-meetingrecurring';
        $rowmeetingrecurring->cells = [$meetingrecurringheader, $meetingrecurringtext];
        $table->data[] = $rowmeetingrecurring;
```

armccoy: I see now. Yeah, that makes sense. Thanks!

view.php
```diff
@@ -256,21 +257,49 @@
...
+        $meetingtimetext->text = get_string('recurringmeetingexplanation', 'mod_zoom');
     } else if ($zoom->recurring && $zoom->recurrence_type != ZOOM_RECURRINGTYPE_NOTIME) {
-        $table->data[] = [get_string('recurringmeeting', 'mod_zoom'), get_string('recurringmeetingthisis', 'mod_zoom')];
+        $meetingrecurringheader = new html_table_cell();
+        $meetingrecurringheader->header = true;
+        $meetingrecurringheader->text = get_string('recurringmeeting', 'mod_zoom');
+        $meetingrecurringtext = new html_table_cell();
+        $meetingrecurringtext->text = get_string('recurringmeetingthisis', 'mod_zoom');
+        $rowmeetingrecurring = new html_table_row();
+        $rowmeetingrecurring ->id = 'zoom_schedule-meetingrecurring';
```
jrchamp, line 276:
```suggestion
        $rowmeetingrecurring->id = 'zoom_schedule-meetingrecurring';
```

jrchamp (approved):
Looks great! Approving and merging so it can go to integration testing.

PR #615, opened 2024-08-27
Add moodle filter adaption to the meeting topic and description

### LINKED ISSUE #610: Moodle display filters for zoom meeting topics and description
Although I understand that zoom topics and descriptions don't take html, they are populated by data entered into zoom activities which are processed through Moodle filters when viewed through Moodle. These filters can introduce html, but some filters (like [filterCodes](https://moodle.org/plugins/filter_filtercodes) and [Generico](https://moodle.org/plugins/filter_generico)) can also be used to template dynamic text, or template common, or variable output depending on context.

For instance, {coursename} is automatically expanded to the full name of the current course, which can be very handy when naming Zooms without having to adde the course name manually.

Similarly, if you want to include dynamic text to be shown when on mobile or not you can use {ifmobile}{/ifmobile} or {ifnotmobile}{/ifnotmobile} to wrap text to be rendered appropriately.

It is also very convenient to template common blocks of text using generico so you can maintain how something is worded globally instead of having to update the same text individually in every course and/or activity in which it is used.

At the moment these shortcodes all show up in zoom topics and descriptions in zoom (without being expanded by the appropriate Moodle filters), despite being rendered properly when viewed in Moodle in the Zoom Activity.

You can learn more about available Moodle filters here: https://docs.moodle.org/404/en/Filters

It appears one can use format=FORMAT_PLAIN when using filters to output filtered text without introducing html to the processed string using the filter api.

https://moodledev.io/docs/4.4/apis/subsystems/output#string-formatting-functions

Specifically it appears that format_string() can be used to preprocess strings before they are set to the topic and description in Zoom (outside of Moodle).

https://moodledev.io/docs/4.4/apis/subsystems/output#format_string

### LINKED ISSUE #616: Respect Zoom's API limits on topic and agenda
From the [Create a meetings API documentation](https://developers.zoom.us/docs/api/rest/reference/zoom-api/methods/#operation/meetingCreate), `topic` cannot exceed 200 characters and `agenda` cannot exceed 2000.

https://github.com/ncstate-delta/moodle-mod_zoom/blob/aa8208402f29574cbecd8de594ce520646c5ec36/classes/webservice.php#L594-L601

karenliulll:
Process Zoom meeting topic and description with proper Moodle filters such as [filterCodes](https://moodle.org/plugins/filter_filtercodes) and [Generico](https://moodle.org/plugins/filter_generico) when being viewed outside of Moodle (tested on our NC State Zoom site).

Fixes #610 
Fixes #616

### DIFF
--- a/backup/moodle2/restore_zoom_stepslib.php
+++ b/backup/moodle2/restore_zoom_stepslib.php
@@ -65,7 +65,9 @@ class restore_activity_structure_step extends \restore_activity_structure_step {
 
         // Either create a new meeting or set meeting as expired.
         try {
-            $updateddata = zoom_webservice()->create_meeting($data);
+            // FIXME: Do we provide course context? That won't have the right activity names etc.
+            $cmid = null;
+            $updateddata = zoom_webservice()->create_meeting($data, $cmid);
             $data = populate_zoom_from_response($data, $updateddata);
             $data->exists_on_zoom = ZOOM_MEETING_EXISTS;
         } catch (moodle_exception $e) {
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -598,7 +598,7 @@ class webservice {
 
         $data = [
             // Process the meeting topic with proper filter.
-            'topic' => substr(format_string($zoom->name, true, $options + ['escape' => false]), 0, 200),
+            'topic' => zoom_apply_filter_on_meeting_name($zoom->name, $options),
             'settings' => [
                 'host_video' => (bool) ($zoom->option_host_video),
                 'audio' => $zoom->option_audio,
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -590,8 +590,14 @@ class webservice {
     private function database_to_api($zoom) {
         global $CFG;
 
-        [$course, $cm] = get_course_and_cm_from_instance($zoom, 'zoom');
-        $context = \context_module::instance($cm->id);
+        if (isset($zoom->coursemodule)) {
+            $cmid = $zoom->coursemodule;
+        } else {
+            [$course, $cm] = get_course_and_cm_from_instance($zoom, 'zoom');
+            $cmid = $cm->id;
+        }
+
+        $context = \context_module::instance($cmid);
 
         $data = [
             // Process the meeting topic with proper filter.
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -601,7 +601,7 @@ class webservice {
 
         $data = [
             // Process the meeting topic with proper filter.
-            'topic' => format_text($zoom->name, FORMAT_MOODLE, ['context' => $context, 'para' => false]),
+            'topic' => substr(format_text($zoom->name, FORMAT_MOODLE, ['context' => $context, 'para' => false]), 0, 200),
             'settings' => [
                 'host_video' => (bool) ($zoom->option_host_video),
                 'audio' => $zoom->option_audio,
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -585,23 +585,20 @@ class webservice {
      * database fields to the appropriate API request fields.
      *
      * @param stdClass $zoom The zoom meeting to format.
+     * @param ?int $cmid The cmid if available.
      * @return array The formatted meetings for the meeting.
      */
-    private function database_to_api($zoom) {
+    private function database_to_api($zoom, $cmid) {
         global $CFG;
 
-        if (isset($zoom->coursemodule)) {
-            $cmid = $zoom->coursemodule;
-        } else {
-            [$course, $cm] = get_course_and_cm_from_instance($zoom, 'zoom');
-            $cmid = $cm->id;
+        $options = [];
+        if (!empty($cmid)) {
+            $options['context'] = \context_module::instance($cmid);
         }
-
-        $context = \context_module::instance($cmid);
 
         $data = [
             // Process the meeting topic with proper filter.
-            'topic' => substr(format_string($zoom->name, true, ['context' => $context, 'escape' => false]), 0, 200),
+            'topic' => substr(format_string($zoom->name, true, $options + ['escape' => false]), 0, 200),
             'settings' => [
                 'host_video' => (bool) ($zoom->option_host_video),
                 'audio' => $zoom->option_audio,
@@ -612,7 +609,7 @@ class webservice {
             $data['agenda'] = substr(content_to_text(format_text(
                 $zoom->intro,
                 FORMAT_MOODLE,
-                ['context' => $context]
+                $options
             ), false), 0, 2000);
         }
 
@@ -780,9 +777,10 @@ class webservice {
      * Take a $zoom object as returned from the Moodle form and respond with an object that can be saved to the database.
      *
      * @param stdClass $zoom The meeting to create.
+     * @param ?int $cmid The cmid if available.
      * @return stdClass The call response.
      */
-    public function create_meeting($zoom) {
+    public function create_meeting($zoom, $cmid) {
         // Provide license if needed.
         $this->provide_license($zoom->host_id);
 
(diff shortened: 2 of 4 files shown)


lib.php
```diff
@@ -290,7 +290,6 @@ function populate_zoom_from_response(stdClass $zoom, stdClass $response) {
     }
 
     $newzoom->meeting_id = $response->id;
-    $newzoom->name = $response->topic;
```
jrchamp, line 293: We added this change because the name was being passed through the filter and then overridden in Moodle. It was leftover from when we used to synchronized with Zoom's topic string.

classes/webservice.php
```diff
@@ -590,15 +590,30 @@ public function get_schedule_for_users($identifier) {
     private function database_to_api($zoom) {
         global $CFG;
 
+        if (isset($zoom->coursemodule)) {
+            $cmid = $zoom->coursemodule;
```
jrchamp, line 594: When creating a new activity. get_course_and_cm_from_instance() was failing. We checked to see what the form was providing and `coursemodule` had the cmid that we needed. It would be great to know if this is a reliable way to get this data (i.e. Does Moodle support us using it this way? Do core modules ever do this?).

Or perhaps there's a different function that will give us the $context in a simpler way that works during add_instance and update_instance?

jrchamp (commented):
Thanks @karenliulll ! There were a couple bumps during our testing, but it's definitely doing some nice things for the Zoom description/agenda. Hopefully we can find a more-reliable way to handle the context piece and determine if Moodle activity names are supported by filters.

classes/webservice.php
```diff
@@ -590,15 +590,30 @@ public function get_schedule_for_users($identifier) {
...
+            [$course, $cm] = get_course_and_cm_from_instance($zoom, 'zoom');
+            $cmid = $cm->id;
+        }
+
+        $context = \context_module::instance($cmid);
+
         $data = [
-            'topic' => $zoom->name,
+            // Process the meeting topic with proper filter.
+            'topic' => substr(format_text($zoom->name, FORMAT_MOODLE, ['context' => $context, 'para' => false]), 0, 200),
```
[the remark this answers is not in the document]
behanw: The Activity name appears to be processed by Moodle filters for me. If I include {coursename} in an activity name it is expanded appropriately when viewed in Moodle.

In the "Manage Filters" settings page you can choose whether to apply filters to only "Content" or, "Content and Headers". You just may not have it set to process your headers for those filters.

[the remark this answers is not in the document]
jrchamp: Thanks @behanw! I did not fully understand the filters before. We should only be using `format_text()` on the description/agenda. For the title, we should be using `format_string()`. That will make sure that the Manage Filters settings for "Content" vs "Content and headings" correctly matches the Moodle web page's filter behavior.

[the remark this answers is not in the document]
behanw: Honestly, there are so many features and settings in Moodle that it's sometimes hard to figure out how to do things. I also suspect everyone who uses Moodle does it a little differently. It certainly took me a while to fully appreciate what was possible with filters, and I probably rely on them too much at times. ;)

Thank you for your wonderful Zoom plugin. It is very appreciated, and I doubt we fully take advantage of all the features it allows for yet!

lib.php
```diff
@@ -187,7 +187,7 @@ function zoom_update_instance(stdClass $zoom, ?mod_zoom_mod_form $mform = null)
 
     // Update meeting on Zoom.
     try {
-        zoom_webservice()->update_meeting($zoom);
+        zoom_webservice()->update_meeting($zoom, $mform->_cm->id);
```
jrchamp, line 190: Assumes $mform will not be null. Would it be safe to null coalesce to `$zoom->coursemodule`? The core module mod_assign seems to use that value.

lib.php
```diff
@@ -99,7 +99,7 @@ function zoom_add_instance(stdClass $zoom, ?mod_zoom_mod_form $mform = null) {
         $zoom->breakoutrooms = $breakoutrooms['zoom'];
     }
 
-    $response = zoom_webservice()->create_meeting($zoom);
+    $response = zoom_webservice()->create_meeting($zoom, $mform->_cm->id);
```
jrchamp, line 102: Assumes $mform will not be null. Would it be safe to null coalesce to `$zoom->coursemodule`? The core module mod_assign seems to use that value.

backup/moodle2/restore_zoom_stepslib.php
```diff
@@ -65,7 +65,9 @@ protected function process_zoom($data) {
 
         // Either create a new meeting or set meeting as expired.
         try {
-            $updateddata = zoom_webservice()->create_meeting($data);
+            // FIXME: Do we provide course context? That won't have the right activity names etc.
+            $cmid = null;
+            $updateddata = zoom_webservice()->create_meeting($data, $cmid);
```
jrchamp, line 70: What happens here? For easier testing, I think that the "duplicate" action uses the backup/restore process, but this code needs to be tested. I'm not sure when the context exists so that we can use it to filter...

jrchamp (commented):
This PR looks great overall and fixes two existing edge cases. The only concern I have now is getting the module context in a reliable way so that the filters can produce the expected output.

jrchamp:
This should be tested to see if/how well the code paths are working:

- [ ] Backup/restore activity
- [ ] Duplicate activity
- [ ] Create new activity
- [ ] Edit existing activity
- [ ] Quick-edit (for example, the quick-rename functionality on the main course page)
- [ ] Recreate expired meeting

karenliulll:
I did some quick testing and came across the following issues - 

1. Error occurred for both creating and updating meetings when testing the code - "Exception - Cannot access protected property mod_zoom_mod_form::$_cm". Should be related to the code changes to the zoom_webservice()->create_meeting() and update_meeting() functions.
2. Quick edit of Zoom meeting title on the main course page will not auto-update the title on the Zoom site outside of Moodle (does quick edit also make the API call like in full edit?)
3. when performing course backup and restore, the meetings in the new course will update the title and description accordingly in Moodle, but not on the Zoom site - for example, if the {coursename} filter is used in the meeting description, it will display the new course name in Zoom but still show the old one on Zoom site. (Looks like the restore function should make the API call via the zoom_webservice()->create_meeting() function, but somehow it does not update the content on the Zoom site. Is it related to the error in item 1 above?)

jrchamp:
> I did some quick testing and came across the following issues
Thank you! 🥂 
 
> 1. Error occurred for both creating and updating meetings when testing the code - "Exception - Cannot access protected property mod_zoom_mod_form::$_cm". Should be related to the value $mform->_cm->id being passed to the zoom_webservice()->create_meeting() and update_meeting() functions in lib.php.

 - $mform->get_context() will provide course context on add and module context on update
 - $mform->get_coursemodule() will provide the $cm on update, but is expected to be null on add. 😞 

> 2. Quick edit of Zoom meeting title on the main course page will not auto-update the title on the Zoom site outside of Moodle (does quick edit also make the API call like in full edit?)

It should, but the `$zoom->name = $zoom->topic;` line was removed (which is still correct). The following code block needs to be modified:

https://github.com/ncstate-delta/moodle-mod_zoom/blob/8dc452f28c3c60c99a1666b825df0837470f8c46/lib.php#L426-L429

My idea is: "Move the `substr(format_string($zoom->name, true, $options + ['escape' => false]), 0, 200),` code to a helper function, such as a lib.php function." That will allow us to perform the same filtering/modifications on the activity name that we do within the webservice so that the values will be comparable. Please check if the `$fullzoom->name = $zoom->name;` line should be removed (it only affects the calendar and grade item names, so if filters are not being applied to those strings then we should keep the line).

> 3. when performing course backup and restore, the meetings in the new course will update the title and description accordingly in Moodle, but not on the Zoom site - for example, if the {coursename} filter is used in the meeting description, it will display the new course name in Zoom but still show the old one on Zoom site. (Looks like the restore function should make the API call via the zoom_webservice()->create_meeting() function, but somehow it does not update the content outside of Moodle.)

I don't know what the sequence is or what the context is as the point when the restore steps run. And it's possible that the full course backup and restore would behave differently than an individual activity backup/restore and even a activity duplicate action. However, I do think that the fixes for question "2" might significantly help fix some of these issues because the "calculated" name will be updated semi-regularly by Moodle and send the updates into Zoom.

lib.php
jrchamp: Although `$zoom->coursemodule` appears to be user controlled, if it is always available during both update and add, and the permissions are already being checked by Moodle to determine that the user has the permission to update/add that item, but it may be safe and best to use `$zoom->coursemodule`:
```suggestion
        zoom_webservice()->update_meeting($zoom, $zoom->coursemodule);
```

lib.php
jrchamp: If this is safe (security wise) and reliable (see previous comment):
```suggestion
    $response = zoom_webservice()->create_meeting($zoom, $zoom->coursemodule);
```

karenliulll:
> This should be tested to see if/how well the code paths are working:
> 
> * [x]  Backup/restore activity
> * [x]  Duplicate activity
> * [x]  Create new activity
> * [x]  Edit existing activity
> * [x]  Quick-edit (for example, the quick-rename functionality on the main course page)
> * [x]  Recreate expired meeting

I tested all the above after the latest code change, it works fine for me.

backup/moodle2/restore_zoom_stepslib.php
jrchamp: I'm not sure how to make this *perfect*, but this is definitely not making things worse. Any restored activity should self-correct after an edit or refresh occurs.

jrchamp (approved):
Thank you so much @karenliulll!

This journey was much more complicated than expected, but the resulting code is definitely an improvement. When we know the context, we are now correctly filtering it without any significant overhead.

PR #620, opened 2024-09-19
webservice: use string 'null' when recording_type is null

### LINKED ISSUE #619: Task Fail: Get meeting recordings from Zoom: null value in column "recordingtype" violates not-null constraint
Moodle 4.1.12 (Build: 20240812)
Zoom Meeting v5.2.32024072500

Get meeting recordings from Zoom (mod_zoom\task\get_meeting_recordings) fails with:

**Error: null value in column "recordingtype" violates not-null constraint** 

This is halting all other meeting recording retrievals.

Since the last update had to do with a modification to recordingtype I am wondering if there was a regression. I haven't been able to find any other bug reports or discussions around this specific error yet.

**Update**: I was able to locate the Zoom meeting activity in Moodle and I can't find anything out of the norm in the settings. It is set up to be a recurring meeting. There have been 4 sessions so far with this particular ID and the previous recordingtype values were set to chat_file. 

Also, a breakout room was created but no participants added. I tried to delete the breakout room just for the fun of it but could not. No error, the room just wouldn't delete after I saved the activity. I tried same in a different activity. Checked system logs, no nginx errors. Unsure if any of this could be related to the original problem of the task fail, but thinking more info the better.

Hoping that additional information helps.

jrchamp:
This allows the data to be loaded in the database and not throw an exception. In theory, the value should never be null, but somehow it is for one of our users. In the rare case where this occurs, Moodle may show the string "null", but I'm not too worried about that being the worst-case.

Fixes #619

Marking this as draft until we have a better idea of what's going on. It shouldn't cause any problems, but it would be nice to know more.

### DIFF
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -1070,7 +1070,7 @@ class webservice {
                     $recordinginfo->meetinguuid = $response->uuid;
                     $recordinginfo->url = $url;
                     $recordinginfo->filetype = $recording->file_type;
-                    $recordinginfo->recordingtype = $recording->recording_type;
+                    $recordinginfo->recordingtype = $recording->recording_type ?? 'null';
                     $recordinginfo->passcode = $response->password;
                     $recordinginfo->recordingstart = strtotime($recording->recording_start);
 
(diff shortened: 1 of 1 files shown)


armccoy (approved):
Looks good to me!

PR #623, opened 2024-10-01
Send ical Notifications (Version 2)

paulandm:
A refactored / cleaned up version of the original PR for the Send ical notifications task.

- The notifications are now based on the zoom event record (instead of the zoom record).

- Was able to incorporate the 'zoom_helper_icalendar_event' function in the creation of the ical event.

- The buffer of 10 minutes before sending out invites remain in place. The idea is still that an ical invite should only be sent once per Zoom event. Since we don't have access to the participants' calendars, we can't actually update already sent ical invites. So if we change the code to send ical invites for a Zoom activity edit action (after the 10 minute buffer period), it will send an additional ical invite instead of actually changing the previous one sent. This might be appropriate behavior for some, but it is not applicable in our use case. @jrchamp I'm happy to discuss some ideas around this 'update' scenario further if required.

- The buffer period of 10 minutes could also potentially move out to a setting if preferred?


paulandm:
I see the 'MESSAGE_DEFAULT_ENABLED' constant in the message provider is causing problems with Moodle 3.9 compatibility. Happy to remove that if necessary.

db/upgrade.php
```diff
@@ -998,6 +998,21 @@ function xmldb_zoom_upgrade($oldversion) {
         // Launch change of precision for field recordingtype.
         $dbman->change_field_precision($table, $field);
 
+        // Conditionally create the Zoom iCal Notification table.
```
paulandm, line 1001: This section should obviously move out when the latest version number is created.

paulandm: Hi Steve

Thanks, I've moved the code into a new savepoint block and updated the version.php file to match.

classes/task/send_ical_notifications.php
```diff
@@ -0,0 +1,293 @@
...
+            foreach ($coursemodules as $coursemod) {
+                if ($coursemod->modname == 'zoom' && $coursemod->instance == $zoomid) {
+                    $availinfo = new \core_availability\info_module($coursemod);
+                    $users = $availinfo->filter_user_list($users);
+                    break;
+                }
+            }
+        }
+        return $users;
+    }
```
jrchamp, line 282: This private method can be removed if the previous suggestion works.

classes/task/send_ical_notifications.php
```diff
@@ -0,0 +1,293 @@
...
+        if ($zoom->registration == ZOOM_REGISTRATION_OFF) {
+            $icalevent->add_property('location', $zoom->join_url);
+        } else {
+            $registrantjoinurl = zoom_get_registrant_join_url($user->email, $zoom->meeting_id, $zoom->webinar);
+            if ($registrantjoinurl) {
+                $icalevent->add_property('location', $registrantjoinurl);
+            } else {
+                $icalevent->add_property('location', $zoom->join_url);
+            }
+        }
```
jrchamp, line 214: Instead of providing the stored join_url, it's probably better to send the user to the Moodle site's join URL. That way they won't circumvent the authentication/logging mechanisms or be left with an outdated/cached version of the URL.

jrchamp (commented):
Hi @paulandm! Thank you for all your work on this. I had a few ideas that took a while to collect. I may have missed some things, so apologies if I have more comments later on items that were already present in this version of the code.

db/install.xml
```diff
@@ -184,5 +184,16 @@
         <KEY NAME="fk_breakoutroomid" TYPE="foreign" FIELDS="breakoutroomid" REFTABLE="zoom_meeting_breakout_rooms" REFFIELDS="id"/>
       </KEYS>
     </TABLE>
+    <TABLE NAME="zoom_ical_notifications" COMMENT="Identifies the zoom event for which ical notifications have been emailed.">
+      <FIELDS>
+        <FIELD NAME="id" TYPE="int" LENGTH="10" NOTNULL="true" SEQUENCE="true"/>
+        <FIELD NAME="zoomeventid" TYPE="int" LENGTH="10" NOTNULL="true" SEQUENCE="false"/>
+        <FIELD NAME="executiontime" TYPE="int" LENGTH="12" NOTNULL="true" SEQUENCE="false" COMMENT="The time when the send ical notifications task was completed successfully."/>
```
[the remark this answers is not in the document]
paulandm: Thanks for picking this up. I've added a unique index to the table for the zoomeventid field.

classes/task/send_ical_notifications.php
```diff
@@ -0,0 +1,293 @@
...
+     * Get an array of users in the format of userid=>user object.
+     * @param string $zoomid The zoom instance id.
+     * @param string $courseid The course id of the course in which the zoom event occurred.
+     * @return array An array of users.
+     */
+    private function get_users_to_notify($zoomid, $courseid) {
+        global $DB;
+        $users = [];
+
+        $sql = 'SELECT distinct ue.userid
```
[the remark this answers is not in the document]
paulandm: Thank you. This is much cleaner.

paulandm:
Implemented the PR feedback.

db/messages.php
```diff
@@ -26,4 +26,12 @@
 
 $messageproviders = [
     'teacher_notification' => [],
+    // The ical notifications task messages.
+    'ical_notifications' => [
+        'defaults' => [
+            'popup' => MESSAGE_DISALLOWED,
+            'email' => MESSAGE_PERMITTED,
```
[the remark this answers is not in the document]
paulandm: Thank you. Will implement this.

db/upgrade.php
```diff
@@ -1002,5 +1002,25 @@ function xmldb_zoom_upgrade($oldversion) {
...
+        $table = new xmldb_table('zoom_ical_notifications');
+
+        if (!$dbman->table_exists($table)) {
+            $table->add_field('id', XMLDB_TYPE_INTEGER, '10', null, XMLDB_NOTNULL, XMLDB_SEQUENCE, null);
+            $table->add_field('zoomeventid', XMLDB_TYPE_INTEGER, '10', null, XMLDB_NOTNULL, null, null);
+            $table->add_field('executiontime', XMLDB_TYPE_INTEGER, '12', null, null, null, null);
+
+            $table->add_key('primary', XMLDB_KEY_PRIMARY, ['id']);
+            $table->add_key('fk_zoomeventid', XMLDB_KEY_FOREIGN, ['zoomeventid'], 'event', ['id']);
+            $table->add_key('unique_zoomeventid', XMLDB_KEY_UNIQUE, ['zoomeventid']);
```
jrchamp, line 1016: Note to self: double check that this is the name created by XMLDB.

db/install.xml
```diff
@@ -184,5 +184,17 @@
         <KEY NAME="fk_breakoutroomid" TYPE="foreign" FIELDS="breakoutroomid" REFTABLE="zoom_meeting_breakout_rooms" REFFIELDS="id"/>
       </KEYS>
     </TABLE>
+    <TABLE NAME="zoom_ical_notifications" COMMENT="Identifies the zoom event for which ical notifications have been emailed.">
```
jrchamp, line 187: Note to self: Generate this using XMLDB and export it using XMLDB. Check for differences.

jrchamp (commented):
Thanks @paulandm! We'll take a closer look at this updated version during our next meeting. There are a few comments from last time, like the MySQL-specific SQL that will need to be addressed before it can be merged, but I'm excited to see the progress.

paulandm:
Thanks @jrchamp I've implemented some improvements related to DB compatibility and have simplified the keys of the table.

classes/task/send_ical_notifications.php
paulandm: @jrchamp Would it be a good idea if we maybe add a feature to the mod/zoom/view.php page to include a optional parameter like 'autojoin', so that the authenticated user is automatically forwarded to the zoom url?

classes/task/send_ical_notifications.php
```diff
@@ -0,0 +1,266 @@
...
+            $messagedata = new \core\message\message();
+            $messagedata->component = 'mod_zoom';
+            $messagedata->name = 'ical_notifications';
+            $messagedata->userfrom = \core_user::get_noreply_user();
+            $messagedata->userto = $user;
+            $messagedata->subject = $zoomevent->name;
+            $messagedata->fullmessage = $zoomevent->description;
+            $messagedata->fullmessageformat = FORMAT_HTML;
+            $messagedata->fullmessagehtml = $zoomevent->description;
+            $messagedata->smallmessage = $zoomevent->name . ' - ' . $zoomevent->description;
```
jrchamp, line 177: Note to tester: Do we need to apply filters like in #615 so the messages reflect the displayed values for the activities? Do the event names and descriptions already have the post-processed value or do they need to have the filter applied? There is a helper function for the activity name, but testing would be needed to determine if the description gets automatically passed through filters or not or if we need to supply the context.

classes/task/send_ical_notifications.php
```diff
@@ -0,0 +1,266 @@
...
+            if (!$serializedical || empty($serializedical)) {
+                mtrace('[Zoom ical Notifications] A problem occurred while trying to serialize the ical data for user ID ' .
+                        $user->id . ' for zoom event ID ' . $zoomevent->id);
+                continue;
+            }
+
+            $icalfileattachment = $filestorage->create_file_from_string($filerecord, $serializedical);
+
+            $messagedata = new \core\message\message();
+            $messagedata->component = 'mod_zoom';
```
jrchamp, line 169: Note to reviewer: Compare to core Moodle to see if this is the recommended pattern for initializing the object.

classes/task/send_ical_notifications.php
```diff
@@ -0,0 +1,266 @@
...
+
+            $ical = $this->create_ical_object($zoomevent, $zoom, $user);
+
+            $filerecord = [
+                'contextid' => \context_user::instance($user->id)->id,
+                'component' => 'user',
+                'filearea' => 'draft',
+                'itemid' => file_get_unused_draft_itemid(),
+                'filepath' => '/',
+                'filename' => clean_filename('icalexport.ics'),
```
jrchamp, line 156: Reviewer: Determine if these is a concern about unique filenames. Do/should these files get cleaned up after use?

classes/task/send_ical_notifications.php
```diff
@@ -0,0 +1,266 @@
...
+
+        $zoom = $DB->get_record('zoom', ['id' => $zoomevent->instance], 'id,registration,join_url,meeting_id,webinar');
+
+        $filestorage = get_file_storage();
+
+        foreach ($users as $user) {
+            // Check if user has "Disable notifications" set.
+            if ($user->emailstop) {
+                continue;
+            }
```
jrchamp, line 146: Reviewer: Is emailstop intended to stop *all* messages or only emails? Are we supposed to be doing this check, or is this supposed to be handled by the messaging system itself?

classes/task/send_ical_notifications.php
```diff
@@ -0,0 +1,266 @@
...
+            $executiontime = 0;
+        }
+        return $executiontime;
+    }
+
+    /**
+     * Set the execution time (the current time) for successful ical notifications sent for the related zoom event id.
+     * @param string $zoomeventid The zoom event id.
+     */
+    private function set_notification_executiontime(string $zoomeventid) {
```
jrchamp, line 117: Reviewer: I would fully expect this to be an integer. Maybe it depends on the database and PHP versions, but that would mean this will break on some systems. Probably best to cast as `(int)` before passing to this function and update the parameter types.

classes/task/send_ical_notifications.php
```diff
@@ -0,0 +1,266 @@
...
+            'tenminutesago' => time() - (60 * 10),
+        ]);
+    }
+
+    /**
+     * Get the execution time (last successful ical notifications sent) for the related zoom event id.
+     * @param string $zoomeventid The zoom event id.
+     * @return string The timestamp of the last execution.
+     */
+    private function get_notification_executiontime(string $zoomeventid) {
```
jrchamp, line 103: Reviewer: Same concern about parameter types (should probably cast to `(int)``)

classes/task/send_ical_notifications.php
armccoy: Yes. This is how it's done in core Moodle.

classes/task/send_ical_notifications.php
armccoy: Agreed!

classes/task/send_ical_notifications.php
jrchamp: Looks like file_get_unused_draft_itemid gives us a unique identifier, so that shouldn't be an issue. I don't know the internal notification system well enough to know what happens if the notification is sent as an internal message within Moodle (as opposed to an email) and thus if it is safe for us to clean up the draft file (or even if the draft files are auto-cleaned by Moodle).

classes/task/send_ical_notifications.php
armccoy: Yes, `emailstop` disables all notifications for the user. Probably better to do the check here instead of letting the messaging system itself.

classes/task/send_ical_notifications.php
jrchamp: To test the remaining concern (does the ics file still work for local messages? [email should be working because it leave Moodle]), set your notification settings to in-Moodle messages and make sure that the ics file still works even though the temp file gets unlinked.

classes/task/send_ical_notifications.php
jrchamp: Ok, this needs work then. 🔧

classes/task/send_ical_notifications.php
jrchamp: Needs work (filter event name similar to Zoom API filtering, see #615). 🔧

classes/task/send_ical_notifications.php
paulandm: @jrchamp Do you need me to have a look at this or are the reviewers looking at this?

classes/task/send_ical_notifications.php
```diff
@@ -0,0 +1,280 @@
...
+
+        $zoom = $DB->get_record('zoom', ['id' => $zoomevent->instance], 'id,registration,join_url,meeting_id,webinar');
+
+        $filestorage = get_file_storage();
+
+        // Apply filters to event name and description.
+        $cminfo = get_fast_modinfo((int)$zoomevent->courseid)->instances['zoom'][(int)$zoomevent->instance];
+        $formatoptions = [];
+        if ($cminfo && !empty($cminfo->id)) {
+            $formatoptions['context'] = \context_module::instance($cminfo->id);
```
jrchamp, line 147: Switch to `use context_module;` and remove backslash.

classes/task/send_ical_notifications.php
```diff
@@ -0,0 +1,280 @@
...
+        WHERE modulename = :zoommodulename
+        AND eventtype = :zoomeventtype
+        AND timemodified >= :onehourago
+        AND timemodified <= :tenminutesago';
+
+        return $DB->get_records_sql($sql, [
+            'zoommodulename' => 'zoom',
+            'zoomeventtype' => 'zoom',
+            'onehourago' => time() - (60 * 60),
+            'tenminutesago' => time() - (60 * 2),
```
jrchamp, line 97: This was changed from 10 minutes to 2 minutes. Was that intended?

classes/task/send_ical_notifications.php
```diff
@@ -0,0 +1,280 @@
...
+     *
+     * @return void nothing.
+     */
+    public function execute() {
+        if (get_config('zoom', 'sendicalnotifications')) {
+            mtrace('Zoom ical Notifications - Starting cron job.');
+            $zoomevents = $this->get_zoom_events_to_notify();
+            if ($zoomevents) {
+                foreach ($zoomevents as $zoomevent) {
+                    $notificationtime = $this->get_notification_time((int)$zoomevent->id);
```
jrchamp, line 58: We prefer a space after the type cast.
```suggestion
                    $notificationtime = $this->get_notification_time((int) $zoomevent->id);
```

There are probably a dozen of these in the PR, but I'll only comment this first one.

classes/task/send_ical_notifications.php
```diff
@@ -0,0 +1,280 @@
...
+        global $DB;
+
+        $users = $this->get_users_to_notify((int)$zoomevent->instance, (int)$zoomevent->courseid);
+
+        $zoom = $DB->get_record('zoom', ['id' => $zoomevent->instance], 'id,registration,join_url,meeting_id,webinar');
+
+        $filestorage = get_file_storage();
+
+        // Apply filters to event name and description.
+        $cminfo = get_fast_modinfo((int)$zoomevent->courseid)->instances['zoom'][(int)$zoomevent->instance];
```
jrchamp, line 144: Since all we use from this function is the Zoom activity's $cmid, I wish there was a simpler way to get that. Maybe get_fast_modinfo() is the best, but I don't know.

jrchamp (commented):
Thanks @paulandm! Yes, @armccoy is taking a look at the code and testing it. If there are issues, we'll let you know what we find. I had a few comments based on my quick scan of the code.

classes/task/send_ical_notifications.php
paulandm: Sorry, that I did that during testing and it sneaked in here. Will revert back to 10 minutes.

classes/task/send_ical_notifications.php
paulandm: Aah, good point. I suppose get_coursemodule_from_instance() makes more sense in this scenario.

classes/task/send_ical_notifications.php
jrchamp: Attachments don't show up in local messages, but the ICS file is already available in Moodle. So a link to the activity was added to the notification so that the local message is useful.

db/install.xml
armccoy: Done! Submitted a PR to Paul's fork.

jrchamp (approved):
Thank you! This new, optional feature will be in the next feature release.

PR #630, opened 2024-11-28
Protect group from license borrowing

### LINKED ISSUE #148: Redefine licenses only from specific Zoom group(s) or exclude specific group(s)
Hi

Is it possible to consider the "redefine licenses" option with only a particular group in Zoom? This would prevent certain users from losing their licenses at times like administrators etc. and the license would be taken only from the group members in Zoom.

Or inverse to be able to exclude specific group from "pool" so that they do not lose their license.

Thank you

izendegi:
On this PR we 
- Add a new setting to select the Zoom group to protect
- When recycling licenses, avoid removing license from users belonging to the protected group

Fixes #148


settings.php
```diff
@@ -171,6 +172,21 @@
         1
     );
     $settings->add($recycleonjoin);
+    
+    // Only call to the web services and load the setting if the connection is OK
+    if (isset($status)) {
+        if ($status == 'connectionok') {
+            $zoomgrps[] = null;
+            $groupobj = zoom_webservice()->get_groups();
```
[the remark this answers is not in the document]
izendegi: Hi, @smbader 

There is no API calling when $status is not set, and it's only set if the URL is [settings.php?section=modsettingzoom](https://github.com/ncstate-delta/moodle-mod_zoom/blob/5756eeb8839667c6c4eb14a358ef9b348dc23a42/settings.php#L39), so there is no that kind of risk in this piece of code.

settings.php
```diff
@@ -171,6 +172,21 @@
         1
     );
     $settings->add($recycleonjoin);
+    
+    // Only call to the web services and load the setting if the connection is OK
+    if (isset($status)) {
+        if ($status == 'connectionok') {
+            $zoomgrps[] = null;
```
jrchamp, line 179:
```suggestion
            $zoomgrps = [];
```

jrchamp (commented):
This is a nice idea to help avoid unwanted reassignment. A few suggestions to help minimize the negative impacts on users who are not using groups or reassignment.

settings.php
```diff
@@ -171,6 +172,21 @@
         1
     );
     $settings->add($recycleonjoin);
+    
+    // Only call to the web services and load the setting if the connection is OK
+    if (isset($status)) {
+        if ($status == 'connectionok') {
```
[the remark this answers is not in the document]
izendegi: Done!

settings.php
izendegi: Done!

classes/webservice.php
```diff
@@ -476,6 +484,15 @@ private function get_least_recently_active_paid_user_id() {
...
     }
 
+    /**
+     * Get a list of Zoom groups
+     *
+     * @return stdClass The call's result in JSON format.
+     */
+    public function get_groups() {
+        return $this->make_call('/groups');
+    }
```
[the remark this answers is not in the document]
izendegi: I've added both an scope grant control and a basic try-catch control

settings.php
```diff
@@ -171,6 +172,21 @@
...
+            $zoomgrps[] = null;
+            $groupobj = zoom_webservice()->get_groups();
+            foreach ($groupobj->groups as $group) {
+                $zoomgrps[$group->id] = $group->name;
+            }
+            $protectedgrp = new admin_setting_configselect('zoom/protectedgrp',
+                get_string('protectedgrp','mod_zoom'),
+                get_string('protectedgrp_desc','mod_zoom'),0,$zoomgrps);
```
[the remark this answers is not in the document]
izendegi: Done!

classes/webservice.php
```diff
@@ -461,7 +468,8 @@ private function get_least_recently_active_paid_user_id() {
         $userslist = $this->list_users();
 
         foreach ($userslist as $user) {
-            if ($user->type != ZOOM_USER_TYPE_BASIC && isset($user->last_login_time)) {
+            if ($user->type != ZOOM_USER_TYPE_BASIC && isset($user->last_login_time)
+                && !in_array($this->protectedgroup,$user->group_ids ?? array(),true)) { // needle: 'PROTECTED' zoom group
```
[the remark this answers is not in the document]
izendegi: I've broke it in several clauses, I think it's much more clear now.

lang/en/zoom.php
```diff
@@ -320,6 +320,8 @@
 $string['privacy:metadata:zoom_meeting_participants:user_email'] = 'The email of the participant';
 $string['privacy:metadata:zoom_meeting_view'] = 'The database table to track users that view the meeting recordings';
 $string['privacy:metadata:zoom_meeting_view:userid'] = 'The id of the user that viewed the recording';
+$string['protectedgrp'] = 'Protect group';
+$string['protectedgrp_desc'] = 'Select a Zoom group to protect member users from license redefining';
```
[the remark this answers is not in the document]
izendegi: Done!

settings.php
```diff
@@ -26,6 +26,7 @@
 
 require_once($CFG->dirroot . '/mod/zoom/locallib.php');
 require_once($CFG->libdir . '/environmentlib.php');
+require_once($CFG->dirroot . '/mod/zoom/classes/webservice.php');
```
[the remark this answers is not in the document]
izendegi: Done!

settings.php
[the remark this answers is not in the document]
izendegi: I've changed that setting to make it multiselect.

classes/webservice.php
```diff
@@ -476,6 +494,28 @@ private function get_least_recently_active_paid_user_id() {
...
+            try {
+                $response = $this->make_call('/groups');
+            } catch (moodle_exception $error) {
+                $response = '';
+            }
+        } else {
+            $response = '';
+        }
+
+        return $response;
```
jrchamp, line 516:
```suggestion
     * @return array Group information.
     */
    public function get_groups() {
        $groups = [];

        // Classic: group:read:admin.
        // Granular: group:read:list_groups:admin.
        // Not essential scope, execute only if scope has been granted.
        if ($this->has_scope(['group:read:list_groups:admin', 'group:read:admin'])) {
            try {
                $response = $this->make_call('/groups');
                $groups = $response->groups ?? [];
            } catch (moodle_exception $error) {
                // Only available for Paid accounts, so ignore error.
            }
        }

        return $groups;
```

settings.php
```diff
@@ -172,6 +172,25 @@
...
+                $zoomgrps[$group->id] = $group->name;
+            }
+        }
+
+        $protectedgroups = new admin_setting_configmultiselect(
+            'zoom/protectedgroups',
+            get_string('protectedgroups', 'mod_zoom'),
+            get_string('protectedgroups_desc', 'mod_zoom'),
+            [],
+            $zoomgrps);
```
jrchamp, line 190:
```suggestion
            $zoomgroups
        );
```

settings.php
```diff
@@ -172,6 +172,25 @@
...
 
+    // Only call to the web services and load the setting if the connection is OK.
+    if (isset($status) && $status === 'connectionok') {
+        $zoomgrps = [];
+        $groupobj = zoom_webservice()->get_groups();
+        if ($groupobj) {
+            foreach ($groupobj->groups as $group) {
+                $zoomgrps[$group->id] = $group->name;
+            }
+        }
```
jrchamp, line 183:
```suggestion
        $zoomgroups = [];
        $groups = zoom_webservice()->get_groups();
        foreach ($groups as $group) {
            $zoomgroups[$group->id] = $group->name;
        }
```

classes/webservice.php
```diff
@@ -461,12 +468,23 @@ private function get_least_recently_active_paid_user_id() {
...
+            }
+            // Skip the protected group.
+            if (!empty($this->protectedgroups) && array_intersect($this->protectedgroups, $user->group_ids ?? [])) {
+                continue;
+            }
+            // We need the login time.
+            if (!isset($user->last_login_time)) {
+                continue;
             }
+            // Count the user only if we're including all users or if the user is on this instance.
```
jrchamp, line 483: I recommend blank lines before comments and between code blocks for readability.
```suggestion
            // Skip Basic user accounts.
            if ($user->type == ZOOM_USER_TYPE_BASIC) {
                continue;
            }

            // Skip the protected group.
            if (!empty($this->protectedgroups) && array_intersect($this->protectedgroups, $user->group_ids ?? [])) {
                continue;
            }

            // We need the login time.
            if (!isset($user->last_login_time)) {
                continue;
            }

            // Count the user only if we're including all users or if the user is on this instance.
```

jrchamp (commented):
Thank you for all of your timely updates! This is very nice and should help users who needs this functionality. After applying the suggested changes, please make sure that it is still working as expected on your system (as I believe you may be one of the users of this functionality).

settings.php
izendegi: If applied a commit to change that var name in all the places

izendegi:
Hi, @jrchamp 

We've been using the original code for more than a year but haven't tested these last changes extensively (I'm on a leave for a couple of weeks), so please give me some weeks to test them and refine if needed.

jrchamp:
Hi @izendegi 

The code looks good to us, so we'll wait for your testing to make sure this version is working for your system. Thank you for your efforts on this!

izendegi:
Hi, @jrchamp 

I've already tested that the last changes work as expected and added a specific error to throw when there is no user to borrow the license from.

classes/webservice.php
```diff
@@ -766,7 +810,11 @@ public function provide_license($zoomuserid) {
             if ($this->paid_user_limit_reached()) {
                 $leastrecentlyactivepaiduserid = $this->get_least_recently_active_paid_user_id();
                 // Changes least_recently_active_user to a basic user so we can use their license.
-                $this->make_call("users/$leastrecentlyactivepaiduserid", ['type' => ZOOM_USER_TYPE_BASIC], 'patch');
+                if ($leastrecentlyactivepaiduserid) {
+                    $this->make_call("users/$leastrecentlyactivepaiduserid", ['type' => ZOOM_USER_TYPE_BASIC], 'patch');
+                } else {
+                    throw new moodle_exception('errornousersfound', 'mod_zoom');
+                }
```
jrchamp, line 817: I agree that it is good to check to match sure that we're making an API call with a good value. However, it doesn't look like this moodle_exception is being caught and it looks like that will prevent the meeting host from being able to join their meeting if there are no licenses available. Where I don't love the idea of failing silently, I do worry about preventing them (and only them) from being able to join the meeting.

izendegi: I added that exception in the first place because when that situation happens (there is no user to borrow the license from) another  exception was already thrown when trying to get the PRO license and there isn't any available.

I agree that throwing an exception is not the best thing to do, so instead of that, I've added the control to try to get the PRO license only if there limit is not reached or if it has been removed from another user, does it make sense to you? Maybe adding a message when this happens could be a good thing to do...

izendegi:
I've also rebased this branch from upstream (v5.2.5)

jrchamp (approved):
This looks good. New feature, so it will be part of a feature release.

lcollong:
Great to have ! :)

PR #631, opened 2024-12-23
ci: add MOODLE_405_STABLE; drop MOODLE_402_STABLE

jrchamp:


smbader (approved):
We are keeping older versions to continue testing related, older PHP versions - correct?

PR #632, opened 2025-01-08
services: remove deprecated local_mobile

jrchamp:


smbader (approved):
3.5 was so long ago....

PR #637, opened 2025-01-31
passcode is encoded via "htmlspecialchars" function Closes #636

### LINKED ISSUE #636: Recording's passcode display bug
Hello everybody!

I think I found a small bug on the displaying of the recording's passcode, when the password generated by zoom to access the recording unluckily contains a sequence of characters which may be sometimes misinterpreted as an HTML character.

In our case, the passcode contained the sequence `&#` which resulted in an erroneous representation of the passcode. This made the recording unavailable to users, as they couldn't see the correct passcode (some OSs display a square, on my system - ubuntu - I could see only the alphanumeric characters of this passcode).

FrancescoTerrosi:
This bugfix Closes #636 by encoding html special characters to display the correct passcode for recordings, in the unlucky case the passcode generated by zoom contains a sequence of characters that may be misinterpreted as html characters

### DIFF
--- a/recordings.php
+++ b/recordings.php
@@ -132,7 +132,7 @@ if (empty($recordings)) {
         }
 
         // Output only one row per grouping.
-        $table->data[] = [$recordingdate, $recordinghtml, $recordingpasscode, $recordingshowhtml];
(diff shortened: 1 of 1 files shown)


jrchamp (approved):
Good find! Thank you for identifying and fixing this issue. 🥂

PR #639, opened 2025-02-07
errors: show more detailed messages

### LINKED ISSUE #627: Show more detailed error messages (when trying to add zoom meeting activity)
can any one help to explain this error please

image

karenliulll:
This PR enhances not_found_exception and shows more detailed messages (the solution is consistent with bad_request_exception). It also ensures that a "user not found" error will be thrown if the user does not have a Zoom account (e.g.: the user is trying to recreate a meeting in Moodle, but they do not have a Zoom account).


locallib.php
```diff
@@ -400,6 +400,10 @@ function zoom_get_user_id($required = true) {
                 $zoomuserid = $zoomuser->id;
                 $cache->set($USER->id, $zoomuserid);
             }
+            // If user does not have a Zoom account, throw an error.
+            if ($required && $zoomuser === false) {
+                throw new moodle_exception('zoomerr_usernotfound', 'mod_zoom');
```
jrchamp, line 405: The `zoomerr_usernotfound` string expects a value to be provided for the link. In the other location where it is used, the link provided is `$config->zoomurl`. In this context, the config variable is not yet available, so perhaps:
```suggestion
                throw new moodle_exception('zoomerr_usernotfound', 'mod_zoom', '', get_config('zoom', 'zoomurl'));
```

jrchamp (commented):
This looks great! Can you show an example of the new error message in action?

karenliulll:
@jrchamp I wasn't able to replicate the scenario in the issue, but I have a dummy example that produces a very similar error message if that's ok - in mod_form.php line 660, if you make the $hostuserid = false, and then go to the settings of any Zoom meeting in Moodle, it should show an error message like "The resource does not exist: User does not exist".
image

However, it does look like the new change of throwing the usernotfound error will cover most cases where $zoomuserid = false, so the not_found_exception should show up less frequently.

locallib.php
```diff
@@ -400,6 +400,10 @@ function zoom_get_user_id($required = true) {
                 $zoomuserid = $zoomuser->id;
                 $cache->set($USER->id, $zoomuserid);
             }
+            // If user does not have a Zoom account, throw an error.
+            if ($required && $zoomuser === false) {
+                throw new moodle_exception('zoomerr_usernotfound', 'mod_zoom', '', get_config('zoom', 'zoomurl'));
+            }
```
jrchamp, line 406: This is the correct change, but I think it highlights that some cases where `zoom_get_user_id()` is called are not prepared for this to fail.

This instance should probably be using `$hostuserid` (which would need to be moved before and outside of where it is currently initialized)
https://github.com/ncstate-delta/moodle-mod_zoom/blob/66b41e4e7dd1ff49e9b898a2d5c6dd2a1846c403/mod_form.php#L797

This should probably be using the Zoom meeting's host_id, not the current user's settings.
https://github.com/ncstate-delta/moodle-mod_zoom/blob/66b41e4e7dd1ff49e9b898a2d5c6dd2a1846c403/classes/webservice.php#L721

jrchamp (commented):
Overall, this looks great! Some pre-existing code might need changes to work correctly.

jrchamp (approved):
This looks great! When integration tests happen, we'll want to check both the positive and negative tests for:

- Creating a new activity
- Editing an existing activity (can use the newly created activity from the previous step)
- Recreate an expired activity

I'm very interested to know what the error messages look like now for those three cases.

PR #642, opened 2025-02-27
Zoom/scopes check

### LINKED ISSUE #621: Check a task's necessary scopes prior to API calls
From #614, we should check the necessary scopes for a task before we make API calls that will fail (especially if it's going to fail silently). This will allow us to use `mtrace()` or similar to provide information to the Moodle server administrators if they are trying to figure out why a task is not working as expected. Because recording/reports/etc scopes are technically optional, this will allow us to let them know if a task is running that has not been granted the scopes for that task to function/provide value.

cbounphengsy:
Updates to check scopes and missing scopes

Fixes #621

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ Optional functionality can be enabled by granting additional scopes:
     - cloud_recording:read:list_user_recordings:admin
     - cloud_recording:read:recording_settings:admin
 - Tracking fields (zoom | defaulttrackingfields)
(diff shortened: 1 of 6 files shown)


jrchamp (commented):
This is looking good! Just a few edits and I think it'll be ready to merge. :partying_face:

smbader (approved):
Good work!

PR #645, opened 2025-03-06
Populate calendar with only items that were not deleted

### LINKED ISSUE #641: Remove calendar entries
We schedule irregular meetings (always the same day and time but not every week). We can do this by setting the schedule as weekly and then deleting the instances we don't want within Zoom. However the original weekly meetings still show up in the Moodle calendar and we can't delete them. 
It would be great if there could be more control over the calendar entries - even just the option to turn off the Moodle calendar events being added would be helpful.

cbounphengsy:
- Updated zoom_calendar_item_update() to check if the occurrence has been "deleted"
- If occurrence status is "deleted", do not add to calendar

Fixes #641

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -512,7 +512,7 @@ function zoom_calendar_item_update(stdClass $zoom) {
     } else if (!empty($zoom->occurrences)) {
         foreach ($zoom->occurrences as $occurrence) {
             $uuid = $occurrence->occurrence_id;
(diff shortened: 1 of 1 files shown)


jrchamp (commented):
Excellent find! I did not expect that Zoom would have a "status => deleted" for deleted occurrences.

jrchamp (approved):

PR #646, opened 2025-03-06
release: v5.3.0

jrchamp:
A new feature for people who use automatic license management to select protected groups (so those users are not automatically managed). Also, a lot of nice user experience bug fixes.


karenliulll (approved):
Looks good!

PR #655, opened 2025-05-01
Update webservice.php generalized autocreate function

yedidiaklein:
I know that the autocreate function is deprecated, but I wrote a plugin for auto creating users here:
https://github.com/yedidiaklein/moodle-local_zoomsyncusers

I use the zoom API from your plugin and I depend on it.

The existing autocreate function is broken and should use action autoCreate and not autocreate.

While fixing it, I also generilized this function to work for create and autoCreate and also to choose if needed the user type.

For backward compatibility I gave default values to the new parameters.

It would be great if you accept this minor PR.

Thanks a lot (and also tnx for this great plugin, that is working very well and is very usefull !)

### DIFF
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -388,7 +388,7 @@ class webservice {
      * @param string $action The account create action: create, autoCreate, custCreate or ssoCreate.
      * @param int $type The user type number.
      * @return bool Whether the user was succesfully created.
-     * @deprecated Has never been used by internal code.
+     * @example https://github.com/yedidiaklein/moodle-local_zoomsyncusers An external plugin that depends on mod_zoom uses this method.
      */
     public function autocreate_user($user, $action = 'autoCreate', $type = ZOOM_USER_TYPE_PRO) {
         // Classic: user:write:admin.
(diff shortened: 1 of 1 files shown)


jrchamp:
Thank you @yedidiaklein! Although we don't use this function in our code, it make sense for it to function as long as we continue to include it. For reference, here is the current API documentation: https://developers.zoom.us/docs/api/users/#tag/users/POST/users

classes/webservice.php
```diff
@@ -385,17 +385,19 @@ private function make_paginated_call($url, $data, $datatoget) {
      * Autocreate a user on Zoom.
      *
      * @param stdClass $user The user to create.
+     * @param string $action The account create action: create, autoCreate, custCreate or ssoCreate.
+     * @param int $type The user type number.
      * @return bool Whether the user was succesfully created.
      * @deprecated Has never been used by internal code.
```
jrchamp, line 391:
```suggestion
     * @example https://github.com/yedidiaklein/moodle-local_zoomsyncusers An external plugin that depends on mod_zoom uses this method.
```

jrchamp (approved):
Thank you! This is a nice fix and flexibility enhancement that may help many people.

yedidiaklein:
Great !
Thank you !

PR #656, opened 2025-05-15
Updated catch to throw error instead of returning false

### LINKED ISSUE #500: Show error details when Changes to the activity Zoom Meeting are not applied
Hi!
In some moodle courses we use the plugin to a great extent. To do this, we create a meeting and then duplicate the activity multiple times, as this generates new meetings with their own meeting ID. Then we change the title, date and time and the prerequisites.

Now the error has occurred that changes are no longer applied and the error message "Zoom cannot be updated". or "error/moodle/cannotupdatemod" appears. Sometimes the changes are then still applied in Moodle, but never in Zoom. This does not only affect changes to the title or date, but also appears when prerequisites are to be changed.
We have observed that the error always seems to occur after editing 40-50 meetings. The next day, changes could be made again and the error occurred again after about 45 processed meetings. It looks like an overload. Can anyone help us further?

Thank you and best regards

Chatharina

cbounphengsy:
Updated catch to throw error to display specific error details

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -196,7 +196,11 @@ function zoom_update_instance(stdClass $zoom, ?mod_zoom_mod_form $mform = null) 
             }
         }
     } catch (moodle_exception $error) {
-        throw $error;
+        if (isset($mform)) {
+            throw $error;
+        } else {
+            return false;
+        }
     }
 
     // Get the updated meeting info from zoom, before updating calendar events.
--- a/lib.php
+++ b/lib.php
@@ -196,7 +196,7 @@ function zoom_update_instance(stdClass $zoom, ?mod_zoom_mod_form $mform = null) 
             }
         }
     } catch (moodle_exception $error) {
-        return false;
+        throw $error;
     }
 
     // Get the updated meeting info from zoom, before updating calendar events.


jrchamp:
Thank you @cbounphengsy ! I added the try/catch via #62 to avoid problems with the OU Dates report. However, that prevents us from being able to provide specific exception information. Based on our discussion today, we have a few options:

1. Only undo the try/catch change (as this PR currently proposes). This is appropriate for update_moduleinfo() which throws an exception anyway if we return false. This will prevent someone from using the OU Dates report if any of their Zoom activities fail to update for any reason.
2. Create a PR to change the OU Dates report to handle exceptions and undo the try/catch after it is merged. This should work well, but has an external dependency which may delay providing value to our users.
3. Modify this change to only throw when `isset($mform)`. This would throw the exception when people are using the modedit form interface, but not affect other code paths (such as OU Dates).

For the moment, I think option 3 is the safest and fastest way to deliver value. I think that option 2 is the ideal state, because the OU Dates report's current approach of "one transaction for all changes" seems like it doesn't allow people to make partial progress on updating things. I would suggest to modify the OU Dates tool such that only changes which fail are rolled back. This would allow for the more useful error messages that our plugin is attempting to provide to be included in the Dates report when an issue occurs (it would also work for other plugins that may be throwing exceptions).

lib.php
```diff
@@ -196,7 +196,7 @@ function zoom_update_instance(stdClass $zoom, ?mod_zoom_mod_form $mform = null)
             }
         }
     } catch (moodle_exception $error) {
-        return false;
+        throw $error;
```
jrchamp, line 199:
```suggestion
        if (isset($mform)) {
            throw $error;
        } else {
            return false;
        }
```

jrchamp (approved):

PR #664, opened 2025-07-04
Fix upgrade.php update SQL

### LINKED ISSUE #665: Upgrade to 2025062600+ is not compatible with Postgres
The SQL syntax used in the upgrade.php for the 2025062600 version is MySQL specific and results in the following error on Postgres:

```
-->mod_zoom
!!! Error writing to database !!!
!! ERROR:  syntax error at or near "JOIN"
LINE 2:                   JOIN mdl_tool_customlang_components tcc ON...
                          ^
UPDATE mdl_tool_customlang tc
                  JOIN mdl_tool_customlang_components tcc ON tc.componentid = tcc.id
                   SET tc.stringid = $1
                 WHERE tcc.name = $2
                   AND tc.stringid = $3
[array (
  0 =&gt; 'calendardescriptionurl',
  1 =&gt; 'mod_zoom',
  2 =&gt; 'calendardescriptionURL',
)]
Error code: dmlwriteexception !!
```

izendegi:
Using a JOIN clause inside an UPDATE is not standard SQL and breaks updating the plugin using PostgreSQL. This PR refactors the query to make it compatible with all supported DBMSs

### DIFF
--- a/db/upgrade.php
+++ b/db/upgrade.php
@@ -1029,7 +1029,7 @@ function xmldb_zoom_upgrade($oldversion) {
         $sql = "UPDATE {tool_customlang} tc
                    SET stringid = :newstringid
                  WHERE tc.stringid = :oldstringid
-                   AND tc.componentid in (SELECT tcc.id 
+                   AND tc.componentid in (SELECT tcc.id
                                             FROM {tool_customlang_components} tcc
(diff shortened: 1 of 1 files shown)


jrchamp (approved):
Confirmed that this works on MySQL. @izendegi has indicated that this works on Postgres, so approved!

PR #669, opened 2025-07-21
release: v5.4.3

jrchamp:
Supersedes v5.4.1 and v5.4.2, solving the stringid case-insensitivity issue by using a different stringid altogether. The lesson learned was "Custom SQL with differently configured databases is something to avoid".


armccoy (approved):
Looks good!

PR #671, opened 2025-07-29
SECURITY: Hardened shell command execution in get_meeting_report.php

OshratLuski:
- Used escapeshellarg() for all string parameters passed to CLI script.
- Ensured only sanitized parameters are passed to the shell command.
- Prevented potential command injection.

### DIFF
--- a/console/get_meeting_report.php
+++ b/console/get_meeting_report.php
@@ -30,8 +30,8 @@ error_reporting(E_ALL);
 ini_set('display_errors', '1');
 
 $courseid = required_param('courseid', PARAM_INT);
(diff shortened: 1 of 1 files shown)


smbader:
Hi Oshrat,

Thank you for contributing, and I understand the need for escaping params before execution. I'm curious why you retracted your change, if it was for the same reason I'm struggling with. I think `optional_param` with type `PARAM_ALPHANUMEXT` is doing more to sanitize params than `escapeshellarg`.  

escapeshellarg adds single quotes around a string and quotes/escapes any existing single quotes.
while...
optional_param with type PARAM_ALPHANUMEXT: Removes everything not `a-zA-Z0-9_-`.
` return preg_replace('/[^A-Za-z0-9_-]/i', '', $param);`

Would love your feedback,
Steve

PR #672, opened 2025-08-07
Release/v5.4.4

jrchamp:
A small fix to add additional protection, just in case. My testing indicated that the current protection mechanisms are doing a lot, but this increases that level of confidence.


smbader (approved):
Future proofed!

PR #673, opened 2025-08-10
fix:  Use new external class names (requires Moodle 4.2)

hamzakhalem:


jrchamp:
Hi @hamzakhalem! Thank you for your contribution. The classes were renamed and class aliases were added in Moodle 4.2.

https://github.com/moodle/moodle/blob/621c6e1048892929cb2f1180a3e8b46b053bf41d/public/lib/externallib.php#L37-L48

The class aliases are still present in Moodle 4.5, 5.0 and - at the moment - the main branch. At this time, we support Moodle 3.7 and newer, so merging this change would break support for Moodle 3.7 through Moodle 4.1. That is okay if - and only if - we are increasing our minimum supported version and increasing our version number to indicate backward incompatibility.

So, I will tag this as part of the mod_zoom "next major release" milestone for when it makes sense to drop support for older versions of Moodle.

smbader:
This patch requires us to drop everything prior to 4.2.  4.1 was an LTS version and just ended roughly 6 months ago.
According to https://stats.moodle.org/ that looks to be about 1/3 of the current Moodle installs.  We may need to wait a little longer on this one.

PR #677, opened 2025-09-18
fix: lowercase alternative host email addresses for case insensitive comparisons

### LINKED ISSUE #618: Case sensitivity on Alternative Hosts field - is it possible to make case insesntive
Hi All,

Thanks for this plugin, we've used it to great success however have noticed that we have an issue where some of our users have capital letters in their email addresses.

Scenario -
User email address: Jo.Blogs@invalid.com
Email address matches in Zoom and Moodle using the specific format above.

Our LMS team adds Jo as an alternative host on a call using the email address "jo.blogs@invalid.com"
When Jo tries to access the meeting, they will get the blue "Join Meeting" button and be unable to start the call. 
If Jo attempts to start the meeting they'll then get an error on the Zoom side saying "You have a meeting that is currently in-progress. Please end it to start a new meeting"
Screenshot 2024-09-11 at 2 58 27 PM

Our LMS team then updates the alternative host to use the correct format "Jo.Blogs@invalid.com" as an alternative host.
When Jo tries to access the meeting, they will get the green "Start Meeting" button and can start the call successfully.

----

For the moment we've asked the LMS team to ensure that they use the correct format for the users, however, they are not always aware of what the exact format should be. Is it possible to make this lookup case insensitive? I'm not sure if this is a problem on Zoom's side or the plugin itself

karenliulll:
Fixes #618.

### DIFF
--- a/locallib.php
+++ b/locallib.php
@@ -599,7 +599,7 @@ function zoom_get_selectable_alternative_hosts_list(context $context) {
         $zoomuser = zoom_get_user($u->email);
         if ($zoomuser !== false && $zoomuser->status === 'active') {
             // Add user to array of users.
-            $selectablealternativehosts[$u->email] = fullname($u);
+            $selectablealternativehosts[strtolower($u->email)] = fullname($u);
         }
     }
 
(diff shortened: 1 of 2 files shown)


jrchamp (commented):
A couple small suggestions to reduce the amount of code duplication.

jrchamp (approved):
Thanks @karenliulll ! This looks great! Definitely a significant improvement over the current behavior, so even if we missed something, this should address 90+% of the needs.

PR #681, opened 2025-10-16
Fix/recording query

jrchamp:
It seems like the recording query is not specific enough, so adding the zoomid to make sure that we're looking up the correct recordings is a good thing to verify.


armccoy (approved):
Looks good.

PR #687, opened 2025-11-26
tests: add generator for Behat tests

armccoy:
- The edits to `classes/webservice.php` and `lib.php` serve to allow Behat tests to circumvent the Zoom webservice check in the same way PHPUnit tests do.
- The file `tests/generator/behat_mod_zoom_generator.php` isn't used in the basic Behat test included (`tests/behat/view_meeting.feature`) but it's included in the event future tests may require it (probably with some additional edits).
- Behat tests for Zoom activity edit page interaction are currently failing, so some additional work to be able to include those related scenarios is needed.

### DIFF
--- a/tests/behat/behat_mod_zoom.php
+++ b/tests/behat/behat_mod_zoom.php
@@ -1,0 +1,96 @@
+<?php
+// This file is part of Moodle - http://moodle.org/
+//
+// Moodle is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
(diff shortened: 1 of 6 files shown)


tests/behat/behat_mod_zoom.php
```diff
@@ -0,0 +1,96 @@
...
+//
+// Moodle is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Moodle.  If not, see <http://www.gnu.org/licenses/>.
+
+require_once(__DIR__ . '/../../../../lib/behat/behat_base.php');
```
jrchamp: Is this require line how other behat tests are written? I would have guessed that Moodle would prefer to use class auto loading or a global/constant to avoid rigid folder navigation (particularly given compatibility with the recent public folder change).

jrchamp (commented):
These suggestions should at least address the syntax issues that are breaking the unit tests. I did not review the functional parts of the test code, because I only have a cell phone at this moment.

jrchamp (commented):
These suggestions might fix the CI errors.

smbader (approved):
We can now create Behat tests for further testing!

PR #697, opened 2026-01-29
Updated view.php to check if there is a role switch

### LINKED ISSUE #191: Bug: Respect switched role on activity overview page
Steps to reproduce:
* Login as teacher
* Go to a course
* Create a Zoom activity instance
* Switch to Student role with the Moodle role switcher in the user menu
* Look at the Zoom activity instance overview page (/mod/zoom/view.php?id=...)

Expected result:
* The information given on the Zoom activity instance overview page is the same as if you would log in with a real student user.
* Especially, you see a "Join meeting" button
* Especially, you don't see the passcode row
* Especially, you don't see the join link row

Actual result:
* The information given on the Zoom activity instance overview page is the same as you have seen before switching the role.
* Especially, you see a "Start meeting" button
* Especially, you see the passcode row
* Especially, you see the join link row

As this can disturb teachers, I would propose to fix this flaw.

cbounphengsy:
Added a check to see if there is a role switch. If so, assume the new role is not the original host.

### DIFF
--- a/lang/en/zoom.php
+++ b/lang/en/zoom.php
@@ -416,7 +416,6 @@ $string['showsecurity'] = 'Show Security section';
 $string['showsecurity_help'] = 'Enabling this option will show the Security section on the meeting activity page.';
 $string['showsecurityonview'] = 'Show Security section on meeting page';
 $string['start'] = 'Start';
-$string['start_meeting'] = 'Start Meeting';
 $string['start_time'] = 'When';
 $string['starthostjoins'] = 'Start video when host joins';
 $string['startpartjoins'] = 'Start video when participant joins';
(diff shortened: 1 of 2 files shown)


jrchamp (commented):
Hi! I was inspired by this PR - the targeted nature of the change helps significantly focus the concern. Thank you for working on this.

jrchamp (commented):
As I mentioned in #191, I think that switching a lot of the `if ($userishost)` checks to use `if ($iszoommanager)` instead is a better way to fix the role switch problem because it leverages the underlying capability changes that form the basis of the Moodle permission system.

smbader (approved):

PR #699, opened 2026-02-12
Added setting for pre-assigned breakout rooms

### LINKED ISSUE #696: FR: Admin-Option to disable pre-assigned Breakout-Rooms
This is not a bug report, but a feature request. As an administrator of a Moodle platform, I would like the ability to disable the option for pre-assigning Breakout Rooms.

Why?
1. Lack of Auto-Updates: If Moodle groups change, pre-assigned groups do not update automatically. The meeting must be edited and saved again manually, which often causes confusion.
2. Email Identifier Conflicts: Zoom uses email addresses as the sole identifier for pre-assignment. This workflow fails if users do not use the same email address for both Zoom and Moodle (which is the default state in our system).
3. Sync Issues: Pre-assignment fails if a user changes their email address on one platform but not the other.
4. Guest Access: Pre-assignment does not work if users join as guests in zoom without logging into Zoom first—a common practice for many meetings.
5. Several Zoom Accounts: Pre-assignment does not work if users are logged in into a different zoom account at the moment the join the meeting.

While the pre-assignment feature is useful when understood and used correctly, it is difficult to instruct every teacher and student on a large-scale moodle platform. To avoid support requests and misunderstandings, we would like the option to disable it.

Requested Solution: It would be very helpful to have a global setting in the plugin configuration to disable this feature platform-wide.

Additional Suggestions:
- An option to disable or enable it for specific categories.
- The introduction of a specific user permission: "Allow user to pre-assign breakout rooms."
- Allow pre-assignment of breakout rooms only if a meeting requires authentication.
- Inform users about necessary conditions for a working pre-assigment at the moment they try to configure it.

cbounphengsy:
Added site admin setting for pre-assigned breakout room (default is yes)
- Added strings for setting
- Hide "Breakout rooms" section within the meeting settings
- Updated api call to return "breakout" = true/false based on admin setting


classes/webservice.php
```diff
@@ -790,10 +790,14 @@ private function database_to_api($zoom, $cmid) {
...
 
-        if (isset($zoom->breakoutrooms)) {
-            $breakoutroom = ['enable' => true, 'rooms' => $zoom->breakoutrooms];
-            $data['settings']['breakout_room'] = $breakoutroom;
+        if ($config->defaultbreakoutroom) {
+            if (isset($zoom->breakoutrooms)) {
+                $breakoutroom = ['enable' => true, 'rooms' => $zoom->breakoutrooms];
+            }
+        } else {
+            $breakoutroom = ['enable' => false];
```
jrchamp: In my comment in settings.php, I encouraged renaming the setting to `preassignbreakoutrooms`. I think that this change in semantics makes it clear that this code should never be setting `enable` to `false`.

My reasoning: The setting is not about whether or not the teacher can use breakout rooms at all in the meeting. The only reason that we set `enable` to `true` for preassigned breakout rooms is that the host's default settings might not have breakout rooms enabled. So we should not set `enable` to `false` unless and until there is a setting that is specifically for turning the breakout room functionality on or off.

jrchamp (commented):
This will be a nice feature for people who do not want the complexity of preassigned breakout rooms. I had a few comments that I think will help.

I miss y'all! ❤️

jrchamp (commented):
A couple renames to match the settings.php file and a performance suggestion. Hope this helps!

jrchamp (commented):
This is looking great! Only a couple small suggestions and I think this is merge-ready.

jrchamp (commented):
Code looks great and response from testing is here: https://github.com/ncstate-delta/moodle-mod_zoom/issues/696#issuecomment-3899297445

It does occur to me that this could change the experience for people who are directly editing in Zoom to preassign breakout rooms, but that conflicts with the philosophy that these activities are created in Moodle, should be managed in Moodle and should honor the settings set in Moodle by the administrators, so this is also philosophically aligned.

smbader (approved):

PR #700, opened 2026-02-12
lib: call grade reset when zoom reset userdata is checked

### LINKED ISSUE #693: Reset might not reset grades
In #370, `zoom_reset_gradebook()` was added to `lib.php`. I assumed this was another situation where Moodle had defined a global function name that needed to be defined in that file. However, it seems like other modules that have a similar function call it directly themselves within a conditional: `if (empty($data->reset_gradebook_grades)) {`

To me, this means that there is no component callback for this method, and it is instead an unused function at this time.

smbader:
There are two options in the course reset that we need to support. 

The first is when the grade reset for the course is checked.  This should be handled already with our existing `zoom_reset_gradebook` function.

The second is when zoom specific reset option is checked which reads, _"Delete all user grades, recording user view data, and meeting participant user data."_ In this case, we were not reseting grades.  So this pull requests adds `zoom_reset_gradebook` to that function.  Since it's a nuclear checkbox that does all, we don't need to check for `$data->reset_gradebook_grades`

Fixes #693

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -873,6 +873,9 @@ function zoom_reset_userdata($data) {
             'item' => get_string('meetingrecordingviewsdeleted', 'zoom'),
             'error' => false,
         ];
+
+        // The Zoom reset checkbox resets all user grades, always.
+        zoom_reset_gradebook($data->courseid);
     }
 
     return $status;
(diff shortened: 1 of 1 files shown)


smbader:
@sgrandh3 will test one final time and if all is well, please feel free to hit the approve button.

sgrandh3:
Checked the "Delete all user grades, recording user view data, and meeting participant user data." checkbox in the course reset . Grades were cleared as expected.

armccoy (approved):
Looks good!

PR #701, opened 2026-02-16
js: Use ESM format for form.js

### LINKED ISSUE #571: Cleanup: AMD modules should be in ESM format
From https://moodledev.io/docs/apis/commonfiles#amd

> Although the AMD module format is supported, all new JavaScript is written in the EcmaScript Module (ESM) format.

If I'm correct, only the `amd/src/form.js` file would need to change.

jrchamp:
Switching to ESM is primarily:

- Switch from a `define` line to multiple `import` lines.
- Items that were being `return`ed, need to be `export`ed.
- `var` becomes `const` unless the value is modifiable, in which case it becomes `let`
- `function()` becomes `() =>`; similarly `function(arg1)` becomes `(arg1) =>`
- Avoid `this`, either by creating a real class or using parameters from the callable.

Other compatibility changes that I made:

- Convert events like `click(` to `on('click', `.
- Use `catch()` instead of `fail()` so we are prepared for real Promises.
- Aliased `get_string`/`get_strings` to `getString`/`getStrings` so we continue to support Moodle <4.3.
  - When support for Moodle 4.2 is dropped, we can use the real `getString`/`getStrings`.

Bugs that testing identified and that are fixed in this PR:

1. Breakout room tabs were not working in Moodle 5.x (Bootstrap 5).
   1. Add two rooms.
   2. Click on the first room tab.
   3. Expected: The first room details should display. Actual: The second room details are still visible and the browser has tried to navigate to `#link2`.
2. Repeat options were not unhiding when switching from a more-restricted list to a less-restricted list.
   1. Choose recurring.
   2. Select monthly, notice that the limit is set to 3.
   3. Select weekly.
   4. Expected: The limit should be set to 12. Actual: The limit is still set to 3.

Fixes #571


smbader:
This contains some nice clean up work and includes the new `data-bs-toggle` used in Bootstrap 5 which is used in Moodle 5.x. 

Testing needs to validate this continues to work on older versions and plays nicely with our breakout room feature request that is also being tested.

sgrandh3:
Hey @jrchamp , can you kindly eloborate " more-restricted list to a less-restricted list" in your description of the bugs. When creating a recurring zoom meeting, i do not see any options for the restrictions list.

jrchamp:
> Hey @jrchamp , can you kindly eloborate " more-restricted list to a less-restricted list" in your description of the bugs. When creating a recurring zoom meeting, i do not see any options for the restrictions list.

Specifically, when Monthly is chosen, the repeat interval only allows options 1, 2 and 3. After choosing Monthly, choosing Weekly should show all options between 1 and 12. The bug is that options 4 through 12 stay hidden, which this PR fixes.

The underlying reason for the bug is that the code that modifies which options are visible is written in a way where the Monthly and Weekly choices only hide options and never show them. As the options available for Monthly are a subset of the options for Weekly, this problem is only apparent when switching from Monthly to Weekly. The term "restrictions" is causing unnecessary confusion in this context: I am only referring to the range of options that are allowed for the repetition intervals.

smbader (approved):

sgrandh3:
After purging the cache, retested for Daily, weekly , monthly and no fixed time.  Verified that in all meetings, the upcoming meeting is displayed. As a future enhancement, it would be nice to have all the recurring sessions also displayed .

PR #398, opened 2022-08-12
mustache: template subdirectories requires Moodle 3.8

jrchamp:
Fixes #397

PR #604, opened 2024-07-24
Mobile app: Remove ionic3 template

jrchamp:
Drops support for old apps, so this will be merged whenever it is time for v6.

Commit was authored by @dpalou (so remember to thank!)

PR #702, opened 2026-02-16
ci: use recommended mobileapp template folder (requires Moodle 3.8)

jrchamp:
In moodle-plugin-ci 4.5.10, logic was added to treat the `templates/mobileapp` folder as special. Initially, it means the templates are automatically skipped, but in the future there may be actually relevant checks performed on these mobile templates.

Warning: Please note that Moodle added support for template folders in Moodle 3.8 (which we discovered previously in #398). Currently, this plugin only requires Moodle 3.7, so this would break compatibility and require an increase to the minimum supported version.


smbader:
I remember when we talked about these templates and why they were ignored.  I'm so happy they will have a home now to be ignored without the special `MUSTACHE_IGNORE_NAMES`.

jrchamp:
Note to self: It would be cleaner to fold #604 into this PR. They are both low-impact, backward-incompatible changes and modify some of the same lines, so let's not create merge conflicts for the maintainer. Even though the other change is older, this is the more-beneficial of the two and has the clearest requirements change.

smbader:
This patch requires us to drop everything prior to 3.8. Moodle version 3.9 ended in December 2023 and I can see justifying dropping support for previous versions.

PR #705, opened 2026-03-18
webservice: remove debug message from admin settings page

### LINKED ISSUE #691: Debug messages on settings page.
Hey mod_zoom development team,
We at Arden have found a issue where debug messages are displayed on the settings page. 

# The issue: 
When the user goes to the settings page `/admin/settings.php?section=modsettingzoom`. 
Debug information appears at the top of the page.

Image

Digging into the issue it seems when `has_scope` is called in `webservice.php`, `mtrace` is also executed. Since `mtrace` outputs to `stdout` by default, it writes to the output buffer - in this case, the webpage. Therefore, the `mtrace` call inside `has_scope` needs to be wrapped in an additional debug check so that it does not display on the webpage in a production environment.

Image

## Environment 
Moodle 4.5
PHP 8.3

karenliulll:
Fixes #691 

The CLI_SCRIPT check remains true when running a task manually through the admin interface. You can see the has_scope check in the output.

Example below: the Get meeting report from Zoom task (\mod_zoom\task\get_meeting_reports) calls the has_scope check. I ran it in my test instance and here is part of the output.
image

### DIFF
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -1242,7 +1242,9 @@ class webservice {
             $this->get_access_token();
         }
 
-        mtrace('checking has_scope(' . implode(' || ', $scopes) . ')');
+        if (CLI_SCRIPT) {
+            mtrace('checking has_scope(' . implode(' || ', $scopes) . ')');
+        }
 
         $matchingscopes = \array_intersect($scopes, $this->scopes);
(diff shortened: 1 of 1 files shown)


smbader (approved):
Thank you for checking that it displays in the "console window" on the schedule run page.

PR #706, opened 2026-03-23
ci: only use requested images; enable opcache

jrchamp:
- conditionally enable only the active database (inspired by the moodle/moodle workflow)
- switch to current version of checkout action (v4 is emitting a deprecation warning)
- enable opcache for CLI (saves ~30 seconds during Moodle install)
- bumps the postgres image to 16, which is required by Moodle `main`/5.3

I'll be pushing for similar changes directly in moodle core or moodle-plugin-ci, but I am already using these for my tests so I wanted to share them with y'all. ❤️

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
 
     services:
       postgres:
-        image: postgres:15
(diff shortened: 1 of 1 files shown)


smbader (approved):
Wow! This looks great, thank you for keeping us up-to-date.

PR #708, opened 2026-04-08
ical Notifications: Fix users to notify

paulandm:
Send ical notifications task: Fix users to notify.

Previously we made use of the 'get_users_by_capability', but we noticed that is not sufficient. 

We copied a course that contained a zoom activity scheduled for a future time.
The copied course doesn't contain any user enrolments (we selected that option).
But then the ical notifications would still be sent to users with high-level capabilities in the copied course's zoom activity. This is incorrect since it should only be sent to users enrolled within the course.

### DIFF
--- a/classes/task/send_ical_notifications.php
+++ b/classes/task/send_ical_notifications.php
@@ -272,7 +272,7 @@ class send_ical_notifications extends scheduled_task {
      */
     private function get_users_to_notify(int $zoomid, int $courseid) {
         $cminfo = get_fast_modinfo($courseid)->instances['zoom'][$zoomid];
-        $users = get_users_by_capability($cminfo->context, 'mod/zoom:view');
+        $users = get_enrolled_users($cminfo->context, 'mod/zoom:view', 0, 'u.*', null, 0, 0, true);
 
         if (empty($users)) {
             return [];


classes/task/send_ical_notifications.php
```diff
@@ -272,7 +272,7 @@ private function create_ical_object(
      */
     private function get_users_to_notify(int $zoomid, int $courseid) {
         $cminfo = get_fast_modinfo($courseid)->instances['zoom'][$zoomid];
-        $users = get_users_by_capability($cminfo->context, 'mod/zoom:view');
+        $users = get_enrolled_users($cminfo->context, 'mod/zoom:view', 0, 'u.*', null, 0, 0, true);
```
jrchamp, line 275: In review: `get_users_by_capability($context, $capability)` and `get_enrolled_users($context, $capability)` only differ in that the latter also requires an enrollment in the course. The next four parameters have been set to the default. Only the final parameter is not default, setting `$active` to `true`, which excludes suspended and expired enrollments. The change overall thus requires active enrollments, which aligns with Moodle's stance that only active course participants should be notified about activity events (like forum posts).

jrchamp (commented):
Thanks @paulandm ! This looks great to me! I added a comment to detail my technical understanding of the change in case it is helpful to other reviewers.

smbader:
Hello Paul!

This is a great contribution! The change looks good as well - we are going to do some quick testing and get this approved soon.

Thank you so much!
Steve

smbader (approved):

PR #709, opened 2026-04-14
webservice: get_recording_url_list error handling

### LINKED ISSUE #684: Recording deleted after AWS outage on October 20, 4am EDT
All the entries (3523) in mdl_zoom_meeting_recordings table were deleted at 4am EDT on October 20. This coincide with the AWS outage and Zoom incident (https://www.zoomstatus.com/incidents/yy70hmbp61r9).

We can see some WS errors in the mod_zoom\task\delete_meeting_recordings task log, but it still continues to delete recordings.

It looks like #539 and #593 did not solve this problem, where recordings are beeing deleted when the WS is not responding or not returning what we expected.

Here is our task log  : 

[logs.pdf](https://github.com/user-attachments/files/23004384/logs.pdf)

### LINKED ISSUE #704: INQUIRY ABOUT ZOOM RECORDINGS
Hello,

I have a question. I have meetings created in Moodle, but the Zoom recordings are deleted periodically. The recording link displayed in Moodle then shows that the recording does not exist because it has been deleted. Is there any way to delete the link displayed in Moodle when the recording is deleted?

Thanks

jrchamp:
We need get_recording_url_list() to return an array of recordings. So when the API gives us an error message that means "there are no recordings", then we need to return an empty array (so that it matches). If the API gives us an error message that means "there was a problem when we tried to find the recordings", then we don't want to pretend that we know how many recordings there are (so we allow other exception types to bubble up). Finally, if there was no API error message, then we should have an API response that we can use. However, the code is written to assume that we know how the API response is structured. If the code is in the structure that we expect, then use it. Otherwise, we should throw our own "unexpected response structure" error message to avoid returning incorrect information.

Fixes #684 
Fixes #704

### DIFF
--- a/classes/webservice.php
+++ b/classes/webservice.php
@@ -1135,9 +1135,22 @@ class webservice {
         // Classic: recording:read:admin.
         // Granular: cloud_recording:read:list_recording_files:admin.
         $url = 'meetings/' . $this->encode_uuid($meetingid) . '/recordings';
-        $response = $this->make_call($url);
 
-        if (!empty($response->recording_files)) {
+        try {
+            $response = $this->make_call($url);
+        } catch (not_found_exception $e) {
+            // If the meeting was not found (1001) or there are no recordings (3301), return an empty array.
+            return [];
+        }
+
+        if (empty($response->recording_files)) {
+            $recordingcount = (int) $response->recording_count;
+            $audiocount = count($response->participant_audio_files);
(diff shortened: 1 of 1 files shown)


cbounphengsy (approved):
If any related issues related arise, the added error-handling will allow us to narrow down the cause.

smbader:
Thank you for this fix! Now we wait for AWS to go down again!

PR #710, opened 2026-04-14
behat: configure during step; javascript is unnecessary

jrchamp:
At this time, the view_meeting Behat test does not require the `@javascript` tag and should therefore not include it. The generator class was setting configuration, which prevented values set in the Behat feature file from being used.

Fun side note: the mod_zoom plugin is saving all of its configuration settings using the outdated naming convention (without the mod_ prefix). Moodle's [developer documentation](https://moodledev.io/docs/5.2/apis/subsystems/admin) has included a note about this since May 2015:

> new activity modules should use 'mod_mymodule/setting' instead of 'mymodule/setting'

Looks like it was primarily using the modern frankenstyle name until #214. Meanwhile, look how naive I was in early 2021: https://github.com/ncstate-delta/moodle-mod_zoom/issues/180#issuecomment-777835649 😅

### DIFF
--- a/tests/behat/view_meeting.feature
+++ b/tests/behat/view_meeting.feature
@@ -2,7 +2,12 @@
 Feature: View a meeting
 
   Background:
-    Given the following "users" exist:
+    Given the following config values are set as admin:
+      | config       | value | plugin | encrypted |
+      | accountid    | test  | zoom   |           |
+      | clientid     | test  | zoom   |           |
+      | clientsecret | test  | zoom   |           |
+    And the following "users" exist:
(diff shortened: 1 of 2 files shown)


jrchamp:
Performance comparison based on recent CI runs for arguably the simplest Behat test ever (something to be proud of!):

- With `@javascript`: 7.2-9.1 seconds
- Without `@javascript`: 4.5-6.6 seconds

cbounphengsy (approved):
Setting the config values in a more Behat way seems like the logical move.

Issue #707, opened 2026-03-26
Prep for Moodle 5.2 release


smbader:
- [x] Check [Moodle 5.2 developer update](https://moodledev.io/docs/5.2/devupdate) documentation, particularly:
  - [x] Activity chooser descriptions refinement
- [ ] Update workflow / automated testing
- [x] Release to Moodle plugin repository

smbader:
Deadline for early bird badge April 24th.

jrchamp:
Looks like they moved up the deadline. From https://moodle.org/mod/forum/discuss.php?d=473769:

> Any plugin with a verified 5.2-compatible version available by **Monday, 20 April 2026**, 23:59 UTC will earn the Early bird 5.2 badge.

smbader:
Yikes, good catch.  Hopefully, we'll have our 5.2 request ready for this Thursdays meeting :)

PR #711, opened 2026-04-14
lang: update modulename_ strings to add support for 5.2

jrchamp:
This helps meet #707 by updating `modulename_help` and adding `modulename_summary`, `modulename_tip`. I added a bit too much text, because I think it is helpful (like lorem ipsum, but better) to get a sense of what it might be.

Before the changes, here's what it looks like:
Current mod_zoom activity description with a header and a single, sad sentence

Here's what `modulename_help` looks like in Moodle 5.1:
Modified mod_zoom activity description on Moodle 5.1 with a header and couple interesting paragraphs

Here's all three strings in Moodle 5.2:
Modified mod_zoom activity description on Moodle 5.2 with a header, a subheader, a couple interesting paragraphs and a tip callout

For me, it's helpful to think of these strings as part of a Bootstrap "card":

- `modulename`: the "card" header
  - very short basic string that is used throughout Moodle
- `modulename_summary`: the "card" subheader/tagline
  - single sentence basic string in Moodle 5.2
- `modulename_help`: the "card" body
  - a paragraph or two string
  - supports HTML in all supported Moodle versions
  - supports Markdown in Moodle 5.2
- `modulename_tip`: the "card" body's callout
  - single sentence basic string in Moodle 5.2

I know that these strings can be better. I fully invite you to make changes (you have edit access to this branch). For the moment, I think this is a good start, even with the duplication. Maybe someone can have coffee with a buddy from LearnTech and take this from "barely passable" to "good enough to merge". 😆 ☕


cbounphengsy (approved):
Providing more information is always a plus!

smbader:
This really looks great, thank you for keeping us up-to-date.  5.2 looks to have so many improvements.

PR #713, opened 2026-04-21
settings: cleanup

jrchamp:
Cleanups:

- `$moodlehashideif` is always `true` because this plugin already requires Moodle 3.7+.
  - Remove "false" branches and tidy.
  - Remove `environmentlib.php` because we no longer use `normalize_version()`.
- Use notification constants instead of static strings. Requires Moodle 3.1.
- Use `new lang_string()` when not immediately using the string. Requires Moodle 2.3.
- Use admin_setting_configcheckbox_with_lock for locked fields. Requires Moodle 2.0.
- Avoid creating unnecessary local variables.

Notes:

- "use" statements are completely ignored until the code that needs them is run. So every version of PHP that supports "use" statements (5.3 and newer) can have a reference like `use fakenamespace\fakeclass;` and the code will work just fine (because it ignores references until they are needed).
- `$hassiteconfig` offers a small amount of defense-in-depth. It's checked before the file is included, but just in case!
- While it will probably always be safe to use the `modsettingzoom` page name, because `plugininfo_mod` (and `admin/settings/plugins.php` before that) provide us with an appropriately instantiated `admin_settingpage`, we should not create our own and should use the name attribute that the class provides and uses as the official section name. One less thing to cause a problem if Moodle were to decide to move all plugins into a single folder.
- I didn't modify the MOODLE_INTERNAL guard in `classes/external.php`, because that will be resolved when #673 is merged.

### DIFF
--- a/backup/moodle2/backup_zoom_activity_task.class.php
+++ b/backup/moodle2/backup_zoom_activity_task.class.php
@@ -23,7 +23,7 @@
  * @license   http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 
-defined('MOODLE_INTERNAL') || die;
+defined('MOODLE_INTERNAL') || die();
 
 require_once($CFG->dirroot . '/mod/zoom/backup/moodle2/backup_zoom_stepslib.php');
 
--- a/db/caches.php
+++ b/db/caches.php
@@ -22,7 +22,7 @@
  * @license    http://www.gnu.org/copyleft/gpl.html GNU GPL v3 or later
  */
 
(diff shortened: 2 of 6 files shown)


cbounphengsy (approved):
Awesome! By removing unnecessary code use and optimizing the settings, we can increase its efficiency and security.

smbader (approved):
A massive Spring cleaning effort!
I learned a couple things as well:

- `new lang_string`
- `admin_setting_configcheckbox_with_lock`

Thanks Champ!

PR #717, opened 2026-05-18
lib: events are visible when the module is visible to the user

### LINKED ISSUE #716: Calendar and Timeline events are not respecting group restrictions in mod_zoom v5.5.0 on Moodle 4.5.7
Hi,

We are using **Moodle 4.5.7** with **mod_zoom v5.5.0**.

We have several Zoom activities in the same course, each restricted to a specific Moodle group. The course is set to use separate groups, and each Zoom activity is restricted to the relevant group. Learners appear to be separated correctly, but tutors still see Zoom events for other tutors’ groups in the Moodle **Calendar** and **Timeline**.

We have already checked the common configuration causes:

* Course group mode is set to separate groups.
* Each Zoom activity is restricted to the correct group.
* Copied Zoom activities have different Zoom meeting IDs.
* Tutor roles have been checked for group/availability bypass capabilities.
* The issue still occurs after reviewing role permissions and activity restrictions.

The problem appears to be that Zoom calendar events are created as course-level action events rather than group-scoped or availability-aware events. As a result, Moodle Calendar and Timeline show Zoom schedule events even when the current user should not see the corresponding group-restricted Zoom activity.

From what we can tell, the calendar event creation in `mod_zoom` sets fields such as:

```php
$event->type = CALENDAR_EVENT_TYPE_ACTION;
$event->modulename = 'zoom';
$event->eventtype = 'zoom';
$event->courseid = $zoom->course;
$event->instance = $zoom->id;
```

but does not appear to set a `groupid`, nor does it appear to implement a visibility callback that re-checks whether the current user can actually see the course module. We noticed this issue after the latest version of the plugin was installed.

Could you please add a permanent fix so that Zoom Calendar and Timeline events respect Moodle group restrictions and activity availability?

A possible fix would be to implement:

```php
mod_zoom_core_calendar_is_event_visible()
```

and return `false` when the related Zoom course module is not visible to the current user, for example by checking the user-specific course module visibility through `get_fast_modinfo()` and `$cm->uservisible`.

Expected behaviour:

* A learner or tutor assigned only to Group A should only see Zoom Calendar/Timeline events for Group A.
* They should not see scheduled Zoom events for Group B, Group C, etc.
* Calendar and Timeline visibility should match the actual visibility/accessibility of the Zoom activity in the course.

Actual behaviour:

* Learners appear to be filtered correctly.
* Tutors see Zoom schedule events from other groups in Calendar and Timeline, even when the Zoom activities are restricted to those groups.

This is problematic because it exposes other groups’ scheduled sessions and creates confusion for tutors.

Could this please be reviewed as a bug and fixed in a future release of `mod_zoom`?

Thanks.

jrchamp:
I saw the suggestion in #716 that we should be implementing `course_calendar_is_event_visible`. The part that I don't understand is why Moodle doesn't run a generic version of this code itself. Maybe some events are intended to be visible to everyone? Please double check that this actually does the right thing. It seems correct, perhaps a bit overly safe, but I'd rather start with safe. Feel free to suggest changes of course.

Fixes #716

### DIFF
--- a/lib.php
+++ b/lib.php
@@ -1365,3 +1365,34 @@ function zoom_cm_info_dynamic(cm_info $cm) {
 function zoom_apply_filter_on_meeting_name($name, $options) {
     return substr(format_string($name, true, $options + ['escape' => false]), 0, 200);
 }
+
+/**
+ * Checks whether a given calendar event should be visible to the user.
+ *
+ * @param calendar_event $event The calendar event object.
+ * @param ?int $userid User ID.
+ * @return bool True if visible, false otherwise.
+ */
+function mod_zoom_core_calendar_is_event_visible(calendar_event $event, $userid = null) {
+    global $USER;
+
+    // Only check our events.
+    if (empty($event->instance) || $event->modulename !== 'zoom') {
+        return false;
+    }
+
+    try {
+        $modinfo = get_fast_modinfo($event->courseid, $userid ?? $USER->id);
+        $cm = $modinfo->instances['zoom'][$event->instance] ?? null;
+
+        if (empty($cm)) {
+            return false;
+        }
+
+        // Return the user's visibility for the module.
+        return $cm->uservisible;
+    } catch (\moodle_exception $e) {
+        // Hide the event if the activity module does not exist.
+        return false;
+    }
+}


smbader (approved):
As mentioned, this really feels like the work Moodle core should do.  This function seems like a place to alter or extend the check in unusual ways if the plugin needed to.

Smart code, thank you!

